PR #5, opened 2021-07-01
Tx Service restructure

jakekidd:


packages/txservice/src/provider.ts
```diff
@@ -0,0 +1,177 @@
...
+
+import { TransactionServiceConfig } from "./config";
+import { ChainError } from "./error";
+import { FullTransaction, MinimalTransaction, ProviderConfig } from "./types";
+
+const { JsonRpcProvider, FallbackProvider } = providers;
+
+/// Could use a more encompassing name, e.g. ChainRpcDispatch, etc
+export class ChainRpcProvider {
+  private provider: providers.JsonRpcProvider | providers.FallbackProvider;
```
LayneHaber, line 15: We should likely prioritize the fallback provider stuff we started in vector much more highly here. How forward looking is this class? Does the contained `provider` cause issues or is that why its an `| providers.FallbackProvider`

May need a refresher on the issues we previously ran into with this

packages/txservice/src/provider.ts
```diff
@@ -0,0 +1,177 @@
...
+  private config: { gasInitialBumpPercent: number; gasPriceMinimum: BigNumber; rpcProviderMaxRetries: number; };
```
LayneHaber, line 16: Should the `gasPriceMaximum` be included here? What about a `timeout` for the confirmation?

packages/txservice/src/provider.ts
```diff
@@ -0,0 +1,177 @@
...
+            user: config.user,
+            password: config.password,
+          }
+        ),
+        priority: config.priority ?? 1,
+        weight: config.weight ?? 1,
+        stallTimeout: config.stallTimeout,
+      }));
+      // TODO: Quorum is set to 1 here, but we may want to reconfigure later. Normally it is half the sum of the weights,
+      // which might be okay in our case, but for now we have a low bar.
```
LayneHaber, line 54: Will this always default to whichever provider returns fastest?

packages/txservice/src/provider.ts
```diff
@@ -0,0 +1,177 @@
...
+      this.provider = new FallbackProvider(fallbackProviderConfigs, 1);
+    }
+
+    this.signer = typeof signer === "string" ? new Wallet(signer, this.provider) : signer.connect(this.provider);
```
LayneHaber, line 58: could the `signer` ever be a mnemonic?

packages/txservice/src/provider.ts
```diff
@@ -0,0 +1,177 @@
...
+  }
+
+  public async sendTransaction(
+    tx: FullTransaction
+  ): Promise<{ response: providers.TransactionResponse | Error; success: boolean }> {
+    const ready = await this.provider.ready;
+    if (!ready) {
+      // Error out, no providers are ready.
+      throw new ChainError(ChainError.reasons.ProviderNotReady);
+    }
```
LayneHaber, line 68: Can you add some documentation around what `ready` means? does that just mean the pollers have started, or does it mean the providers are synced?

packages/txservice/src/provider.ts
```diff
@@ -0,0 +1,177 @@
...
+    // TODO: Check getBlockNumber() > 0.
```
LayneHaber, line 69: Still don't love this check, ideally it would be that the `sync` call returns true, but tbh I'm only 80% certain that call exists and am not familiar at all with the behavior of that endpoint in prod cases where the node is truly out of sync

packages/txservice/src/provider.ts
```diff
@@ -0,0 +1,177 @@
...
+
+    // Define task to send tx with proper nonce
+    const task = async (): Promise<{ response: providers.TransactionResponse | Error; success: boolean }> => {
+      try {
+        // Send transaction using the passed in callback.
+        // const stored = this.nonces.get(chainId);
```
LayneHaber, line 75: We will eventually have to store the nonces in the class to get rid of the race conditions around repeated calls to `pending` and `pending` actually updating

packages/txservice/src/provider.ts
```diff
@@ -0,0 +1,177 @@
...
+        // const nonceToUse: number = nonce ?? stored ?? (await signer.getTransactionCount("pending"));
+        const pending: number = await this.signer.getTransactionCount("pending");
+        const { to, data, chainId, value, gasPrice, nonce } = tx;
```
LayneHaber, line 78: do we want to do any validation on the `chainId` or `gasPrice` here?

packages/txservice/src/provider.ts
```diff
@@ -0,0 +1,177 @@
...
+        });
+        // After calling tx fn, set nonce to the greatest of
+        // stored, pending, or incremented
+        // const pending = await signer.getTransactionCount("pending");
+        // const incremented = (response?.nonce ?? nonceToUse) + 1;
+        // Ensure the nonce you store is *always* the greatest of the values
+        // const toCompare = stored ?? 0;
+        // if (toCompare < pending || toCompare < incremented) {
+        //   this.nonces.set(chainId, incremented > pending ? incremented : pending);
+        // }
```
LayneHaber, line 95: Teehee this was my trashy code, lets update it to the nicer stuff instead, [here](https://github.com/connext/vector/blob/da603eb661bf888a267d802ba315e27c6c0ae028/modules/contracts/src.ts/services/ethService.ts#L210-L232)

:)

packages/txservice/src/provider.ts
```diff
@@ -0,0 +1,177 @@
...
+        if (!response) {
+          throw new ChainError(ChainError.reasons.FailedToSendTx);
+        }
+        return { response, success: true };
+      } catch (e) {
+        return { response: e, success: false };
+      }
+    };
+    // Queue up the execution of the transaction.
+    return await this.queue.add(task);
```
LayneHaber, line 106: amazing task looks fresh and clean!

packages/txservice/src/provider.ts
```diff
@@ -0,0 +1,177 @@
...
+  }
+
+  public async confirmTransaction(
+    transactionHash: string,
+    timeout: number,
+  ) {
+    // We are using waitForTransaction here to leverage the timeout functionality internal to ethers.
+    // IF this times out, ethers will reject with ("timeout exceeded", Logger.errors.TIMEOUT)
+    // Alternatively, it could reject with ("transaction was replaced", Logger.errors.TRANSACTION_REPLACED)
+    await this.provider.waitForTransaction(transactionHash, this.confirmationsRequired, timeout);
```
LayneHaber, line 116: I believe this does return a `TransactionReceipt` so you should only need the one call

packages/txservice/src/provider.ts
```diff
@@ -0,0 +1,177 @@
...
+    return await this.provider.getTransactionReceipt(transactionHash);
+  }
+
+  public async readTransaction(tx: MinimalTransaction) {
```
LayneHaber, line 120: Generally we'll want to build in some retries here as well. RPCs are fickle 🙄

packages/txservice/src/provider.ts
```diff
@@ -0,0 +1,177 @@
...
+        to: tx.to,
+        data: tx.data,
+      });
+      return readResult;
+    } catch (e) {
+      throw new ChainError(ChainError.reasons.ContractReadFailure, { error: jsonifyError(e) });
+    }
+  }
+
+  // TODO: Cache gas prices.
```
LayneHaber, line 132: 💯

packages/txservice/src/transaction.ts
```diff
@@ -0,0 +1,216 @@
+import { delay, jsonifyError } from "@connext/nxtp-utils";
+import { BigNumber, providers } from "ethers";
+import { BaseLogger } from "pino";
+
+import { TransactionServiceConfig } from "./config";
+import { ChainError } from "./error";
+import { ChainRpcProvider } from "./provider";
+import { FullTransaction, MinimalTransaction } from "./types";
+
+export class Transaction {
```
LayneHaber, line 10: Wow not even all the way through it, already love the separation

LayneHaber: Would be interesting to see it as an event emitter as well, especially since everything is event driven atm

packages/txservice/src/transaction.ts
```diff
@@ -0,0 +1,216 @@
...
+
+  public get fullTransaction(): Promise<FullTransaction> {
+    return new Promise(async (resolve) => 
+      resolve({
+        ...this.minTx,
+        gasPrice: await this.gasPrice,
+        nonce: this.nonce,
+      })
+    );
+  }
```
LayneHaber, line 31: Is there a chance of this completely swallowing up any errors around getting the gas price?

packages/txservice/src/transaction.ts
```diff
@@ -0,0 +1,216 @@
...
+        throw _response;
+      }
+    }
+    const response = _response as providers.TransactionResponse;
+
+    // Save nonce if applicable.
+    if (this.nonce === undefined) {
+      this.nonce = response.nonce;
+    }
+    // TODO: Should we be checking for wild case where this.nonce !== response.nonce?
```
LayneHaber, line 93: at the very least we should log it and everything related to this


cus that would be wild

packages/txservice/src/transaction.ts
```diff
@@ -0,0 +1,216 @@
...
+    const waitForReceipt = async (): Promise<providers.TransactionReceipt | undefined> => {
+      attempts += 1;
+      // Save all reverted receipts for a check in case our Promise.race evaluates to be undefined.
+      const reverted: providers.TransactionReceipt[] = [];
+      // Make a pool of promises for resolving each receipt call (once it reaches target confirmations).
+      const receipt = await Promise.race<any>(
+        this.responses
+          .map((response) => {
+            return new Promise(async (resolve) => {
+              const r = await this.provider.confirmTransaction(response.hash, confirmationTimeout);
```
LayneHaber, line 128: We should probably consider making the confirmation timeout specific to various chains

LayneHaber (commented):
Much *much* cleaner than what exists in vector, really want to get this load tested too.

packages/txservice/src/provider.ts
rhlsthrm: I think it should be based on provider syncing. Maybe we should change it to `synced` instead of ready.

packages/txservice/src/provider.ts
rhlsthrm: We should directly call the JSON-RPC method eth_syncing: https://eth.wiki/json-rpc/API. Let's bypass the provider-based checks completely for this.

packages/txservice/src/provider.ts
rhlsthrm: Yes we need to store it in the class, but we also need a way to recover from sync issues. One thing we are seeing in Vector is that sometimes the class-stored nonce gets out of sync with chain then it's fully unrecoverable. The only thing that fixes this case is restarting the router which makes it re-sync with chain. In the new system we need to be able to detect and recover from this case. Let's chat more about this @jakekidd.

Also to add on to @LayneHaber's comment, we don't just _eventually_ need this, we need the stateful nonce-tracking now, so may as well implement it.

rhlsthrm: Also it seems like this: https://docs.ethers.io/v5/api/experimental/#experimental-noncemanager

Is purpose-built for tracking nonces like this. Maybe we should just use it?

packages/txservice/src/provider.ts
rhlsthrm: We should validate those on the external-facing method.

packages/txservice/src/transaction.ts
```diff
@@ -0,0 +1,216 @@
...
+  public nonce?: number;
+  public nonceExpired = false;
+
+  // TODO: Not a fan of this arrangement, this code could be tightened up a bit.
+  public get gasPrice(): Promise<BigNumber> {
+    return this.getGasPrice();
+  }
+  private _gasPrice?: BigNumber;
+
+  public get fullTransaction(): Promise<FullTransaction> {
```
rhlsthrm, line 23: Is it weird for a getter like this to be async? Should it just be a private method?

packages/txservice/src/transaction.ts
rhlsthrm: Yes, the config should fully be chain-dependent, confirmation timeout, blocks to wait, etc.

packages/txservice/tsconfig.tsBuildInfo
```diff
@@ -0,0 +1 @@
...,[83,55],[84,56],[85,57],[117,58],[86,52],[87,59],[88,60],[90,61],[91,62],[92,63],[95,52],[96,64],[97,65],[98,66],[100,52],[101,67],[102,68],[114,69],[104,70],[105,71],[106,72],[108,66],[110,73],[111,66],[166,74],[165,75],[167,76],[164,77],[159,78],[160,79],[157,80],[193,81],[191,82],[192,83],[169,84],[170,85],[174,86],[171,84],[172,84],[161,87],[188,88],[197,89],[198,90],[196,91],[195,92],[182,93],[183,94],[178,95],[187,96],[179,74],[184,97],[175,98],[185,99]],"semanticDiagnosticsPerFile":[39,38,37,42,41,47,48,134,135,136,137,34,35,36,33,61,62,63,65,64,49,138,142,140,139,141,54,56,59,57,58,143,50,44,43,40,123,119,124,125,126,118,133,128,127,129,120,130,131,122,132,121,150,148,149,151,146,145,144,55,147,152,153,155,154,45,156,60,46,53,51,52,181,177,115,66,68,116,69,70,71,72,73,74,75,76,77,78,79,80,81,82,67,112,83,84,85,117,86,87,88,89,90,91,92,93,94,95,96,97,98,99,100,101,102,103,114,104,105,106,107,108,109,113,110,111,166,165,167,164,194,158,159,160,157,193,191,192,190,189,169,170,174,171,172,173,8,7,2,9,10,11,12,13,14,15,16,3,4,20,17,18,19,21,22,23,5,24,25,26,27,6,28,29,30,31,1,32,161,188,197,198,196,195,182,176,183,162,168,178,186,187,163,179,184,175,185,180]},"version":"4.3.4"}
```
rhlsthrm, line 1: Weird, this should be getting ignored, lol.

rhlsthrm (commented):
Great job so far. I like the separated class for transaction management. 

Let's add in the nonce management here then I think we can probably get the mergeable so we can start testing it on the live routers.

packages/txservice/src/provider.ts
jakekidd: previous issues were largely due to the fact that we were using providers all over the place.

No more. All provider instantiation and usage (with the exception of the injected provider in frontend/sdk atm) will be centralized here

packages/txservice/src/provider.ts
jakekidd: We could literally pass in the whole config, I think I'll just do that

packages/txservice/src/provider.ts
jakekidd: yep, assuming all providers have a weight of 1

I could leave it empty but my worry is what if like >half of providers fail, but one provider is working perfectly? Then the fallback provider will fail

We should always keep going so long as one (albeit one *trustworthy*) provider is still working.

packages/txservice/src/provider.ts
jakekidd: idk, since it's not being injected from frontend with this current architecture. But I thought, well, that could change, so better leave it

If we'd like to require it to be Signer type for now though, that's fine, lmk

packages/txservice/src/provider.ts
jakekidd: Sure I'll c/p definition from here:
https://docs.ethers.io/v5/api/providers/provider/#Provider-ready

and I'll change our error to be `ProviderNotSynced`

packages/txservice/src/provider.ts
jakekidd: lol yea, gotta get on this

packages/txservice/src/transaction.ts
jakekidd: Can you give me either an example or an approximation of what events it should be emitting? Like should we just emit on send and confirm?

packages/txservice/src/transaction.ts
jakekidd: Yeah gonna add a log.warning I guess

packages/txservice/src/transaction.ts
jakekidd: I will just remove confirmationTimeout from args here, and have it be a property of the provider instead, since the provider is chain-dependent

packages/txservice/src/provider.ts
```diff
@@ -0,0 +1,177 @@
...
+            url: config.url,
+            user: config.user,
+            password: config.password,
+          }
+        ),
+        priority: config.priority ?? 1,
+        weight: config.weight ?? 1,
+        stallTimeout: config.stallTimeout,
+      }));
+      // TODO: Quorum is set to 1 here, but we may want to reconfigure later. Normally it is half the sum of the weights,
```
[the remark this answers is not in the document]
jakekidd: That poses a very strange dilemma, but should be simple enough to solve. Have quorum be set by some optional argument passed into constructor (possibly a bool flag to indicate whether to set the quorum as half the sum of weights or just leave it as 1), then we will have our read/write classes instantiate providers differently

packages/txservice/src/provider.ts
```diff
@@ -0,0 +1,177 @@
...
+    }
+    // TODO: Check getBlockNumber() > 0.
+
+    // Define task to send tx with proper nonce
+    const task = async (): Promise<{ response: providers.TransactionResponse | Error; success: boolean }> => {
+      try {
+        // Send transaction using the passed in callback.
+        // const stored = this.nonces.get(chainId);
+        // const nonceToUse: number = nonce ?? stored ?? (await signer.getTransactionCount("pending"));
+        const pending: number = await this.signer.getTransactionCount("pending");
```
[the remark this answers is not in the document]
jakekidd: Yea. I'm thinking that we can achieve a cleaner approach here than what we had before since we are using the `Transaction` class, which can manage its own nonce once it's assigned one.

That being said, we still run into the problem of `this.signer.getTransactionCount("pending");` not returning the correct nonce or something. I also have to look over the latest code on main bc I know layne changed some nonce-related things (I vaguely recall from last week, but need refresher)

packages/txservice/tsconfig.tsBuildInfo
jakekidd: will remove from diff, might need a rebase or something

packages/txservice/src/transaction.ts
jakekidd: I was thinking it'd be useful for external calling in the event we want to log this transaction's properties

packages/txservice/src/provider.ts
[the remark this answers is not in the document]
jakekidd: Actually, circling back, this may be difficult. FallbackProvider doesn't allow for a change in quorum. So we can't just update the quorum if we're doing a read, and change it back for write.

Meaning we will need sep. provider instances for different quorums, and a write/read class would need to have access to both instances... which is weird

packages/txservice/src/provider.ts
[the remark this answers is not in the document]
jakekidd: Okay, see new solution. We are using `NonceManager` now, which should resolve our issues in getting nonce as it increments it internally. Worth some testing, ofc

packages/txservice/src/provider.ts
rhlsthrm: Should we even include a quorum then? It seems like our architecture is slightly different.

packages/txservice/src/provider.ts
rhlsthrm: It looks like the way you've structured it is you can pass in either a signer or a private key. I think this is fine, if there's a mnemonic it can be instantiated into a signer outside of this. Also this way we can pass in the injected signer.

packages/txservice/src/transaction.ts
rhlsthrm: Events to emit:
TRANSACTION_ATTEMPT_SUBMITTED (called on every gas bump)
TRANSACTION_CONFIRMED
TRANSACTION_FAILED

Check out [SDK](https://github.com/connext/nxtp/blob/main/packages/sdk/src/sdk.ts) for examples of how to set up the `attach` events. We like this much better than the standard node event emitter interface.

packages/txservice/src/provider.ts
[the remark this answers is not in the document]
rhlsthrm: Hmm, interesting. Let's just stick with the first provider to return strategy for now. Hopefully the `ready`/`synced` stuff will catch non-synced providers.

packages/txservice/src/provider.ts
jakekidd: The quorum under the hood by default will be half the sum of the weights - meaning, normally speaking, we need at least 1/2 of the providers to agree.

But we want the txservice to keep going as long as 1 provider is working, so we want a quorum of 1.

packages/txservice/src/provider.ts
jakekidd: Cool, I will keep then

packages/txservice/src/transaction.ts
jakekidd: Thank you @rhlsthrm , super helpful! Will get on this

jakekidd:
Re-requested reviews - the only step left is to fix `transaction.confirm`

LayneHaber (approved):
Looks awesome! Some questions and comments, but overall think this looks great. Can merge pending test fixes

packages/txservice/src/provider.ts
```diff
@@ -0,0 +1,259 @@
...
+    providerConfigs: ProviderConfig[],
+    private readonly config: TransactionServiceConfig,
+  ) {
+    this.confirmationsRequired = chainConfig.confirmations ?? config.defaultConfirmationsRequired;
+    this.confirmationTimeout = chainConfig.confirmationTimeout ?? config.defaultConfirmationTimeout;
+    this.chainId = chainId;
+    // TODO: Quorum is set to 1 here, but we may want to reconfigure later. Normally it is half the sum of the weights,
+    // which might be okay in our case, but for now we have a low bar.
+    // NOTE: This only applies to fallback provider case below.
+    this.quorum = 1;
```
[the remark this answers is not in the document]
rhlsthrm: 429/500 errors would throw, not return a successful response from the provider call. The getBlockNumber yes is a different case.

PR #8, opened 2021-07-03
Contract changes

LayneHaber:


LayneHaber:
NOTE: don't have to wait for demo to work before merging, it's already broken and all the contract deploys should be tagged from main. @rhlsthrm is there a good way to do that?

There are still some changes that are imminent for the contracts, this is a PR to review the existing tests

packages/contracts/test/lib/libAsset.spec.ts
```diff
@@ -33,7 +32,7 @@ describe("LibAsset", function() {
     loadFixture = createFixtureLoader([wallet, other]);
   });
 
-  beforeEach(async function() {
+  beforeEach(async function () {
```
LayneHaber, line 35: Generally we dont have to test things that aren't explicitly code written by us (i.e. reverting if the `assetId` is not an address. But, I also think the more tests the better so I'm happy to leave it

packages/contracts/test/lib/libAsset.spec.ts
```diff
@@ -72,8 +71,8 @@ describe("LibAsset", function() {
       expect(BigNumber.isBigNumber(res)).to.be.true;
```
LayneHaber, line 71: Need stronger assertions here

packages/contracts/test/lib/libAsset.spec.ts
```diff
@@ -72,8 +71,8 @@ describe("LibAsset", function() {
...
     });
 
-    it.skip("should return Erc20 asset balance if Non-AddressZero", async () => {
-      const Erc20TokenAddress = "";
+    it("should return Erc20 asset balance if Non-AddressZero", async () => {
+      const Erc20TokenAddress = token.address;
       const res = await libAssetTest.getOwnBalance(Erc20TokenAddress);
       expect(BigNumber.isBigNumber(res)).to.be.true;
```
LayneHaber, line 77: need stronger assertions here

packages/contracts/test/lib/libAsset.spec.ts
```diff
@@ -83,21 +82,22 @@ describe("LibAsset", function() {
...
       await expect(libAssetTest.transferEther(AddressZero, "1")).to.be.reverted;
     });
-    it.skip("should error if recipient call fails", async () => {});
 
     it.skip("happy case: transferEther", async () => {
       const amount = BigNumber.from(1);
-      await libAssetTest.connect(wallet).transferEther(receiver.address, amount, { from: wallet.address });
+      await libAssetTest.connect(wallet).transferEther(receiver.address, amount);
     });
   });
```
LayneHaber, line 90: Stronger assertions, not skipped

packages/contracts/test/lib/libAsset.spec.ts
```diff
@@ -83,21 +82,22 @@ describe("LibAsset", function() {
...
       await token.balanceOf(wallet.address);
 
-      const res = await libAssetTest
-        .connect(wallet)
-        .transferERC20(token.address, receiver.address, BigNumber.from(1), { from: wallet.address });
+      const approveRes = await token.connect(wallet).approve(libAssetTest.address, amount);
+      await approveRes.wait();
+
+      const res = await libAssetTest.connect(wallet).transferERC20(token.address, receiver.address, amount);
       await res.wait();
```
LayneHaber, line 101: Stronger assertions, not skipped

packages/contracts/test/lib/libErc20.spec.ts
```diff
@@ -11,7 +12,7 @@ import { TestERC20 } from "../../typechain/TestERC20";
 
 const createFixtureLoader = waffle.createFixtureLoader;
 
-describe("LibErc20", function() {
+describe("LibErc20", function () {
```
LayneHaber, line 15: Ideally should be an arrow function. If you need t manipulate things *about* the test (i.e. if not local and you call `evm_mine`) and use `this` then you can leave it as a `function ()` over `() => {}`

packages/contracts/test/lib/libErc20.spec.ts
```diff
@@ -40,22 +41,15 @@ describe("LibErc20", function() {
...
-  describe("#wrapCall", () => {
-    it.skip("should error if asset doesn't exist", async () => {});
-    it.skip("should error if params are buggy", async () => {});
-  });
-
   describe("#approve", () => {
-    it.skip("should error if params are buggy", async () => {});
     it("happy case:approve", async () => {
       await libERC20Test.connect(wallet).approve(token.address, other.address, "1");
     });
```
LayneHaber, line 47: Stronger assertions

packages/contracts/test/lib/libErc20.spec.ts
```diff
@@ -40,22 +41,15 @@ describe("LibErc20", function() {
...
 
   describe("#transferFrom", () => {
-    it.skip("should error if params are buggy", async () => {});
-    it.skip("happy case: transferFrom", async () => {
-      const approveRes = await token.connect(wallet).approve(receiver.address, "1");
+    it("happy case: transferFrom", async () => {
+      const approveRes = await token.connect(wallet).approve(libERC20Test.address, "1");
       await approveRes.wait();
 
       const res = await libERC20Test.connect(wallet).transferFrom(token.address, wallet.address, receiver.address, "1");
```
LayneHaber, line 55: Stronger assertions

packages/contracts/test/lib/libErc20.spec.ts
```diff
@@ -64,11 +58,9 @@ describe("LibErc20", function() {
...
 
   describe("#transfer", () => {
-    it.skip("should error if params are buggy", async () => {});
     it.skip("happy case: transfer", async () => {
-      await token.balanceOf(wallet.address);
-
-      const res = await libERC20Test.connect(wallet).transfer(token.address, receiver.address, "1");
+      const amount = BigNumber.from(1);
+      const res = await libERC20Test.connect(wallet).transfer(token.address, receiver.address, amount);
       await res.wait();
```
LayneHaber, line 64: Not skipped, stronger assertions

packages/contracts/test/temp.spec.ts
```diff
@@ -2,7 +2,7 @@ import { Contract, providers } from "ethers";
 
 import artifact from "../artifacts/contracts/test/TestERC20.sol/TestERC20.json";
 
-describe("Sample", function () {
+describe.skip("Sample Script", function () {
   const BscUSDC = "0x8AC76a51cc950d9822D68b83fE1Ad97B32Cd580d";
 
   const rpcUrl = "https://bsc-dataseed.binance.org";
```
LayneHaber, line 8: Can we delete this whole file or is this for testing against remote chains? If so, we shouldn't be executing only some test suites, and instead we should be executing the *entire* test suite to make sure that it can just be pointed at some testnet and we can run it

packages/sdk/src/utils/listener.ts
```diff
@@ -78,7 +78,7 @@ export class TransactionManagerListener {
         receivingChainId: txData.receivingChainId.toNumber(),
         callData: txData.callData,
         transactionId: txData.transactionId,
-        blockNumber: txData.blockNumber.toNumber(),
+        preparedBlockNumber: txData.blockNumber.toNumber(),
```
LayneHaber, line 81: We will need to reflect this change in the subgraph once we redeploy. Shouldn't block merging, just a reminder

packages/contracts/test/transactionManager.spec.ts
```diff
@@ -20,10 +23,8 @@ import { getOnchainBalance } from "./utils";
 
 const { AddressZero } = constants;
 
-type VariantTransactionData = {
-  amount: string;
-  blockNumber: BigNumberish;
-  expiry: string;
+const advanceBlockTime = async (desiredTimestamp: number) => {
```
LayneHaber, line 26: These are things (including the `getWallets()` call) that may be especially difficult to replicate if we aren't using a local hardhat instance. Ideally, all of these tests would be structured such that we can easily point these contract tests at any "evm compatible" chain to make sure these things work.

As a sidenote, we'll want to be *especially* careful around the assumptions were making with `SafeMath` when pointing at various rollup flavors

LayneHaber: Also (i can't highlight the right lines, sorry) we need to make sure we are using *real* deployments (i.e. including the multisender contracts) for these tests and test fixtures

LayneHaber: To avoid confusion, we should also make sure we add a set amount of liquidity to *all* the test wallets. We'll want to structure the utils now so they fund the addresses to a given level *only* if needed (i.e. if I want 5, and have 3, I only send an additional 2. If I want 5 and have 7, I don't send anything). We did some of this in the vector integration tests

packages/contracts/test/transactionManager.spec.ts
```diff
@@ -81,16 +82,16 @@ describe("TransactionManager", function () {
       receivingAddress: receiver.address,
       callData: "0x",
       transactionId: hexlify(randomBytes(32)),
-      sendingChainId: 1337,
-      receivingChainId: 1338,
+      sendingChainId: (await transactionManager.chainId()).toNumber(),
+      receivingChainId: (await transactionManagerReceiverSide.chainId()).toNumber(),
```
LayneHaber, line 86: We should rely on these and remove unnecessary overrides within the tests themselves

packages/contracts/test/transactionManager.spec.ts
```diff
@@ -240,9 +244,21 @@ describe("TransactionManager", function () {
...
+    const variantDigest = await getVariantTransactionDigest({
+      amount: record.amount,
+      expiry: record.expiry,
+      preparedBlockNumber: receipt.blockNumber,
+    });
+
+    expect(await instance.variantTransactionData(invariantDigest)).to.be.eq(variantDigest);
+    // expect(await instance.activeTransactionBlocks(preparer.address));
+
+    // const activeBlock = await instance.activeTransactionBlocks(transaction.user, 0);
```
LayneHaber, line 256: We should return these assertions, they are necessary

LayneHaber: We'll likely have to account for the following cases:
1. multiple txs are prepared in the same block for the same user
2. user has existing prepared txs

LayneHaber (commented):
Few things:

1. We should probably improve the interfaces for these helpers. The defaulting is a bit inconsistent, and could lead to confusion when revisiting
2. We should file an issue with hardhat and try to figure out what's going on with the events on the contracts. Don't love that we can only assert them from receipts :(
3. We should add tests to make sure you can't take the same action 2x (i.e. can't prepare the same tx twice, can't fulfill the same tx twice, can't cancel the same tx twice, can't fulfill *and* cancel a tx)

packages/contracts/test/transactionManager.spec.ts
```diff
@@ -299,20 +315,36 @@ describe("TransactionManager", function () {
...
+      amount: record.amount,
+      expiry: record.expiry,
+      preparedBlockNumber: 0,
+    });
+
+    expect(await instance.variantTransactionData(invariantDigest)).to.be.eq(variantDigest);
     // Assert event
     await assertReceiptEvent(receipt, "TransactionFulfilled", {
-      txData: { ...transaction, amount: record.amount, expiry: record.expiry, blockNumber: record.blockNumber },
+      txData: { ...transaction, ...record },
```
LayneHaber, line 347: We should make sure we have tests with actual relayers, not sure if we do

packages/contracts/test/transactionManager.spec.ts
```diff
@@ -486,8 +490,6 @@ describe("TransactionManager", function () {
       ).to.be.revertedWith("removeLiquidity: INSUFFICIENT_FUNDS");
     });
 
-    it.skip("should error if transfer fails", async () => {});
-
```
LayneHaber, line 490: We should return this test. We could use `smock` instead of writing a special ERC20 token transfer

packages/contracts/test/transactionManager.spec.ts
```diff
@@ -530,18 +559,109 @@ describe("TransactionManager", function () {
...
 
-    it("should revert if param sending or receiving chainId doesn't match chainId variable", async () => {});
+    it("should revert if param sending or receiving chainId doesn't match chainId variable", async () => {
+      const { transaction, record } = await getTransactionData({ sendingChainId: 1340, receivingChainId: 1341 });
+      await expect(
+        transactionManager
+          .connect(user)
+          .prepare(transaction, record.amount, record.expiry, "0x", "0x", { value: record.amount }),
+      ).to.be.revertedWith("prepare: INVALID_CHAINIDS");
+    });
```
LayneHaber, line 569: Ideally we'd refactor the utils to make them rejectable

packages/contracts/test/transactionManager.spec.ts
```diff
@@ -530,18 +559,109 @@ describe("TransactionManager", function () {
...
+      expect(receipt.status).to.be.eq(1);
+
+      const variantDigest = await getVariantTransactionDigest({
+        amount: record.amount,
+        expiry: record.expiry,
+        preparedBlockNumber: receipt.blockNumber,
+      });
+
+      expect(await transactionManager.variantTransactionData(invariantDigest)).to.be.eq(variantDigest);
+
```
LayneHaber, line 597: Use `prepareAndAssert`

packages/contracts/test/transactionManager.spec.ts
```diff
@@ -616,17 +736,167 @@ describe("TransactionManager", function () {
...
+        ),
+      ).to.be.revertedWith("fulfill: INVALID_VARIANT_DATA");
+    });
+
+    it("should revert if expiry of transaction is behind current blockstamp", async () => {
+      const { transaction, record } = await getTransactionData();
+
+      const relayerFee = "10";
+
+      const { blockNumber } = await prepareAndAssert(transaction, record, user, transactionManager);
```
LayneHaber, line 765: We should probably just have `prepareAndAssert` return the proper `record`

packages/contracts/test/transactionManager.spec.ts
```diff
@@ -616,17 +736,167 @@ describe("TransactionManager", function () {
...
+
+      const invariantDigest = getInvariantTransactionDigest(transaction);
+      const variant = {
+        amount: record.amount,
+        expiry: record.expiry,
+        preparedBlockNumber: blockNumber,
+      };
+      const variantDigestBeforeFulfill = getVariantTransactionDigest(variant);
+
+      expect(await transactionManager.variantTransactionData(invariantDigest)).to.be.eq(variantDigestBeforeFulfill);
```
LayneHaber, line 775: These checks seem unnecessary if using the `prepareAndAssert` utils

packages/contracts/test/transactionManager.spec.ts
```diff
@@ -616,17 +736,167 @@ describe("TransactionManager", function () {
...
+        ),
+      ).to.be.revertedWith("fulfill: ROUTER_MISMATCH");
+    });
 
-    it.skip("should revert iff it's receiving chain and relayer fee is non zero and transfer failed", async () => {});
-    it.skip("should revert iff it's receiving chain and calldata is non zero bytes and transfer failed", async () => {});
-    it.skip("should revert iff it's receiving chain and calldata is zero bytes and MultisendInterpreter failed", async () => {});
-    it.skip("should revert iff it's receiving chain and calldata is zero bytes and Interpreter isn't approved for respective amount when asset ERC20", async () => {});
+    it.skip("Happy case: iff it's receiving chain and calldata is non-zero bytes and MultisendInterpreter failed", async () => {});
+    it.skip("Happy case: iff it's receiving chain and calldata is non-zero bytes and MultisendInterpreter success", async () => {});
```
LayneHaber, line 899: Important, and we have to check that IF multisender fails, then the fallback address is transferred to (ie. `try-catch` is handled)

packages/contracts/test/lib/libAsset.spec.ts
sanchaymittal: yeah we don't really need it, going to clean it up.

packages/contracts/test/lib/libAsset.spec.ts
sanchaymittal: done

packages/contracts/test/lib/libErc20.spec.ts
sanchaymittal: updating to () => {}

packages/contracts/test/temp.spec.ts
sanchaymittal: removing this file

packages/sdk/src/utils/listener.ts
sanchaymittal: @rhlsthrm took care of this.

packages/contracts/test/transactionManager.spec.ts
sanchaymittal: gotchaa.

packages/contracts/test/transactionManager.spec.ts
sanchaymittal: we don't have one, but I don't think if that is necessary for the contract unit test

packages/contracts/test/transactionManager.spec.ts
sanchaymittal: I already tested these cases in the lib test so we don't really smock it in transaction Manager

packages/contracts/test/transactionManager.spec.ts
sanchaymittal: gotchaa

packages/contracts/test/transactionManager.spec.ts
sanchaymittal: why? To verify?

packages/contracts/test/transactionManager.spec.ts
sanchaymittal: These checks actually helped me to check the signature created by utils.

packages/contracts/test/transactionManager.spec.ts
sanchaymittal: updated

packages/contracts/test/transactionManager.spec.ts
sanchaymittal: Instead of pivoting out test setup from hardat to something compatible to run with test network. We can use fork setup to test contracts out, https://hardhat.org/guides/mainnet-forking.html.

This way we can test with any evm compatible chains.

sanchaymittal: we are actually adding set amount of liquidity to router wallet in beforeEach

packages/contracts/test/transactionManager.spec.ts
sanchaymittal: > multiple txs are prepared in the same block for the same user
we can test this out to check the activeTransactionBlocks>

> user has existing prepared txs
already taking care of this in multiple test cases

> assert the liquidity changes/lockup
> assert no additional funds were sent to the contract (i.e. finalContractBalance == initialContractBalance

will add these

PR #9, opened 2021-07-05
Sign tx ids for multipath routing

LayneHaber:
## Problem

- Multirouting transactions cannot be unlocked atomically when the entire `TransactionData` is signed (`router`, `callDataHash` are likely different)

## Solution
- Sign only the `transactionId`, which can be the same across all transactions


rhlsthrm:
This seems unsafe. Couldn't someone reuse a signature if there's an already used transactionId? If it's been fulfilled, someone can reuse a previous transactionId since we don't check there. Maybe we should store in the contract a mapping of used transactionIds.

LayneHaber:
It's the user who runs the auction that decides the tx id. They are also the ones who would get screwed if there was collision with their sigs. It seems generally okay to rely on the user choosing a sufficient source of randomness for this imo.

The issue I'm more concerned about is a race for the user unlocking the sender side transactions in the multipath payment before the routers can unlock the sender side and cancel the receiver side

LayneHaber:
> It's the user who runs the auction that decides the tx id. They are also the ones who would get screwed if there was collision with their sigs. It seems generally okay to rely on the user choosing a sufficient source of randomness for this imo.
> 
> The issue I'm more concerned about is a race between the user unlocking all the requisite sender side transactions and the routers who (once the sig is broadcast, but before it is submitted to chain) cancel the receiver-side payment and fulfill the sender-side payment

This is really only a problem when the expiry elapses, so it should be fine

rhlsthrm:
Will require significant subgraph changes to make sure IDs are not based on transactionId only. We can probably use `${transactionId}-${router}` to distinguish. Is this combination guaranteed unique @LayneHaber ?

LayneHaber:
> Will require significant subgraph changes to make sure IDs are not based on transactionId only. We can probably use `${transactionId}-${router}` to distinguish. Is this combination guaranteed unique @LayneHaber ?

I mean, it is not guaranteed unique by the contracts, and there's no great way to guarantee uniqueness there. Generally, the only uniqueness guarantee you have is that a user is *highly incentivized* to ensure the `transactionId-user` combination is unique. This is incentivized because:

1. User chooses the crosschain `transactionId` at the time of the auction broadcast
2. User's signature can be replayed if they do not choose a secure enough source of randomness, allowing *their* money to be stolen

Though it is trivial to enforce uniqueness on the same chain, the important case for is for crosschain, so enforcing same-chain uniqueness adds costs for minimal benefits. It is also important to note the only thing that guarantees crosschain *digest* uniqueness is the `transactionId`, so only signing the `transactionId` instead of the `digest` has fairly low security impacts

rhlsthrm:
> Generally, the only uniqueness guarantee you have is that a user is highly incentivized to ensure the transactionId-user combination is unique.

But in this case, if it's a split transaction there will be multiple transactions prepared with the `transactionId-user` combo right? I'm just trying to figure out what will work for the subgraph because a malicious actor could purposely break our subgraph by purposely adding colliding IDs. We might need to think about this more.

LayneHaber:
Merging this, `transactionId-user-router` should always be unique

PR #13, opened 2021-07-06
Demo updates

rhlsthrm:


packages/contracts/test/transactionManager.spec.ts
```diff
@@ -705,7 +705,7 @@ describe("TransactionManager", function () {
       ).to.be.revertedWith("prepare: INSUFFICIENT_LIQUIDITY");
     });
 
-    it("happy case: prepare by Bob for ERC20", async () => {
+    it.only("happy case: prepare by Bob for ERC20", async () => {
```
LayneHaber, line 708: Remove

packages/sdk/src/utils/listener.ts
```diff
@@ -4,40 +4,42 @@ import {
...
+import { TransactionManager } from "@connext/nxtp-contracts/typechain";
+import { providers } from "ethers";
 import { Evt } from "evt";
 
-import { NxtpSdkEvent, NxtpSdkEvents, TransactionCompletedEvent, NxtpSdkEventPayloads } from "../sdk";
-
 import { getTransactionManagerContract } from "./contract";
-// Define event types
+
+export const TransactionManagerEvents = {
```
LayneHaber, line 13: Do the sender/receiver ones need to be here?

LayneHaber (changes requested):
Just remove the .only

packages/sdk/src/utils/listener.ts
rhlsthrm: The way I'm thinking about this is the TransactionManagerListener class is directly tied to the contract so it mirrors the contract events exactly. At the higher SDK level, we consume these events and translate them into the SDK specific events which are sender/receiver.

PR #14, opened 2021-07-06
Multicall fixed

LayneHaber:
- Removes gnosis multicall
- Moves to an interpreter pattern, similar to the `WithdrawHelper` in vector
- Adds a `address callTo` to the `InvariantTransactionData` struct


LayneHaber:
Build broken because subgraph event signature changed, not sure how to fix this. Can we document it somewhere?

PR #24, opened 2021-07-14
Update algorithm to remove need for loops

LayneHaber:
Fixes c4 issue numbers:
- 27
- 24


sanchaymittal:
A bit unclear why are we using three mappings to track the active block number.

Can we do this:

```
  /// @dev Number of transactions in the block for that user.
  /// user -> block -> number of prepare transactions in a single block
 mapping(address => mapping(uint256 => uint256))  public activeTransactionBlocks;

function prepare(){
...
 activeTransactionBlocks(user)(block.number) += 1;
...
}

// removing active blocks
require(activeTransactionBlocks[user][preparedBlock] > 0, "removeUserActiveBlocks: NONE");
activeTransactionBlocks[user][preparedBlock] -= 1;

base0010 (approved):

PR #33, opened 2021-07-14
[Gas Optimisation] Use unchecked

sanchaymittal:
Fixes issue from the audit with number
- [74](https://github.com/code-423n4/2021-07-connext-findings/issues/74)


sanchaymittal:
Not using unchecked on lines
- https://github.com/connext/nxtp/blob/9956506d1fab8624d4a96a076848fe90a5c647bd/packages/contracts/contracts/TransactionManager.sol#L372

- https://github.com/connext/nxtp/blob/9956506d1fab8624d4a96a076848fe90a5c647bd/packages/contracts/contracts/TransactionManager.sol#L528

cause variable gets into the local scope.

packages/contracts/contracts/TransactionManager.sol
```diff
@@ -98,7 +98,9 @@ contract TransactionManager is ReentrancyGuard, ITransactionManager {
     }
 
     // Update the router balances
-    routerBalances[router][assetId] += amount;
+    unchecked {
+      routerBalances[router][assetId] += amount;
+    }
```
LayneHaber, line 103: can't this overflow?

packages/contracts/contracts/TransactionManager.sol
```diff
@@ -355,7 +361,9 @@ contract TransactionManager is ReentrancyGuard, ITransactionManager {
       require(msg.sender == txData.router, "fulfill: ROUTER_MISMATCH");
 
       // Complete tx to router for original sending amount
-      routerBalances[txData.router][txData.sendingAssetId] += txData.amount;
+      unchecked {
+        routerBalances[txData.router][txData.sendingAssetId] += txData.amount;
+      }
```
LayneHaber, line 366: this seems like it could also overflow

packages/contracts/contracts/TransactionManager.sol
```diff
@@ -257,7 +261,9 @@ contract TransactionManager is ReentrancyGuard, ITransactionManager {
       );
 
       // Decrement the router liquidity
-      routerBalances[invariantData.router][invariantData.receivingAssetId] -= amount;
+      unchecked {
+        routerBalances[invariantData.router][invariantData.receivingAssetId] -= amount;
+      }
```
LayneHaber, line 266: maybe add a comment referencing the `require` to explain why its unchecked here

LayneHaber (changes requested):

LayneHaber:
> Not using unchecked on lines
> 
> * https://github.com/connext/nxtp/blob/9956506d1fab8624d4a96a076848fe90a5c647bd/packages/contracts/contracts/TransactionManager.sol#L372
> * https://github.com/connext/nxtp/blob/9956506d1fab8624d4a96a076848fe90a5c647bd/packages/contracts/contracts/TransactionManager.sol#L528
> 
> cause variable gets into the local scope.

Not sure I understand how this means it uses less gas?

PR #45, opened 2021-07-16
[WIP] Non-constant supply tokens

LayneHaber:
Fixes the following audit issue:
- [68](https://github.com/code-423n4/2021-07-connext-findings/issues/68)


LayneHaber:
Changes:

- `removeLiquidity` now takes a number of `shares` rather than a given amount. There are helpers to make adjustments off chain for this
- `amount` is emitted as a separate field in the event. The `amount` is no longer guaranteed, only the number of `shares` is guaranteed
- `amount` is no longer included in `VariantTransactionData`, instead it is the number of `shares` associated with the transaction. (**NOTE**: the number of `shares` is assigned based on the `amount` passed into the `prepare` function)
-

LayneHaber:
Closing

PR #59, opened 2021-07-21
55 documentation

### LINKED ISSUE #55: [documentation] Add tsdocs to external functions
Modules to be documented (in priority order):

- [x] contracts
- [x] router
- [x] sdk
- [ ] transaction service
- [x] subgraph
- [x] utils

LayneHaber:
Fixes #55


packages/contracts/contracts/lib/LibAsset.sol
```diff
@@ -41,6 +56,11 @@ library LibAsset {
         IERC20(assetId).transfer(recipient, amount);
     }
 
+    /// @notice Transfers tokens from a sender to a given recipient
+    /// @param assetId Token address to transfer
+    /// @param from Address of sender/owner
+    /// @param to Address of recipient/spender
+    /// @param amount Amount to transfer from owner to spender
```
jakekidd, line 63: Would have preferred to use `/* */` since that's the typical styling, but will settle for `///`

Do you prefer the latter? Just want to agree on 1 style since consistency makes things look nice :)

LayneHaber: Oof yeah it's mostly cus these files don't get linted and the contract docs were already done. But I can probably make these changes since it is more consistent

LayneHaber:
Failing linting until txservice docs done 😢

packages/contracts/contracts/lib/LibAsset.sol
rhlsthrm: I think `/** */` is actually the standard. https://tsdoc.org

rhlsthrm:
> Failing linting until txservice docs done 😢

Shouldn't we just merge without the linting and then add the lint rule? Seems unnecessary to keep purely doc comments in a branch that's not main.

package.json
```diff
@@ -10,6 +10,7 @@
     "concurrently": "^6.2.0",
     "eslint": "^7.30.0",
     "eslint-config-prettier": "^8.3.0",
+    "eslint-plugin-jsdoc": "^35.5.0",
```
rhlsthrm, line 13: Can we use tsdoc instead of jsdoc? Not super familiar with the tradeoffs, but if tsdoc is more Typescript-compatible it seems better? https://tsdoc.org/pages/packages/eslint-plugin-tsdoc/

rhlsthrm:
More info about TsDoc, seems like it's the right thing to use instead of JsDoc: https://www.infoq.com/news/2018/04/tsdoc-format-announced/

rhlsthrm (approved):

packages/contracts/src/errors.ts
```diff
@@ -51,6 +67,12 @@ export const getFullError = (error: string): string => {
   return fullError;
 };
 
+/**
+ * Gets the contract error from the human readable string
+ *
+ * @param error - Contract error
+ * @returns Human readable translation of the error
+ */
```
jakekidd, line 75: We need to be consistent - are we doing comments using `/* */` format or `///` ?

packages/contracts/contracts/lib/LibAsset.sol
jakekidd: Should we not also specify the type for each param ? or is that unnecessary?

jakekidd: Scratch that ^ looks like type isn't necessary to specify according to standards.

jakekidd: Just realized this is solidity, so I'm assuming the /* */ format comment isn't relevant here

jakekidd (approved):
lgtm

package.json
jakekidd: Yes, we should, as I believe the interpreter will behave more consistently and take static typing into account

jakekidd (commented):
Actually, I agree with @rhlsthrm - would prefer tsdoc or jsdoc

package.json
LayneHaber: I originally did, but there's no way to enforce that things are actually documented with TSDoc which is *super* annoying

LayneHaber:
Recap on jsdoc/tsdoc:
- TSDoc doesn't have the ability to force that documentation is present, but everyone agrees it would be nicer

Potential solution (in future PRs):
- Use both TSDoc and JSDoc:
    - JSDoc should be configured to only enforce presence of documentation
    - TSDoc should be configured to enforce structure of documentation that *is* present. 
- Establish a code styling guide for TSDoc strings, and make sure that is eslint enforced

PR #66, opened 2021-07-22
Various improvements for TransactionService

jakekidd:
This PR should address the following concerns/issues:
- Load tests for further validation.
- Add TSDoc format docstrings (may be merged with #59).
- Add timeout for confirmation. Timeout should extend if confirmations are received.
- Fix the confirmation cycle to use an expiring (non-hanging/blocking) read method with minimal provider calls.
- Handle replacement tx on submit.


jakekidd:
This should have been draft, unfortunately. Won't compile in current state. WIP

PR #68, opened 2021-07-23
Sdk c4 68

sanchaymittal:


LayneHaber (approved):
Would like @rhlsthrm to review as well but lgtm!

packages/sdk/src/sdk.ts
```diff
@@ -365,31 +377,80 @@ export class NxtpSdk {
     }
 
     const inbox = generateMessagingInbox();
-    const receivedResponsePromise = new Promise<AuctionResponse>((res, rej) => {
+
+    let AuctionBids: AuctionResponse[] = [];
```
rhlsthrm, line 381: Don't capitalize variable name like this.

packages/sdk/src/sdk.ts
```diff
@@ -365,31 +377,80 @@ export class NxtpSdk {
...
+        } else {
+          AuctionBids.push(data);
+          if (AuctionBids.length >= 5) {
+            return;
+          }
         }
+      });
 
+      AuctionBids.sort((a, b) => {
+        return parseFloat(b.bid.amountReceived) - parseFloat(a.bid.amountReceived);
```
rhlsthrm, line 402: We can't parseFloat here, we are dealing with BigNumbers that could be larger than JS numbers.

packages/sdk/src/sdk.ts
```diff
@@ -365,31 +377,80 @@ export class NxtpSdk {
...
+          this.logger.error(
+            { method, methodId, signer, receivingChainId, receivingAssetId, router: data.bid.router },
+            routerLiq.error.message,
+          );
+          return;
+        }
 
-        // TODO: compare multiple responses
+        // check if the price changes unfovorably by more than the slippage tolerance(percentage).
+        const priceImpact = ((parseFloat(data.bid.amountReceived) - parseFloat(amount)) / parseFloat(amount)) * 100;
```
rhlsthrm, line 436: Need to make this BigNumber math.

rhlsthrm (commented):
Needs some fixes.

packages/sdk/src/sdk.ts
sanchaymittal: done

sanchaymittal:
fixes: https://github.com/connext/nxtp/pull/72

PR #72, opened 2021-07-25
fix: naming, parse scope

sanchaymittal:


packages/sdk/src/sdk.ts
```diff
@@ -477,7 +477,8 @@ export class NxtpSdk {
         }
 
         // check if the price changes unfovorably by more than the slippage tolerance(percentage).
-        const priceImpact = ((parseFloat(data.bid.amountReceived) - parseFloat(amount)) / parseFloat(amount)) * 100;
+        const priceImpact =
+          (parseFloat(BigNumber.from(data.bid.amountReceived).sub(amount).toString()) / parseFloat(amount)) * 100;
```
rhlsthrm, line 481: You can't parseFloat on BigNumbers. You need to do integer division. We have the `calculateExchange` function in the utils that's meant for this.

rhlsthrm (changes requested):
Still needs changes.

packages/sdk/src/sdk.ts
sanchaymittal: We need float division cause price impact is going to be in percentage.
Not sure how we can use the `calculateExchange()` func.

packages/sdk/src/sdk.ts
```diff
@@ -432,21 +433,21 @@ export class NxtpSdk {
...
+          if (auctionBids.length >= 5) {
             return;
           }
         }
       });
 
-      AuctionBids.sort((a, b) => {
-        return parseFloat(b.bid.amountReceived) - parseFloat(a.bid.amountReceived);
+      auctionBids.sort((a, b) => {
+        return BigNumber.from(b.bid.amountReceived).sub(a.bid.amountReceived).toNumber();
```
rhlsthrm, line 447: This will error because BigNumbers can't be negative.

rhlsthrm (changes requested):

PR #87, opened 2021-07-28
Unset flag once subgraph updated

LayneHaber:
Router submits duplicate txs to chain because of a race between router handling the tx and the subgraph being properly updated. This PR updates the flags to be unset only when the subgraph is updated. 

**NOTE**: Not sure if this is the best approach to prevent duplicate receiver-fulfill handlings. Doing so would mean querying the `SenderFulfilled` transactions, which would be an ever-growing list.

### DIFF
--- a/packages/router/src/handler.ts
+++ b/packages/router/src/handler.ts
@@ -688,6 +688,7 @@ export class Handler {
       );
       // If success, update metrics
     } else {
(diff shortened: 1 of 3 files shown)


rhlsthrm:
Let's rethink this with the new revert info we get.

PR #96, opened 2021-07-28
[DNM] Changes for Optimism contracts

rhlsthrm:
Need to formalize these changes to live side by side with our current contracts.


packages/contracts/contracts/TransactionManager.sol
```diff
@@ -204,9 +205,7 @@ contract TransactionManager is ReentrancyGuard, ProposedOwnable, ITransactionMan
     require(routerBalance >= amount, "#RL:008");
 
     // Update router balances
-    unchecked {
-      routerBalances[msg.sender][assetId] = routerBalance - amount;
-    }
+    routerBalances[msg.sender][assetId] = routerBalance - amount;
```
LayneHaber, line 208: If we're going to leave this change for both chains we should probably remove the require above

packages/contracts/contracts/TransactionManager.sol
```diff
@@ -1,14 +1,15 @@
...
 
 import "./interfaces/IFulfillInterpreter.sol";
 import "./interfaces/ITransactionManager.sol";
 import "./interpreters/FulfillInterpreter.sol";
 import "./ProposedOwnable.sol";
 import "./lib/LibAsset.sol";
-import "@openzeppelin/contracts/utils/Address.sol";
-import "@openzeppelin/contracts/security/ReentrancyGuard.sol";
-import "@openzeppelin/contracts/utils/cryptography/ECDSA.sol";
+import "./lib/OVM_Address.sol";
```
LayneHaber, line 10: No way we can see a diff on this is there?

packages/contracts/contracts/TransactionManager.sol
```diff
@@ -360,9 +359,7 @@ contract TransactionManager is ReentrancyGuard, ProposedOwnable, ITransactionMan
 
       // Decrement the router liquidity
       // using unchecked because underflow protected against with require
-      unchecked {
-        routerBalances[invariantData.router][invariantData.receivingAssetId] = balance - amount;
-      }
+      routerBalances[invariantData.router][invariantData.receivingAssetId] = balance - amount;
```
LayneHaber, line 362: Same as above

packages/contracts/contracts/lib/LibAsset.sol
```diff
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: UNLICENSED
-pragma solidity 0.8.4;
+pragma solidity >=0.7.6;
 
-import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+import "./OVM_SafeERC20.sol";
```
LayneHaber, line 4: Where is this contract? Same with the `Address` one..?

What's the diff between `SafeERC20` and `OVM_SafeERC20`?

rhlsthrm: Added

PR #99, opened 2021-07-28
Adjustments

jakekidd:
Minor adjustments to tx service logic including:

- ReadTransaction type with the `value` and `from` fields removed: for doing tx reads.

- prevents this (supposedly impossible case) from happening with a flag: "submitFailed".
    tx submit -> success -> confirm -> timeout -> bump gas -> tx submit -> revert! -> confirm -> timeout (?*) -> bump gas ??!? -> tx submit ??!?
If `submitFailed`, then we only try to confirm once.

- various sanity checks

### DIFF
--- a/packages/txservice/src/provider.ts
+++ b/packages/txservice/src/provider.ts
@@ -8,7 +8,7 @@ import { BaseLogger } from "pino";
 
 import { TransactionServiceConfig, ProviderConfig, validateProviderConfig, ChainConfig } from "./config";
 import { parseError, RpcError, TransactionError, TransactionReadError, TransactionServiceFailure } from "./error";
-import { FullTransaction, WriteTransaction, CachedGas, ReadTransaction } from "./types";
+import { FullTransaction, CachedGas, ReadTransaction } from "./types";
 
 const { StaticJsonRpcProvider, FallbackProvider } = providers;
 
--- a/packages/txservice/src/provider.ts
+++ b/packages/txservice/src/provider.ts
@@ -8,7 +8,7 @@ import { BaseLogger } from "pino";
 
 import { TransactionServiceConfig, ProviderConfig, validateProviderConfig, ChainConfig } from "./config";
 import { parseError, RpcError, TransactionError, TransactionReadError, TransactionServiceFailure } from "./error";
-import { FullTransaction, MinimalTransaction, CachedGas } from "./types";
+import { FullTransaction, WriteTransaction, CachedGas, ReadTransaction } from "./types";
 
 const { StaticJsonRpcProvider, FallbackProvider } = providers;
 
@@ -158,7 +158,7 @@ export class ChainRpcProvider {
    * @throws ChainError.reasons.ContractReadFailure in the event of a failure
    * to read from chain.
    */
-  public readTransaction(tx: MinimalTransaction): ResultAsync<string, TransactionError> {
+  public readTransaction(tx: ReadTransaction): ResultAsync<string, TransactionError> {
     return this.resultWrapper<string>(this.readTransaction.name, async () => {
       try {
         const readResult = await this.signer.call(tx);
--- a/packages/txservice/src/provider.ts
+++ b/packages/txservice/src/provider.ts
@@ -105,37 +105,36 @@ export class ChainRpcProvider {
    * and a success boolean indicating whether the process did result in an error.
    */
   public sendTransaction(tx: FullTransaction): ResultAsync<providers.TransactionResponse, TransactionError> {
+    // Do any parsing and value handling work here if necessary.
+    const transaction = {
+      to: tx.to,
+      data: tx.data,
+      chainId: tx.chainId,
+      gasPrice: tx.gasPrice,
+      nonce: tx.nonce,
+      value: BigNumber.from(tx.value || 0),
+    };
+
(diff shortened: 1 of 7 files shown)


packages/txservice/src/provider.ts
```diff
@@ -105,37 +105,36 @@ export class ChainRpcProvider {
...
           if (response == null) {
-            // Check to see if ethers returned null or undefined for the response; if so, handle as error case.
             throw new TransactionServiceFailure("Ethers returned a null or undefined transaction response.", {
-              transaction: tx,
+              transaction,
               response,
             });
           }
+
+          // We increment the nonce here, as we know the transaction was sent (response is defined).
```
LayneHaber, line 136: Has this ever happened before?

packages/txservice/src/types.ts
```diff
@@ -2,7 +2,13 @@ import { BigNumber, BigNumberish } from "ethers";
...
 import { TransactionServiceFailure } from "./error";
 
-export type MinimalTransaction = {
+export type ReadTransaction = {
+  chainId: number;
+  to: string;
+  data: string;
+};
+
+export type WriteTransaction = {
```
jakekidd, line 11: I should have WriteTransaction extend ReadTransaction

packages/txservice/src/transaction.ts
```diff
@@ -260,14 +260,19 @@ export class Transaction {
...
+    // Sanity checks.
     if (this.receipt == null) {
       // Receipt is undefined or null. This normally should never occur.
       throw new TransactionServiceFailure("Unable to obtain receipt: ethers responded with null.", {
         method,
         receipt: this.receipt,
       });
     } else if (this.receipt.status === 0) {
+      // This should never occur. We should always get a TransactionReverted error in this event.
       throw new TransactionServiceFailure("Transaction was reverted but TransactionReverted error was not thrown.");
```
LayneHaber, line 271: We should include the hash on the error here

packages/txservice/src/transaction.ts
```diff
@@ -276,7 +281,9 @@ export class Transaction {
       const result = await this.provider.confirmTransaction(response, undefined, 60_000 * 20);
       if (result.isErr()) {
         // No errors should occur during this confirmation attempt.
-        throw result.error;
+        throw new TransactionServiceFailure("Unable to confirm transaction even after receiving 1 confirmation.", {
+          error: result.error,
+        });
```
LayneHaber, line 285: should include hash as a separate json field

packages/txservice/src/txservice.ts
```diff
@@ -115,33 +115,35 @@ export class TransactionService {
...
         try {
           await this.submitTransaction(transaction, requestContext);
         } catch (error) {
+          // TODO: Check to see if the error is an AlreadyMined error, as we probably don't need to
+          // log this here if it is.
           this.logger.warn(
             { method, methodId, requestContext },
-            `(${transaction.id}, ${transaction.attempt}) ${error}`,
+            `Submit failed.`,
           );
```
LayneHaber, line 137: log error message at least?

packages/txservice/src/txservice.ts
```diff
@@ -115,33 +115,35 @@ export class TransactionService {
...
-          }
-          // IF this is the first attempt, throw.
+          // IFF this is the first attempt, throw.
           // Otherwise, we should go ahead and get the receipt.
           if (transaction.attempt <= 1) {
             throw error;
           }
+          // Save the submit error to indicate we've failed here; this should be the last execution
+          // of this loop. We won't throw the error below (it could be harmless, e.g. if tx is already
+          // mined, but we have logged it above.
```
LayneHaber, line 145: is it logged?

packages/txservice/src/txservice.ts
```diff
@@ -152,7 +154,7 @@ export class TransactionService {
             { method, methodId, requestContext },
             `(${transaction.id}, ${transaction.attempt}) ${error}`,
           );
```
LayneHaber, line 156: if we want these (id especially) to be json-searchable, we should make sure they're logged in the json above under a common field. if this is within the `requestContext` already that's fine

packages/txservice/src/provider.ts
jakekidd: Not that I know of, but I'm going into this new txservice with the mantra "if something goes wrong, I wanna know *exactly* what it is (and what it isn't)"

packages/txservice/src/transaction.ts
jakekidd: We should include error in context in general, good catch

packages/txservice/src/transaction.ts
jakekidd: kk

packages/txservice/src/txservice.ts
jakekidd: My bad, will move that in there.

packages/txservice/src/txservice.ts
jakekidd: Yes, in the warning above (it should have been in the context)

packages/txservice/src/txservice.ts
jakekidd: Will include that with these changes

LayneHaber (approved):

PR #109, opened 2021-07-31
Validate expiry

### LINKED ISSUE #107: [router] Handle error where expiry is too short
Error comes back as "execution reverted: #p:013", handle this properly by cancelling sender side.

rhlsthrm:
Fixes #107


packages/router/src/handler.ts
```diff
@@ -568,6 +569,8 @@ export class Handler {
     // that user is only sending stuff that makes sense is possibly ok since otherwise
     // they're losing gas costs
 
+    const receiverExpiry = mutateExpiry(txData.expiry);
```
jakekidd, line 572: This could throw, are we catching that?
```ts
throw new Error("Expiration already happened, cant prepare");
```

rhlsthrm: Nice catch, I totally missed this.

PR #110, opened 2021-08-04
[txservice] Load tests for local chain

### LINKED ISSUE #108: [txservice] Load testing local chain
Load tests for txservice should basically execute as many (simple) tx's as possible on a live test chain, and give us a metric to see how many tx's we can manage at once.

jakekidd:
Fixes #108

### DIFF
--- a/packages/integration/test/load/txserviceConcurrency.ts
+++ b/packages/integration/test/load/txserviceConcurrency.ts
@@ -34,7 +34,8 @@ const txserviceConcurrencyTest = async (maxConcurrency: number, step = 1): Promi
   // })();
 
   // Create manager
-  const manager = new OnchainAccountManager(config.chainConfig, config.mnemonic, maxConcurrency);
+  logger.info({ agents: maxConcurrency }, "Created manager. This may take a bit...");
+  const manager = new OnchainAccountManager(config.chainConfig, config.mnemonic, Math.min(maxConcurrency, 1000));
   logger.info({ agents: maxConcurrency }, "Created manager");
 
 
--- a/packages/integration/test/load/txserviceConcurrency.ts
+++ b/packages/integration/test/load/txserviceConcurrency.ts
@@ -10,9 +10,15 @@ import { OnchainAccountManager } from "../utils/accountManager";
 
 /**
  * Sets up a basic concurrency test through the core TransactionService only. Will slowly add more agents up to `maxConcurrency` sending transactions.
- *
  * @param maxConcurrency - Concurrency to build up to. We start at 1, and work up to this number. On each iteration, the concurrency value will be the number of agents who will do transactions simultaneously.
  * @param step - Step by which we increase the concurrency.
+ * 
+ * @remarks
+ * Quick start:
+ * 1. First run
+ *        yarn workspace @connext/nxtp-integration docker:all:down && yarn workspace @connext/nxtp-integration docker:services:up && bash setup-integration-test.sh
+ * 2. In another terminal run
+ *        yarn workspace @connext/nxtp-integration concurrency:txservice
  */
 const txserviceConcurrencyTest = async (maxConcurrency: number, step = 1): Promise<void> => {
   const config = getConfig();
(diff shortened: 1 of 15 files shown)


packages/txservice/src/error.ts
```diff
@@ -221,11 +221,13 @@ export const parseError = (error: any): NxtpError => {
...
-    throw new TransactionReverted(TransactionReverted.reasons.ExecutionFailed, undefined, context);
+    return new TransactionReverted(TransactionReverted.reasons.ExecutionFailed, undefined, context);
   } else if (message.match(/always failing transaction/)) {
-    throw new TransactionReverted(TransactionReverted.reasons.AlwaysFailingTransaction, undefined, context);
+    return new TransactionReverted(TransactionReverted.reasons.AlwaysFailingTransaction, undefined, context);
   } else if (message.match(/gas required exceeds allowance/)) {
-    throw new TransactionReverted(TransactionReverted.reasons.GasExceedsAllowance, undefined, context);
+    return new TransactionReverted(TransactionReverted.reasons.GasExceedsAllowance, undefined, context);
+  } else if (message.match(/tx doesn't have the correct nonce|another transaction with same nonce|same hash was already imported|transaction nonce is too low|nonce too low/)) {
+    return new AlreadyMined(AlreadyMined.reasons.NonceExpired, context);
```
jakekidd, line 230: This should match all nonce messages better. Learned today that they can come back as a SERVER_ERROR. In particular, they come back like this from ganache run locally

(Also the above `throw`s were a bug, changed to `return`

packages/integration/test/utils/accountManager.ts
```diff
@@ -81,4 +81,22 @@ export class OnchainAccountManager {
...
+    excluding.push(this.funder);
+    const filtered = this.wallets.filter((n) => {
+      const addrs = excluding.map((e) => e.address);
+      return !addrs.includes(n.address);
+    });
+    if (filtered.length === 0) {
+      throw new Error("Failed to get random wallet: none left after filtering.");
+    }
+    return filtered[Math.floor(Math.random() * filtered.length)];
+  }
```
jakekidd, line 101: @base0010 this ok?

base0010: looks fine to me.

LayneHaber:
Looks like transaction service tests are failing

rhlsthrm (approved):

packages/integration/test/load/txserviceConcurrency.ts
```diff
@@ -0,0 +1,125 @@
...
+      .fill(0)
+      .map((_, i) => {
+        const task = async () => {
+          const agent = manager.getRandomWallet();
+          const amount = "100000";
+          const txInfo: TransactionInfo = {
+            start: Date.now(),
+          };
+          try {
+            await txservice.sendTx(
```
base0010, line 81: Do you think it would make sense to test for different types of states that can get returned implicitly, like should we make sure this hits a contract in a way that we know returns a call_revert for example. Or is that outside the wheelhouse of txservice's concerns?

base0010 (approved):
Runs with local services and local chains

packages/integration/test/load/txserviceConcurrency.ts
```diff
@@ -0,0 +1,125 @@
...
+
+import { TransactionInfo } from "../utils/sdkManager";
+import { getConfig } from "../utils/config";
+import { OnchainAccountManager } from "../utils/accountManager";
+
+/**
+ * Sets up a basic concurrency test through the core TransactionService only. Will slowly add more agents up to `maxConcurrency` sending transactions.
+ *
+ * @param maxConcurrency - Concurrency to build up to. We start at 1, and work up to this number. On each iteration, the concurrency value will be the number of agents who will do transactions simultaneously.
+ * @param step - Step by which we increase the concurrency.
```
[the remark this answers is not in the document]
jakekidd: They are all *from* the same signer/account, yes, but go *to* diff "agents" or accounts

jakekidd:
> Looks like transaction service tests are failing

fixed

packages/integration/test/load/txserviceConcurrency.ts
jakekidd: That's not really in scope here bc we're just sending basic tx's, not contract calls (def no transaction manager contract calls), really

jakekidd: That *should* eventually be a part of integration testing though

packages/integration/test/load/txserviceConcurrency.ts
```diff
@@ -0,0 +1,125 @@
...
+                from: manager.funder.address,
+                data: "0x00",
+                value: BigNumber.from(amount),
+              } as WriteTransaction,
+              {
+                id: `C${concurrency}:U${i+1}`,
+                origin: "concurrencyTest",
+              } as RequestContext,
+            );
+            // TODO: Confirm after sending tx that balance reflects new amount?
```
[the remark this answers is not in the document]
jakekidd: cool, good point. gonna clarify that in the TODO (make it a readTx)

PR #114, opened 2021-08-05
Fix cancellation

LayneHaber:
- Fixes cancellation issues on sending chain
- Adds a new `ActiveTransaction` type to track sending and receiving chain variant data to allow cancellations on both chains by the sdk
- Updates test-ui to properly cancel things

**NOTE** changes the API for the `getActiveTransactions`


packages/sdk/src/sdk.ts
```diff
@@ -876,11 +875,45 @@ export class NxtpSdk {
...
+          "Mined cancel tx",
+        );
+        if (cancelParams.txData.sendingChainId === chainId) {
+          this.evts.SenderTransactionCancelled.post({
+            txData: cancelParams.txData,
+            relayerFee: cancelParams.relayerFee,
+            caller: cancelReceipt.from,
+            transactionHash: cancelReceipt.transactionHash,
+          });
+        }
```
LayneHaber, line 913: Receiver cancellations will be caught by the subgraph, but there is no good way (with current logic) to catch this on the subgraph on the receiver side

packages/sdk/src/subgraph.ts
```diff
@@ -95,10 +95,10 @@ export const createSubgraphEvts = (): {
 };
 
 export type ActiveTransaction = {
-  txData: TransactionData;
+  crosschainTx: CrosschainTransactionData;
   status: NxtpSdkEvent;
   bidSignature: string;
-  caller: string;
+  caller: string; // TODO: which caller tho?
```
LayneHaber, line 101: This doesn't make sense. There could be a receiver prepare caller, a sender/receiver fulfill caller, or a sender/receiver cancel caller. We should simply remove this from the type, or make it more specific. But as it stands this is very ambiguous

packages/sdk/src/sdk.ts
```diff
@@ -876,11 +875,45 @@ export class NxtpSdk {
   public async cancelExpired(cancelParams: CancelParams, chainId: number): Promise<providers.TransactionResponse> {
```
LayneHaber, line 875: Since this has nothing to do with expiries (i.e. user could, and should, cancel receiver-side transfers to signal to the router to unlock the sender side if there was a validation error, or some things changed). We should rename it to just `cancel`

LayneHaber: And do some basic validation on what chain is being cancelled and if it is expired (or use the tx service to run estimate gas and get the proper revert message)

packages/utils/src/transactionManager.ts
```diff
@@ -22,6 +22,12 @@ export type VariantTransactionData = {
 };
 export type TransactionData = InvariantTransactionData & VariantTransactionData;
 
+// Used to include *all* info for both sending and receiving crosschain data
+export type CrosschainTransactionData = InvariantTransactionData & {
+  sending: VariantTransactionData;
+  receiving?: VariantTransactionData;
+};
```
LayneHaber, line 29: A concern I have is around putting a requirement of the sdk consumer to know too much about the protocol to understand this type. That being said, I think it is *really* weird that the `amount`, `expiry`, or `preparedBlockNumber` would change on an active transaction once it changes status

packages/sdk/src/subgraph.ts
sanchaymittal: I think we can check the caller side from status, but we should surely make it clearer.
if we update it with every updated transaction status or if want to track each method(fulfill/cancel) caller on both sides.

PR #199, opened 2021-08-11
Historical data

rhlsthrm:


packages/sdk/src/subgraph.ts
```diff
@@ -295,6 +295,94 @@ export class Subgraph {
     return all;
   }
 
+  async getHistoricalTransactions(): Promise<CrosschainTransaction[]> {
```
LayneHaber, line 298: This should likely include any that were cancelled as well, not simply ones that were fulfilled

LayneHaber (changes requested):

packages/sdk/src/subgraph.ts
rhlsthrm: The idea was to do that in steps, starting with resolved only. Let me see if it's easy enough to get in here.

rhlsthrm: Done.

PR #218, opened 2021-08-16
[router] Handle transactions with unsupported chains

### LINKED ISSUE #131: [router] 4.4 Check transactions pulled from subgraph against the router's config

jakekidd:
Fixes #131


rhlsthrm (changes requested):
I don't think this is the right place for this logic. This should be done in the getActiveTransactions function itself, which should be config-aware.

jakekidd:
> I don't think this is the right place for this logic. This should be done in the getActiveTransactions function itself, which should be config-aware.

Got it. I wasn't sure if we wanted to take the extra step to propagate the config 1 more level or not here lol

rhlsthrm:
Actually I already incorporated these fixes into another PR so I think we can close this.

Issue #168, opened 2021-08-09
[router] 4.21 Make sure native asset is properly handled in prepare


LayneHaber:

rhlsthrm:
Not sure what needs to be done since this is all handled by the contract, but let's double-check.

PR #228, opened 2021-08-18
add: integration test for Native assets

sanchaymittal:
closes: #168 
closes: #242


packages/integration/test/integration.spec.ts
```diff
@@ -251,4 +253,47 @@ describe("Integration", () => {
     const fulfillEvent = await fulfillEventPromise;
     expect(fulfillEvent).to.be.ok;
   });
+
+  it.skip("should send Native tokens", async function () {
```
rhlsthrm, line 257: Why is this skipped?

rhlsthrm (changes requested):

packages/integration/test/integration.spec.ts
sanchaymittal: I was doing some changes and wanted to check the before config.

PR #237, opened 2021-08-19
233 sdk messaging

### LINKED ISSUE #233: [sdk] Refactor auction messaging to use a static endpoint and not subscribe/unsubscribe

rhlsthrm:
Fixes #233


rhlsthrm (commented):
Annoying to have to move back to the async static function since we already have integrations using the other method. Also we didn't want to init messaging immediately necessarily on page load. We should probably have a separate `init` function.

LayneHaber (approved):

PR #245, opened 2021-08-20
[DNM] Blockcade monitoring

LayneHaber:
Just a simple PR to get some ideas across. should not be merged

### DIFF
--- a/packages/txservice/src/provider.ts
+++ b/packages/txservice/src/provider.ts
@@ -506,6 +506,9 @@ export class ChainRpcProvider {
           // TODO: handle error
           return;
         }
+        // NOTE: this may not be the best way to confirm this tx. the point is
+        // it should *not* be bumped, it should be dealt with
+        // here and immediately to free up the queue of pending txs
         await this.confirmTransaction(response.value, 2);
         return;
       }
--- a/packages/txservice/src/provider.ts
+++ b/packages/txservice/src/provider.ts
@@ -41,6 +41,8 @@ export class ChainRpcProvider {
   private cachedGas?: CachedGas;
   private cachedDecimals: Record<string, number> = {};
 
+  private readonly pendingTxs: Record<number, { response: providers.TransactionResponse; added: number }> = {};
+
   public readonly confirmationsRequired: number;
   public readonly confirmationTimeout: number;
 
@@ -113,6 +115,8 @@ export class ChainRpcProvider {
 
     // TODO: #146 We may ought to do this instantiation in the txservice constructor.
     this.signer = typeof signer === "string" ? new Wallet(signer, this.provider) : signer.connect(this.provider);
+
+    this.monitorBlockades();
   }
 
   /**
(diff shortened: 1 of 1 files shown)


packages/txservice/src/provider.ts
```diff
@@ -218,7 +225,14 @@ export class ChainRpcProvider {
...
       // See issue: https://github.com/ethers-io/ethers.js/issues/1775
-      return (response as any).wait(confirmations ?? this.confirmationsRequired, timeout ?? this.confirmationTimeout);
+      const receipt = (response as any).wait(
+        confirmations ?? this.confirmationsRequired,
+        timeout ?? this.confirmationTimeout,
+      );
+
+      // Remove from pending txs once its been confirmed
+      delete this.pendingTxs[response.nonce];
+      return receipt;
```
jakekidd, line 235: What happens if `response.wait` throws?

packages/txservice/src/provider.ts
```diff
@@ -450,4 +464,85 @@ export class ChainRpcProvider {
   private incrementNonce() {
     this._nonce++;
   }
+
+  private async monitorBlockades() {
```
jakekidd, line 468: Def not a fan of having this in the ChainRpcProvider, but it may have to do for now. Eventually, would be best to be have a self-contained class for this.

packages/txservice/src/provider.ts
```diff
@@ -450,4 +464,85 @@ export class ChainRpcProvider {
...
+    while (true) {
+      // First, check to see that there are pending txs
+      if (Object.keys(this.pendingTxs).length === 0) {
+        return;
```
jakekidd, line 472: How would this work? `this.monitorBlockades` would be called in constructor and then immediately return... did you mean to maybe do a `delay(1000);` then `continue;` or something?

packages/txservice/src/provider.ts
```diff
@@ -450,4 +464,85 @@ export class ChainRpcProvider {
...
+      }
+
+      // Get the blocking nonce and corresponding record
+      const blockade = await this.provider.getTransactionCount("pending");
```
jakekidd, line 476: Shouldn't this just be the bottom of the stack?

jakekidd: Oh I see, you're using this as an index

packages/txservice/src/provider.ts
```diff
@@ -450,4 +464,85 @@ export class ChainRpcProvider {
...
+
+      // If the blockade is our *current* nonce, we can ignore since it will
+      // be sent at the next tx
+      // TODO: ^^ is this a safe assumption
+      if (blockade === (await this.getNonce())) {
+        return;
+      }
+
+      const { added, response } = this.pendingTxs[blockade] ?? {};
```
jakekidd, line 485: In this case, blockade should always === 0, though, no?

packages/txservice/src/provider.ts
```diff
@@ -450,4 +464,85 @@ export class ChainRpcProvider {
...
+      const TOO_LONG = 2 * 60 * 1000; // should be config determined, for now 2min
+
+      const blocked = Object.keys(this.pendingTxs).filter((nonce) => parseInt(nonce) > blockade).length;
+
+      const gasPrice =
+        elapsed >= TOO_LONG || blocked > 10
+          ? BigNumber.from("0xmaxgas")
+          : response.gasPrice!.add(BigNumber.from("0xgasbump").mul(2));
+
+      const resubmitted = await this.sendTransaction({
```
jakekidd, line 532: This def creates an antipattern with what's already in place with txservice, but I totally agree with this logic surface level

LayneHaber:
Closing, this was only illustrative

PR #248, opened 2021-08-21
[DNM] [txservice] Serialized transaction validation

jakekidd:
- Handle TODOs
- Make a backfill fn for a bunk tx (TODO should mark this)
- Handle dispatch vs provider antipattern. There's two maps in txservice and it's not pretty. We may want to remodel dispatch to either extend ChainRpcProvider or just move all the ChainRpcProvider functionality over here.
- Unit test fixes (make sure you close out any antipatterns first)
- Lots of load testing

### DIFF
--- a/packages/txservice/src/dispatch.ts
+++ b/packages/txservice/src/dispatch.ts
@@ -26,11 +26,11 @@ class TransactionBuffer {
 
   // Move latest pending to completed.
   // @returns boolean whether operation was successful, or if there was no pending tx to shift.
-  public async shift(): Promise<boolean> {
+  public async shift(id: string): Promise<boolean> {
     // TODO: Trim the validated stack when it gets "too long" to save memory space.
     return await this.accessQueue.add((): boolean => {
-      const transaction = this.pending.shift();
-      if (!transaction) {
+      const transaction = this.pending.find((transaction) => transaction.id === id);
+      if (transaction == null) {
         return false;
       }
       this.validated.push(transaction);
@@ -168,29 +168,24 @@ export class TransactionDispatch {
       try {
         await transaction.validate();
         // **shift from pending stack, push to validated stack
-        await this.buffer.shift(transaction.id);
+        const success = await this.buffer.shift(transaction.id);
+        if (!success) {
+          this.logger.warn({
+            id: transaction.id,
+          }, "Failed to locate transaction in pending buffer.");
+        }
       } catch (error) {
-        // timeout? (e.g. after 60sec).
-        // (optional sanity check): getTransaction, check that it exists on chain
-        // doesn't exist?
-        // TODO: backfill with a bunk tx
-        // note: even if the tx is *not* the blockade, we need to backfill this one, or it will block all txs that follow.
-
         if (error instanceof TimeoutError) {
+          // TODO: (optional sanity check): getTransaction, check that it exists on chain
+            // doesn't exist? TODO: backfill with a bunk tx
+            // note: even if the tx is *not* the blockade, we need to backfill this one, or it will block all txs that follow.
+
           // This will bump gas price and append this transaction to the end of the validation queue again.
-          // TODO: nested try catch = not pretty
-          try {
-            transaction.bumpGasPrice();
-            await transaction.submit();
-            // add this tx to the end of the validation queue
-            this.addToValidation(transaction);
-          } catch (e) {
-            // error?
-            // validate that the transaction failed on chain ??
-            // didn't fail on chain? -> WTF
-            // mark tx with error
-            // return;
-          }
+          // TODO: if error, validate that the transaction failed on chain ??
+          transaction.bumpGasPrice();
+          await transaction.submit();
+          // add this tx to the end of the validation queue
+          this.addToValidation(transaction);
         } else {
           throw error;
         }
--- a/packages/txservice/src/dispatch.ts
+++ b/packages/txservice/src/dispatch.ts
@@ -1,74 +1,41 @@
-import { BigNumber, providers, Signer } from "ethers";
+import { BigNumber } from "ethers";
 import { BaseLogger } from "pino";
 import PriorityQueue from "p-queue";
-// import { getUuid } from "@connext/nxtp-utils";
 
-// import { TransactionServiceConfig } from "./config";
 import { ChainRpcProvider } from "./provider";
-import { FullTransaction, GasPrice, WriteTransaction } from "./types";
-import { AlreadyMined, TransactionReplaced, TransactionReverted, TransactionServiceFailure } from "./error";
+import { GasPrice, WriteTransaction } from "./types";
+import { Transaction } from "./transaction";
+import { TimeoutError, TransactionReverted } from "./error";
 
-class MetaTransaction {
-  // Responses, in the order of attempts made for this tx.
-  public responses: providers.TransactionResponse[] = [];
-  // Receipt that we received for the on-chain transaction that was mined with
-  // the desired number of confirmations.
-  public receipt?: providers.TransactionReceipt;
-
-  // Which transaction attempt we are on.
-  private _attempt = 0;
-  /**
-   * Getter to return the internal attempt
-   *
-   * @returns The _attempt of the transction
-   */
-  public get attempt(): number {
-    return this._attempt;
-  }
-
-  /**
-   * Retrieves all data needed to format a full transaction, including current gas price set, nonce, etc.
-   */
-  public get data(): FullTransaction {
-    return {
-      ...this.minTx,
-      gasPrice: this.gasPrice.get(),
-      nonce: this.nonce,
-      gasLimit: this.gasPrice.limit,
-    };
-  }
-
-  /**
-   * A data structure used for management of the lifecycle of one on-chain transaction.
-   *
-   * @param logger The pino.BaseLogger instance we use for logging.
-   * @param provider The ChainRpcProvider instance we use for interfacing with the chain.
-   * @param minTx The minimum transaction data required to send a transaction.
-   * @param gasPrice Initialized gas price tracker for this transaction.
-   */
-  constructor(
-    public readonly minTx: WriteTransaction,
-    public readonly nonce: number,
-    private readonly gasPrice: GasPrice,
-  ) {}
-}
-
+// TODO: sep file
 // Thread safe stack manager.
 class TransactionBuffer {
   // Both of the following are functionally treated as stacks (LIFO).
-  private completed: MetaTransaction[] = [];
-  private pending: MetaTransaction[] = [];
+  private validated: Transaction[] = [];
+  private pending: Transaction[] = [];
   // Should be used to access either of the stacks.
   private readonly accessQueue: PriorityQueue = new PriorityQueue({ concurrency: 1 });
 
   // Push a new tx to pending.
-  public async push() {
+  public async push(transaction: Transaction) {
     // TODO: We may want to cap the pending stack here.
+    await this.accessQueue.add(() => {
+      this.pending.push(transaction);
+    });
   }
 
   // Move latest pending to completed.
-  public async shift() {
-    // TODO: Trim the completed stack.
+  // @returns boolean whether operation was successful, or if there was no pending tx to shift.
+  public async shift(): Promise<boolean> {
+    // TODO: Trim the validated stack when it gets "too long" to save memory space.
+    return await this.accessQueue.add((): boolean => {
+      const transaction = this.pending.shift();
+      if (!transaction) {
+        return false;
+      }
+      this.validated.push(transaction);
+      return true;
+    });
   }
 
   /**
@@ -87,45 +54,58 @@ class TransactionBuffer {
     return (await this.getLastTx())?.nonce;
   }
 
-  private async getLastTx(): Promise<MetaTransaction | null> {
-    return await this.accessQueue.add(() => {
+  private async getLastTx(): Promise<Transaction | null> {
+    return await this.accessQueue.add((): Transaction | null => {
       if (this.pending.length > 0) {
         return this.pending[this.pending.length - 1];
-      } else if (this.completed.length > 0) {
-        return this.completed[this.completed.length - 1];
+      } else if (this.validated.length > 0) {
+        return this.validated[this.validated.length - 1];
       }
       return null;
     });
   }
 }
 
-export class RpcDispatch {
-  private readonly submitQueue: PriorityQueue = new PriorityQueue({ concurrency: 1 });
+// TODO: We may want to remodel dispatch to either extend ChainRpcProvider or just move all the ChainRpcProvider functionality over here.
+// For now, a sep. class for simplicity of introducing new functionality.
+/**
+ * @classdesc Dispatches and monitors transaction queue.
+ */
+export class TransactionDispatch {
   public readonly chainId: number;
 
+  private readonly submitQueue: PriorityQueue = new PriorityQueue({ concurrency: 1 });
+  // A queue for execution of transaction validation. When it completes, we'll shift the transaction from
+  // pending stack -> validated stack in the buffer.
+  private readonly validateQueue: PriorityQueue = new PriorityQueue({ concurrency: 1 });
+  private readonly buffer: TransactionBuffer = new TransactionBuffer();
+
   /**
+   * Centralized transaction dispatch management class.
    *
    * @param logger The pino.BaseLogger instance we use for logging.
    * @param provider The ChainRpcProvider instance we use for interfacing with the chain.
    */
-  constructor(
-    private readonly logger: BaseLogger,
-    public readonly provider: ChainRpcProvider
-  ) {
+  constructor(private readonly logger: BaseLogger, private readonly provider: ChainRpcProvider) {
     this.chainId = this.provider.chainId;
   }
 
-  public async submit(tx: WriteTransaction) {
+  public async submit(minTx: WriteTransaction): Promise<Transaction> {
+    const method = this.submit.name;
+    // get gas estimate
     let gasLimit: BigNumber;
-    let result = await this.provider.estimateGas(tx);
+    let result = await this.provider.estimateGas(minTx);
     if (result.isErr()) {
       if (result.error instanceof TransactionReverted) {
+        // If we get a TransactionReverted error, that means the gas estimate call
+        // indicated our transaction would fail on-chain. The details of the failure will
+        // be included in the error.
         throw result.error;
       }
       this.logger.warn(
         {
-          method: this.submit.name,
-          transaction: tx,
+          method,
+          transaction: minTx,
           error: result.error,
         },
         "Estimate gas failed due to an unexpected error.",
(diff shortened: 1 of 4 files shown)


packages/txservice/src/dispatch.ts
```diff
@@ -0,0 +1,195 @@
...
+    // TODO: Trim the validated stack when it gets "too long" to save memory space.
+    return await this.accessQueue.add((): boolean => {
+      const transaction = this.pending.find((transaction) => transaction.id === id);
+      if (transaction == null) {
+        return false;
+      }
+      this.validated.push(transaction);
+      return true;
+    });
+  }
```
LayneHaber, line 39: We should be deleting completed transactions from memory rather than hanging on to them

packages/txservice/src/dispatch.ts
```diff
@@ -0,0 +1,195 @@
...
+import { TimeoutError, TransactionReverted } from "./error";
+
+// TODO: sep file
+// Thread safe stack manager.
+class TransactionBuffer {
+  // Both of the following are functionally treated as stacks (LIFO).
+  private validated: Transaction[] = [];
+  private pending: Transaction[] = [];
+  // Should be used to access either of the stacks.
+  private readonly accessQueue: PriorityQueue = new PriorityQueue({ concurrency: 1 });
```
LayneHaber, line 17: I don't understand why you need to serialize access so the transactions, or really what the `validated` stack is for?

packages/txservice/src/dispatch.ts
```diff
@@ -0,0 +1,195 @@
...
+  /**
+   * Get the current nonce value.
+   *
+   * @remarks
+   * Caller should still be prepared to get the incorrect nonce back. For instance,
+   * if the provider that just handled our sent tx has suddenly gone offline, this
+   * method may give the wrong nonce. This can be solved by making additional calls to
+   * submit the tx.
+   *
+   * @returns A number value for the current nonce.
```
LayneHaber, line 50: `current` nonce is not specific enough -- do you mean the current nonce that a tx should be submitted at or the current nonce that needs to be cleared

LayneHaber: Seems like (as paired with below) its always the nonce that you will submit a transaction at

packages/txservice/src/dispatch.ts
```diff
@@ -0,0 +1,195 @@
...
+    return await this.accessQueue.add((): Transaction | null => {
+      if (this.pending.length > 0) {
+        return this.pending[this.pending.length - 1];
+      } else if (this.validated.length > 0) {
+        return this.validated[this.validated.length - 1];
+      }
+      return null;
+    });
+  }
+}
```
LayneHaber, line 67: Why does accessing this have to be serialized in a queue?

packages/txservice/src/dispatch.ts
```diff
@@ -0,0 +1,195 @@
...
+/**
+ * @classdesc Dispatches and monitors transaction queue.
+ */
+export class TransactionDispatch {
+  public readonly chainId: number;
+
+  private readonly submitQueue: PriorityQueue = new PriorityQueue({ concurrency: 1 });
+  // A queue for execution of transaction validation. When it completes, we'll shift the transaction from
+  // pending stack -> validated stack in the buffer.
+  private readonly validateQueue: PriorityQueue = new PriorityQueue({ concurrency: 1 });
```
LayneHaber, line 80: We should keep these `submit` / `validate` closer to the nomenclature used when talking about transactions (i.e. `submit` / `mined` / `confirmed`)

packages/txservice/src/dispatch.ts
```diff
@@ -0,0 +1,195 @@
...
+    this.chainId = this.provider.chainId;
+  }
+
+  public async submit(minTx: WriteTransaction): Promise<Transaction> {
+    const method = this.submit.name;
+    // get gas estimate
+    let gasLimit: BigNumber;
+    let result = await this.provider.estimateGas(minTx);
+    if (result.isErr()) {
+      if (result.error instanceof TransactionReverted) {
```
LayneHaber, line 99: `instanceof` can have unpredictable behavior, you'll need to use a different condition here

packages/txservice/src/dispatch.ts
```diff
@@ -0,0 +1,195 @@
...
+        {
+          method,
+          transaction: minTx,
+          error: result.error,
+        },
+        "Estimate gas failed due to an unexpected error.",
+      );
+      throw result.error;
+    } else {
+      gasLimit = result.value;
```
LayneHaber, line 115: the `gasLimit` should always leave some buffer above the estimate instead of being used directly

packages/txservice/src/dispatch.ts
```diff
@@ -0,0 +1,195 @@
...
+        // If transaction buffer returns null, that indicates the buffer is empty; meaning we haven't sent any transactions yet.
+        let nonce = await this.buffer.getNonce();
+        if (!nonce) {
+          const result = await this.provider.getTransactionCount();
+          if (result.isErr()) {
+            throw result.error;
+          } else {
+            nonce = result.value;
+          }
+        }
```
LayneHaber, line 138: This *entire* process should be handled within the buffer imo -- we should always only need to look 1 place for the logic around handling nonces

packages/txservice/src/dispatch.ts
```diff
@@ -0,0 +1,195 @@
...
+        await transaction.validate();
+        // **shift from pending stack, push to validated stack
+        const success = await this.buffer.shift(transaction.id);
+        if (!success) {
+          this.logger.warn({
+            id: transaction.id,
+          }, "Failed to locate transaction in pending buffer.");
+        }
+      } catch (error) {
+        if (error instanceof TimeoutError) {
```
LayneHaber, line 178: `instanceof` behavior issues noted above

packages/txservice/src/provider.ts
```diff
@@ -129,55 +124,7 @@ export class ChainRpcProvider {
...
-   * to read from chain.
-   */
-  public readTransaction(tx: ReadTransaction): ResultAsync<string, TransactionError> {
-    return this.resultWrapper<string>(async () => {
-      try {
-        return await this.signer.call(tx);
-      } catch (error) {
-        throw new TransactionReadError(TransactionReadError.reasons.ContractReadError, { error });
-      }
+      return await this.signer.sendTransaction(transaction);
```
LayneHaber, line 127: Where does this tie into the buffer, if at all?

packages/txservice/src/transaction.ts
```diff
@@ -158,35 +114,39 @@ export class Transaction {
...
     // Increment transaction attempts made.
     this._attempt++;
+    this.timestamp = Date.now();
 
     // Send the tx.
     let result = await this.provider.sendTransaction(this.data);
 
     // If the error is a NonceExpired error and we haven't submitted yet, we want to keep
     // trying to send here. Reason being that it may take a few tries to get the correct
     // transaction count back from the provider.
```
LayneHaber, line 141: Couldn't this also mean you're trying to submit a tx at some nonce `n` that has already been confirmed at least once or has a gas bump in the mempool? This seems like it would be weird on restarts if the signer is frozen

LayneHaber (commented):
This seems slightly more complicated than is necessary and the function of the `TransactionBuffer` class (especially even why access must be serialized) is still unclear to me. It seems there is no point where we are checking against what the provider is saying the blockade is, or any notion of backfilling

packages/txservice/src/dispatch.ts
jakekidd: A: serializing access here was the equivelent of having a "thread-safe" flag for accessing something, which is a common practice with multithreading in other languages
It serializes all reads, because time is sort of of the essence here

That being said, it may be *totally* unnecessary. I think worth trying with and without for sure

B: validated is to provide lookback to get the current nonce. to get the current nonce, we look back through pending and then validated. If nothing, we just call provider.getTransactionCount

packages/txservice/src/dispatch.ts
jakekidd: I 100% agree we should eventually delete them

In fact, it might only be worth hanging onto the "last validated transaction" - see above comment - or maybe even just "last validated nonce"

packages/txservice/src/dispatch.ts
jakekidd: Fuck you right. Should be more specific. We could make this `getLastNonce` or we could just send back `(await this.getLastTx())?.nonce + 1;`

packages/txservice/src/dispatch.ts
jakekidd: minedQueue yeah that works

jakekidd: Although we should *not* broadcast a tx is mined until after it received the desired # of transactions though so...

packages/txservice/src/dispatch.ts
jakekidd: this was in original code so uh. dang that's dangerous. we should purge that from txservice entirely then, my b

packages/txservice/src/dispatch.ts
jakekidd: Sure. again, this was original c/p code. If you could figure out what that buffer should be and/or implement it, that'd be great

packages/txservice/src/dispatch.ts
jakekidd: Sure, agreed. would be cleaner

jakekidd: But should only ever be called here

packages/txservice/src/provider.ts
jakekidd: dispatch handles queuing now. I separated interests regarding queuing. It gave me more space to work lol. we may just wanna integrate dispatch into this class entirely
and maybe just rename this class dispatch, tbh, cause that's what it is

packages/txservice/src/transaction.ts
jakekidd: Yes, this model is entirely incompatible with the new model I have of assigning the nonce once we create the tx

jakekidd: great spot!

packages/txservice/src/transaction.ts
```diff
@@ -258,46 +201,111 @@ export class Transaction {
...
+        previousGasPrice: currentPrice.toString(),
+        newGasPrice: bumpedGasPrice.toString(),
+      },
+      "Bumping tx gas price for reattempt.",
+    );
+  }
+
+  private timeUntilExpiry(): number {
+    const expiry = this.timestamp + this.provider.confirmationTimeout;
+    return expiry - Date.now();
```
jakekidd, line 270: This could be negative, but we shouldn't let that prevent "attempting" to confirm right? Maybe have a minimum of 2-5 seconds just to let validate() attempt to validate

packages/txservice/src/monitor/buffer.ts
```diff
@@ -0,0 +1,49 @@
...
+  private buffer: Map<number, Transaction> = new Map();
+
+  public get pending(): Transaction[] {
+    return this.orderedKeys()
+      .map((nonce) => this.get(nonce) ?? null)
+      .filter(tx => tx != null && !tx.didFinish) as Transaction[];
+  }
+
+  public insert(nonce: number, transaction: Transaction, overwrite = false) {
+    // TODO: detect gap in nonce? throw if there's an attempt to insert out of order?
```
LayneHaber, line 22: Don't think that this is necessarily the responsibility of this class. Would be open to adding it if there were a good way to handle it, which seems like there isn't without going through the loop or adding duplicate logic to determine if you should enter a backfill or something weird

packages/txservice/src/monitor/buffer.ts
```diff
@@ -0,0 +1,49 @@
...
+    // TODO: Trim backend of buffer? Maybe keep a maximum of 1000 transactions?
```
LayneHaber, line 23: Transactions should be deleted from the buffer once they are fully confirmed

packages/txservice/src/monitor/monitor.ts
```diff
@@ -0,0 +1,242 @@
...
+   */
+  public async createTransaction(minTx: WriteTransaction): Promise<Transaction> {
+    // Estimate gas here will throw if the transaction is going to revert on-chain for "legit" reasons. This means
+    // that, if we get past this method, we can safely assume that the transaction will go through on submit, saving for
+    // instances where the provider malfunctions.
+    const gas = await this.getGas(minTx);
+    // Queue up the transaction with these values.
+    const result = await this.queue.add(async (): Promise<{ value: Transaction | Error; success: boolean }> => {
+      try {
+        // NOTE: This call must be here, serialized within the queue, as it is depenedent on current pending transaction count.
```
LayneHaber, line 57: depenedent

packages/txservice/src/monitor/monitor.ts
```diff
@@ -0,0 +1,242 @@
...
+        // Create a new transaction instance to track lifecycle.
+        const transaction = new Transaction(this.logger, this.provider, minTx, nonce, gas);
+        this.buffer.insert(nonce, transaction);
+        return { value: transaction, success: true };
+      } catch (e) {
+        // validate that the transaction failed on chain?
+        // didn't fail on chain? log this occurrence w/ error but ignore error ?!? -> WTF?
+        // if it did fail on chain, THROW
+        return { value: e, success: false };
+      }
```
LayneHaber, line 68: will this only catch the `estimateGas` errors? if so that's fine, just want to clarify

packages/txservice/src/monitor/monitor.ts
```diff
@@ -0,0 +1,242 @@
...
+    }
+  }
+
+  private async getGas(transaction: WriteTransaction): Promise<Gas> {
+    const method = this.getGas.name;
+    // Get gas estimate.
+    let gasLimit: BigNumber;
+    let result = await this.provider.estimateGas(transaction);
+    if (result.isErr()) {
+      if (result.error instanceof TransactionReverted) {
```
LayneHaber, line 83: 😢

packages/txservice/src/monitor/monitor.ts
```diff
@@ -0,0 +1,242 @@
...
+        {
+          method,
+          transaction: transaction,
+          error: result.error,
+        },
+        "Estimate gas failed due to an unexpected error.",
+      );
+      throw result.error;
+    } else {
+      gasLimit = result.value;
```
LayneHaber, line 99: (i know i already mentioned this but i resolved that comment so adding it here)

we should make sure we add some buffer to the gas limit here. like maybe 20% of the estimate or something. cc @rhlsthrm for best practice

packages/txservice/src/monitor/monitor.ts
```diff
@@ -0,0 +1,242 @@
...
+    // we haven't sent any transactions yet, and we'll use this value here.
+    const pending = result.value;
+    // Increment if we got back a value from buffer as we want the value of last tx's nonce + 1.
+    return Math.max(buffer + 1, pending);
+  }
+
+  private async backfillLoop() {
+    // TODO: Make sure this loop is throw-proof
+    while (true) {
+      delay(1000);
```
LayneHaber: why delay? also we should `await` it if its necessary

packages/txservice/src/monitor/monitor.ts
```diff
@@ -0,0 +1,242 @@
...
+        continue;
+      }
+      const result = await this.provider.getTransactionCount();
+      if (result.isErr()) {
+        // TODO: log
+      }
+      const currentNonce = result.isOk() ? result.value : -1;
+      const tx: Transaction | undefined = this.buffer.get(currentNonce);
+      if (tx == null) {
+        // BC it could just be the latest nonce and we're currently not doing anything
```
LayneHaber, line 163: is this comment out of date? it seems the case its referencing is handled above (by checking buffer length)

packages/txservice/src/monitor/monitor.ts
```diff
@@ -0,0 +1,242 @@
...
+        // TODO: backfill
+        this.backfill(currentNonce, undefined, "NOT_FOUND");
```
LayneHaber, line 165: all calls to `backfill` should likely be awaited. otherwise i think were signing ourselves up for lots of replacement underpriced errors

packages/txservice/src/monitor/monitor.ts
```diff
@@ -0,0 +1,242 @@
...
+      } else {
+        if (tx.didFinish || tx.isBackfill) {
+          // IF the transaction did finish already, or this is already being backfilled (from a previous iteration
+          // here), we can ignore this one.
+          // TODO: Do we actually want to proceed even if a tx IS a backfill? What would that accomplish? Because if
+          // even the backfill isn't working, then how would ANOTHER backfill solve anything? In fact, maybe we want to
+          // shut things down if even a backfill tx isn't going through?
```
LayneHaber, line 172: 100% we should shut things down if the backfill tx isn't going through. This means we should make sure that the backfill is both awaited and submitted at a high gas price

packages/txservice/src/monitor/monitor.ts
```diff
@@ -0,0 +1,242 @@
...
+          this.backfill(currentNonce, tx, "EXPIRED");
+        } else {
+          if (tx.attempt > 5) {
+            try {
+              // You get 1 hail mary. (This should throw if it fails).
+              await this.hailMary(tx);
+            } catch (error) {
+              // If it fails from there, kill tx and backfill
+              await tx.kill(error);
+              this.backfill(currentNonce, tx, "TTL_HM_FAILED", { error });
```
LayneHaber, line 187: think this should also be awaited

packages/txservice/src/monitor/monitor.ts
```diff
@@ -0,0 +1,242 @@
...
+      throw addressResult.error;
+    }
+    // Sending a 0 wei transaction to fill the gap. We only have to eat the cost of gas here.
+    const minTx: WriteTransaction = {
+      chainId: this.chainId,
+      value: BigNumber.from(0),
+      data: "0x",
+      to: addressResult.value,
+    };
+    const gas = await this.getGas(minTx);
```
LayneHaber: we should not be resubmitting this backfill tx multiple times. instead we should minimize the number of times were "backfilling" but expedite that as much as possible by sending it at a high gas. we definitely dont want to end up in situations where we are backfilling transactions, it gets added to the buffer, and then we are waiting for the gas to get bumped sufficiently high so it eventually goes through

open to counterarguments though, since obviously the "f it send it at max" isn't an elegant solution

packages/txservice/src/provider.ts
```diff
@@ -113,71 +108,32 @@ export class ChainRpcProvider {
 
     // TODO: #146 We may ought to do this instantiation in the txservice constructor.
     this.signer = typeof signer === "string" ? new Wallet(signer, this.provider) : signer.connect(this.provider);
+
+    // TODO: This is an antipattern, or at least feels like one, in conjuction with the
+    // createTransaction() and sendTransaction() methods below. If sendTransaction() should only ever be called within
+    // Transaction class's submit() method, why should it be exposed below?
```
LayneHaber, line 114: good q

packages/txservice/src/txservice.ts
```diff
@@ -121,57 +120,84 @@ export class TransactionService {
...
+          } else {
             throw error;
-          } else if (error instanceof AlreadyMined) {
-            // If the error is an AlreadyMined error, we probably don't need to log a warning here.
+          }
+        }
+
+        // Validate: wait for 1 confirmation.
+        try {
+          // TODO: Should we emit event to listeners for this step?
```
LayneHaber, line 155: no, we should only emit events when we are certain of the outcome. if we are still waiting for more confirmations before we accept it, we shouldn't emit an external event. if the event is used internally only, thats fine

LayneHaber (commented):
Design looks clean, still lots of todos but stoked to get this tested!

jakekidd:
> This seems slightly more complicated than is necessary and the function of the `TransactionBuffer` class (especially even why access must be serialized) is still unclear to me. It seems there is no point where we are checking against what the provider is saying the blockade is, or any notion of backfilling

Better (in terms of structure) now?

packages/txservice/src/monitor/buffer.ts
jakekidd: That does make sense. I'm worried there's an edge case here *somehow* even though I don't see that being possible, so you're right, lets go with that

packages/txservice/src/monitor/monitor.ts
jakekidd: It is 100% depenedent i swaer

packages/txservice/src/monitor/monitor.ts
jakekidd: Yes, these comments can be removed

packages/txservice/src/monitor/monitor.ts
jakekidd: yep all backfills should, thanks

packages/txservice/src/monitor/monitor.ts
jakekidd: No, it's the solution we definitely want for now. but leaving this comment thread open in case @rhlsthrm has opinions

packages/txservice/src/provider.ts
jakekidd: yep, this is the last "antipattern" related q, and it's kind of a biggie. There's got to be a more elegant way to structure this, but for now, if it works, it works.

packages/txservice/src/txservice.ts
jakekidd: Kk

packages/txservice/src/provider.ts
LayneHaber: Can you explain more about why it *is* exposed after posing the question/describe the antipattern you're concerned about in a bit more depth?

packages/txservice/src/monitor/buffer.ts
jakekidd: Yeah, I mean... maybe we could do a debug log, make sure that monitor is picking up on those gaps... but I think we'll know pretty quickly whether or not monitor is doing its job lol

LayneHaber:
> what do we do if a backfill fails? shut down the whole system? how? lol

Create an evt in the `TransactionMonitor` class that will emit an `abort` event. This sets a flag on the topmost class that means every single request to send a transaction immediately throws a `ServiceAborted` error or something

Also adding in the following:
- [ ] Make sure the `buffer` class has a `poll` event or something watched by the `monitor` that will trigger the looping so we can cut down on provider requests if there aren't any transactions being sent

LayneHaber:
> handle error case: AlreadyMined && attempt === 1

hm when does this case occur again?

packages/txservice/src/monitor/monitor.ts
jakekidd: Currently we are submitting at max and we are awaiting

I think rn the question is how do we shut things down? Do we signal to parent provider class that this isn't working, and we should go down for maintenence?

It may be worth implementing a "non-functioning" hibernation mode for provider. Something that just polls every 10 minutes to see if provider is working again. Idk, future stuff

jakekidd:
> > handle error case: AlreadyMined && attempt === 1
> 
> hm when does this case occur again?

See txservice line 131
Could be implemented pretty easily, just need to implement in a non-ugly way lol

jakekidd:
> > what do we do if a backfill fails? shut down the whole system? how? lol
> 
> Create an evt in the `TransactionMonitor` class that will emit an `abort` event. This sets a flag on the topmost class that means every single request to send a transaction immediately throws a `ServiceAborted` error or something
> 
> Also adding in the following:
> 
> * [ ]  Make sure the `buffer` class has a `poll` event or something watched by the `monitor` that will trigger the looping so we can cut down on provider requests if there aren't any transactions being sent

This is a good solution... I'll go with that. Idk if I'm understanding the buffer poll event though? The looping should always continue so long as there are pending tx's... I guess we could use that event to trigger polling to restart() and stop() if there's no pending left, but that will have same effect as current logic.

packages/txservice/src/monitor/buffer.ts
jakekidd: done

packages/txservice/src/monitor/monitor.ts
jakekidd: added to TODOs above

Issue #176, opened 2021-08-09
[router] Metatx fee validation


rhlsthrm:
Keep a configurable threshold of minimum accepted fee that the router will accept before sending a metatx. Check user sig on this.

Issue #137, opened 2021-08-08
[router] Auction validation


LayneHaber:
- [x] Make sure router `swapPool` includes all assets/chains in swap.
- [x] Get the fee rate from the config.

PR #258, opened 2021-08-23
fee config

sanchaymittal:
closes: #176
closes: #137

### DIFF
--- a/packages/integration/ops/config/router/config.json
+++ b/packages/integration/ops/config/router/config.json
@@ -5,13 +5,17 @@
       "providers": ["http://172.17.0.1:8545"],
       "confirmations": 1,
       "subgraph": "http://172.17.0.1:8010/subgraphs/name/connext/nxtp",
-      "transactionManagerAddress": "0x8CdaF0CD259887258Bc13a92C0a6dA92698644C0"
+      "transactionManagerAddress": "0x8CdaF0CD259887258Bc13a92C0a6dA92698644C0",
+      "feePercentage": 0.05,
+      "safeRelayerFee": 100
     },
     "1338": {
       "providers": ["http://172.17.0.1:8546"],
       "confirmations": 1,
       "subgraph": "http://172.17.0.1:9010/subgraphs/name/connext/nxtp",
-      "transactionManagerAddress": "0x8CdaF0CD259887258Bc13a92C0a6dA92698644C0"
+      "transactionManagerAddress": "0x8CdaF0CD259887258Bc13a92C0a6dA92698644C0",
+      "feePercentage": 0.05,
+      "safeRelayerFee": 100
     }
   },
   "logLevel": "info",
--- a/packages/integration/ops/config/router/config.json
+++ b/packages/integration/ops/config/router/config.json
@@ -6,7 +6,6 @@
       "confirmations": 1,
       "subgraph": "http://172.17.0.1:8010/subgraphs/name/connext/nxtp",
       "transactionManagerAddress": "0x8CdaF0CD259887258Bc13a92C0a6dA92698644C0",
-      "feePercentage": 0.05,
       "safeRelayerFee": 100
     },
     "1338": {
@@ -14,7 +13,6 @@
       "confirmations": 1,
       "subgraph": "http://172.17.0.1:9010/subgraphs/name/connext/nxtp",
       "transactionManagerAddress": "0x8CdaF0CD259887258Bc13a92C0a6dA92698644C0",
-      "feePercentage": 0.05,
       "safeRelayerFee": 100
     }
   },
--- a/packages/router/src/config.ts
+++ b/packages/router/src/config.ts
@@ -25,9 +25,6 @@ import {
 import { config as dotenvConfig } from "dotenv";
 import { fetchJson } from "ethers/lib/utils";
 
-const minimumFeePercentageAllowed = 0; // 0.00%
-const maximumFeePercentageAllowed = 15; // 15.00%
-
 const MIN_GAS = utils.parseEther("0.1");
 
 dotenvConfig();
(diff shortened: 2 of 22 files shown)


packages/router/src/config.ts
```diff
@@ -25,10 +25,18 @@ import {
 import { config as dotenvConfig } from "dotenv";
 import { fetchJson } from "ethers/lib/utils";
 
+const minimumFeeAllowed = 0;
```
rhlsthrm, line 28: This name is confusing, what fee?

packages/router/src/config.ts
```diff
@@ -112,7 +123,7 @@ export const getEnvConfig = (chainData: Map<string, any> | undefined): NxtpRoute
     let json: string;
 
     if (process.env.NXTP_CONFIG_FILE) {
-      console.log("process.env.NXTP_CONFIG_FILE: ", process.env.NXTP_CONFIG_FILE);
+      console.info("process.env.NXTP_CONFIG_FILE: ", process.env.NXTP_CONFIG_FILE);
```
rhlsthrm, line 126: Remove this console log completely.

packages/router/src/config.ts
```diff
@@ -186,9 +197,9 @@ export const getEnvConfig = (chainData: Map<string, any> | undefined): NxtpRoute
     // format: { [chainId]: { [chainName]: { "contracts": { "TransactionManager": { "address": "...." } } } }
     if (!chainConfig.transactionManagerAddress) {
       try {
-        nxtpConfig.chainConfig[chainId].transactionManagerAddress = (
-          Object.values((contractDeployments as any)[chainId])[0] as any
-        ).contracts.TransactionManager.address;
+        nxtpConfig.chainConfig[chainId].transactionManagerAddress = (Object.values(
```
rhlsthrm, line 200: We should probably make this a `util` since we use it in `test-ui` also.

packages/router/src/lib/operations/auction.ts
```diff
@@ -155,6 +139,24 @@ export const newAuction = async (
     });
   }
 
+  // getting the swap rate from the receiver side config
+  const swapRate = (1 - parseFloat(config.chainConfig[receivingChainId].feePercentage.toString()) / 100).toString();
```
rhlsthrm, line 143: Swap rate and fee are two different things. We need to do the following:
- Get swap rate from AMM (leave a placeholder for this, hardcoded to `1`).
- Add additional `0.05%` fee hardcoded.

packages/router/src/config.ts
```diff
@@ -25,10 +25,18 @@ import {
 import { config as dotenvConfig } from "dotenv";
 import { fetchJson } from "ethers/lib/utils";
 
+const minimumFeeAllowed = 0;
+const maximumFeeAllowed = 15;
+
+const minimumRelayerFeeAllowed = 0;
+const maximumRelayerFeeAllowed = 5;
```
rhlsthrm, line 32: Why is there a max relayer fee?

packages/router/src/config.ts
```diff
@@ -63,6 +71,11 @@ export const TChainConfig = Type.Object({
   subgraph: Type.String(),
   transactionManagerAddress: Type.String(),
   minGas: Type.String(),
+  feePercentage: Type.Number({ minimum: minimumFeeAllowed, maximum: maximumFeeAllowed }),
```
rhlsthrm, line 74: Fee percentage should not come from here, the swap rate comes from AMM.

packages/router/src/config.ts
```diff
@@ -63,6 +71,11 @@ export const TChainConfig = Type.Object({
...
+  safeRelayerFeePercentage: Type.Number({
```
rhlsthrm, line 75: This should not be in here. The relayer fee will not be a percentage. It will simply be the gas fee + some additional cost (probably in USD).

packages/router/src/lib/helpers/prepare.ts
```diff
@@ -28,19 +27,24 @@ export const validExpiryBuffer = (buffer: number) => buffer > ONE_DAY_IN_SECONDS
...
+  amount: string,
+  inputDecimals: number,
+  swapRate: string,
+  outputDecimals: number,
+): string => {
   if (amount.includes(".")) {
     throw new AmountInvalid(amount);
   }
-  return calculateExchangeWad(BigNumber.from(amount), inputDecimals, SWAP_RATE, outputDecimals).toString();
+  return calculateExchangeWad(BigNumber.from(amount), inputDecimals, swapRate, outputDecimals).toString();
```
rhlsthrm, line 47: First calculate with swap rate from AMM (hardcoded to 1). Then subtract flat 0.05% fee.

rhlsthrm (changes requested):
Needs some changes.

packages/integration/ops/config/router/config.json
```diff
@@ -5,13 +5,17 @@
       "providers": ["http://172.17.0.1:8545"],
       "confirmations": 1,
       "subgraph": "http://172.17.0.1:8010/subgraphs/name/connext/nxtp",
-      "transactionManagerAddress": "0x8CdaF0CD259887258Bc13a92C0a6dA92698644C0"
+      "transactionManagerAddress": "0x8CdaF0CD259887258Bc13a92C0a6dA92698644C0",
+      "feePercentage": 0.05,
```
rhlsthrm, line 9: This 0.05% fee should not be allowed to be configured by users. The idea is this is a flat fee across the network.

rhlsthrm (changes requested):
Almost there but still don't want to make the flat fee configurable.

packages/integration/ops/config/router/config.json
sanchaymittal: Router should be able to change the fee they want charge.

rhlsthrm (approved):

PR #277, opened 2021-08-27
Added Build Integration Test Script

base0010:


rhlsthrm (changes requested):
Should not push the deleted typechain stuff. Run a `yarn build:all` again it should regenerated them.

rhlsthrm (approved):

PR #287, opened 2021-08-30
282 cache auction bids

### LINKED ISSUE #282: [router] Cache outstanding auction bids

rhlsthrm:
Fixes #282

﻿- Types fixes
- Cache structure
- More cache stuff
- Redis stuff
- Make tests work
- Remove memory stuff
- Tests


packages/router/src/adapters/redisCache/cache.ts
```diff
@@ -0,0 +1,40 @@
...
+  // set the key to expire in the given number of seconds
+  const key = `outstanding-liquidity:${chainId}:${assetId}:${transactionId}`;
+  await redis.setex(key, expiresInSeconds, amount.toString());
+};
+
+export const removeOutstandingLiquidity = async (r: RemoveOutstandingLiquidityParams): Promise<void> => {
+  const redis = getRedis();
+  const { transactionId, assetId, chainId } = r;
+
+  const key = `outstanding-liquidity:${chainId}:${assetId}:${transactionId}`;
```
LayneHaber, line 38: Why is the `transactionId` in the key?

packages/router/src/adapters/redisCache/index.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export const getRedis = (): Redis => redis;
+
+export const getCache = async (): Promise<Cache> => {
+  const { config } = getContext();
+  if (config.redisUrl) {
+    redis = new RedisClient(config.redisUrl);
+  } else {
+    redis = new RedisMock();
+  }
```
LayneHaber, line 21: There are some commands / features that the redis mock hasn't yet implemented. Think well need to test this with high rates of spammy quotes to see how the mock performs before saying okay to this as a solution

packages/router/src/adapters/subgraph/index.ts
```diff
@@ -37,7 +16,7 @@ export const getSdks = (): Record<number, Sdk> => {
   return sdks;
 };
 
-export const subgraphContractReader = (): ContractReader => {
+export const getSubgraphContractReader = async (): Promise<ContractReader> => {
```
LayneHaber, line 19: If we're renaming it why not rename it to `getContractReader`

packages/router/src/lib/entities/cache.ts
```diff
@@ -0,0 +1,23 @@
...
+  transactionId: string;
+  assetId: string;
+  chainId: number;
+};
+
+export type Cache = {
+  storeOutstandingLiquidity: (s: StoreOutstandingLiquidityParams) => Promise<void>;
+  getOutstandingLiquidity: (g: GetOutstandingLiquidityParams) => Promise<BigNumber>;
+  removeOutstandingLiquidity: (r: RemoveOutstandingLiquidityParams) => Promise<void>;
+};
```
LayneHaber, line 23: How are you using this type to check if you have the liquidity to respond to an auction? Seems like we'd have to pull all of them, calculate funds, then we'd know how much we have left to offer. Don't love the "pull all of them" step

packages/router/src/adapters/redisCache/cache.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  // use SCAN to get all the keys for the given assetId and chainId
+  const [, keys] = await redis.scan(0, "match", `outstanding-liquidity:${chainId}:${assetId}*`);
+
+  // use a set to get rid of duplicates, apparently SCAN can return duplicated
+  // https://redis.io/commands/scan#scan-guarantees
+  const amounts = await redis.mget([...new Set(keys)]);
+  const num = amounts.filter((x) => !!x).reduce((acc, amount) => acc.add(BigNumber.from(amount)), constants.Zero);
+  return num;
+};
```
LayneHaber, line 23: How efficient is the `scan` at high load?

packages/router/src/lib/operations/auction.ts
```diff
@@ -163,20 +163,45 @@ export const newAuction = async (
...
+    chainId: receivingChainId,
+  });
+  logger.info(
+    { method, methodId, requestContext, balance: outstandingLiquidity.toString() },
+    "Got available liquidity from contract reader",
+  );
+
+  availableLiquidity = availableLiquidity.sub(outstandingLiquidity);
+
+  if (availableLiquidity.lt(amountReceived)) {
```
LayneHaber, line 184: We'll need to configure this properly. It won't work with lifi nxtp as-is, and will quickly cause the router to erroneously report that there's not enough liquidity. load tests will not catch this because they request a quote 1x per transfer

packages/router/src/adapters/redisCache/cache.ts
rhlsthrm: So it can be deleted by `transactionId`.

packages/router/src/adapters/redisCache/index.ts
rhlsthrm: We are using the most highly used, basic Redis commands which are most definitely the first ones that they have implemented so I'm not worried if our functionality is there. I agree we need to load test with spammy quotes, but I don't see the connection between spammy quotes and the mock.

packages/router/src/adapters/subgraph/index.ts
rhlsthrm: Because it's specifically a subgraph implementation of contract reader. The adapters should specify what type of real-world implementation it is.

packages/router/src/lib/entities/cache.ts
rhlsthrm: The function hides the implementation of how you calculate the outstanding liquidity. I'm not sure I understand exactly what your concern is, but you can see from the implementation of the `getOutstandingLiquidity` function that it does pull all the saved bids and add them up.

packages/router/src/adapters/redisCache/cache.ts
rhlsthrm: Redis is built to do this type of stuff. I did a bunch of research to find this strategy, which is basically built around this exact use case of array elements that individually expire.

packages/router/src/lib/operations/auction.ts
rhlsthrm: Yes, we need to change the LiFi implementation to match this, but it seems like this is how we need the functionality to be implemented on our end.

LayneHaber: What if it's a dry run?

packages/router/src/adapters/redisCache/index.ts
LayneHaber: It's just a note about the use of the mock in prod in general for future compatibility/ functionality

packages/router/src/lib/operations/auction.ts
rhlsthrm: Dry run will not remove outstanding liquidity. That was the idea behind dry run to begin with. I think this is a bug, thanks for pointing out.

packages/router/src/adapters/redisCache/index.ts
rhlsthrm: We definitely aren't using the mock in prod, all deployments will definitely have full Redis instances.

LayneHaber: Ah got it, I misunderstood at the standup

PR #294, opened 2021-08-31
286 test ui

sanchaymittal:
## Checklist

- [ ] Update CHANGELOG.md.


packages/test-ui/src/config.ts
```diff
@@ -0,0 +1,16 @@
+import { SwapConfig } from "./constants";
+
+export const swapConfig: SwapConfig[] = [
```
rhlsthrm, line 3: Why is this committed? We don't want these coins in our repo. This should be part of .env and the UI should allow you to manually input token addresses.

rhlsthrm (changes requested):
Not sure I understand the fix, can you post screenshots of the UX?

packages/test-ui/src/config.ts
sanchaymittal: Inputting custom addresses was getting a cumbersome process at the front-end for testing. Cause we would need two fields to enter token addresses for sending and receiving chain. And the way we read our chainId is from the swap config so the code will need more unwanted changes which potentially can create more bugs.

Exposing the config in code is the best UX we can offer to the router hosts to quickly test things.

rhlsthrm: The whole idea of this task is for the router operators to not need to run their own test-ui. I wanted to find a way to make this as easy as possible, of course, it will still be cumbersome. I think one idea is:
- Drop down on Asset, choose from configured asset, or else choose "Custom".
- If "Custom", allow them to put in their own sending and receiving asset IDs.

Don't worry about chains, the UI will by default support all the chains.

rhlsthrm (approved):

PR #300, opened 2021-09-01
[test-ui] add router balances table

jakekidd:
## The Problem

Need to see router balances for all stable coins across chains.

## The Solution

Display router balances in a table w/ a "Reload" button.

## Checklist

- [ ] Run load tests.
- [ ] Update CHANGELOG.md.

### DIFF
--- a/packages/test-ui/src/components/Router.tsx
+++ b/packages/test-ui/src/components/Router.tsx
@@ -1,4 +1,4 @@
-import { Button, Checkbox, Col, Form, Input, Row, Typography } from "antd";
+import { Button, Checkbox, Col, Form, Input, Row, Typography, Table, Divider, Tooltip } from "antd";
 import { BigNumber, constants, Contract, providers, Signer, utils } from "ethers";
 import { ReactElement, useEffect, useState } from "react";
 import { ChainData, ERC20Abi } from "@connext/nxtp-utils";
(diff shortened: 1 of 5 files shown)


packages/test-ui/src/components/Router.tsx
```diff
@@ -14,9 +14,41 @@ type RouterProps = {
...
+    usdc: "0x2791bca1f2de4661ed88a30c99a7a9449aa84174",
+    usdt: "0xc2132d05d31c914a87c6611c10748aeb04b58e8f",
+    dai: "0x8f3cf7ad23cd3cadbd9735aff958023239c6a063",
+  },
+  250: {
+    usdc: "0x04068da6c83afcfa0e13ba15a6696662335d5b75",
+    usdt: "0x049d68029688eabf473097a2fc38ef61633a3c7a",
+    dai: "0x8d11ec38a3eb5e956b052f67da8bdc9bef8abf3e",
+  },
+};
```
LayneHaber, line 38: if we have the names, can we pull this from the `chainData`? may be easier in the future if people want to check for a list of names too

packages/test-ui/src/components/Router.tsx
```diff
@@ -84,16 +113,69 @@ export const Router = ({ web3Provider, signer, chainData }: RouterProps): ReactE
...
-    form.setFieldsValue({
-      currentLiquidity: liquidity,
-    });
+    const liquidity = await getLiquidity(form.getFieldValue("assetId"));
     return liquidity;
   };
 
+  // Refreshes the balances table with human readable values for each asset on current chain.
+  const refreshBalances = async (): Promise<void> => {
+    if (!routerAddress || !(routerAddress.length === 42)) {
```
LayneHaber, line 122: you can do `isValidAddress` from ethers here instead

packages/test-ui/src/components/Router.tsx
```diff
@@ -84,16 +113,69 @@ export const Router = ({ web3Provider, signer, chainData }: RouterProps): ReactE
...
+      }
+      const txm = new Contract(_txManager.address, _txManager.abi, provider);
+      const assets = ASSETS[chainId];
+      for (const [name, assetId] of Object.entries(assets)) {
+        try {
+          // TODO: redundant code with getLiquidity and getDecimals
+          const liquidity = await txm.routerBalances(routerAddress, assetId);
+          const token = new Contract(assetId, ERC20Abi, provider);
+          const _decimals = decimals[assetId.toLowerCase()] ? decimals[assetId.toLowerCase()] : await token.decimals();
+          decimals[assetId.toLowerCase()] = _decimals;
```
LayneHaber, line 152: native asset cases here

LayneHaber (changes requested):
Should use `chainData` more if possible, and need to handle the native asset cases. otherwise yes, you're doing it right :)

packages/test-ui/src/components/Router.tsx
jakekidd: is that in chaindata? lol lemme see

jakekidd: oh we do, nice. yep will fix

packages/test-ui/src/components/Router.tsx
jakekidd: kk

LayneHaber (approved):

PR #304, opened 2021-09-01
Signature sanitized

LayneHaber:
## The Problem

- some signatures had invalid `v` values which were reverting onchain but handled offchain

## The Solution

- sanitize signatures before returning

## Checklist

- [ ] Run load tests.
- [x] Update CHANGELOG.md.


jakekidd (approved):
good logic lgtm

PR #319, opened 2021-09-08
add: sanitation at router

sanchaymittal:


packages/router/src/adapters/contract/contract.ts
```diff
@@ -18,6 +28,76 @@ export const getContractAddress = (chainId: number): string => {
 export const getTxManagerInterface = () =>
   new Interface(TransactionManagerArtifact.abi) as TTransactionManager["interface"];
 
+export const prepareSanitationCheck = async (
+  chainId: number,
+  nxtpContractAddress: string,
+  invariantTransactionData: InvariantTransactionData,
+) => {
```
LayneHaber, line 35: Do we really want to add in an onchain transaction for each active transaction loop? This is going to make our provider calls skyrocket and seems an inelegant solution to problems with the subgraph

packages/router/src/adapters/subgraph/subgraph.ts
```diff
@@ -227,6 +218,14 @@ export const getActiveTransactions = async (_requestContext?: RequestContext): P
...
 
           if (!receiving) {
+            // if not synced, cancel
+            if (!synced) {
+              return {
+                crosschainTx: sdkSenderTransactionToCrosschainTransaction(senderTx),
+                payload: {},
+                status: CrosschainTransactionStatus.ReceiverNotConfigured,
+              } as ActiveTransaction<"ReceiverNotConfigured">;
+            }
```
LayneHaber, line 228: This should really have a separate status that is handled that way. The subgraph being unsynced is not the same as the `ReceiverNotConfigured` status, which is otherwise only given to transactions where the `transaction.receivingChainId` is not configured (see [here](https://github.com/connext/nxtp/blob/b20e9b8ec68b21b4bd2ee1b836411373e6ceaab8/packages/router/src/adapters/subgraph/subgraph.ts#L131-L157))

LayneHaber (commented):
Think overall this contains inelegant solutions to two underlying problems:

- the `status` should be unique if the subgraph is unsynced, and handled separately
- if the subgraph is unreliable, we should address that rather than adding chain reads

Since this is a needed hot fix to address issues on prod, I'm fine with merging, but this solution does not address the core problems

packages/router/src/adapters/contract/contract.ts
rhlsthrm: It's one provider call per tx per loop which shouldnt be that much but I agree. We may not even need this anymore.

packages/router/src/adapters/subgraph/subgraph.ts
rhlsthrm: Agree, was too lazy to add another status.

PR #366, opened 2021-09-21
auction request limit

wanglonghong:
- limits auction request
- create and update testcase for auction rate

## Checklist

- [x] Run load tests.
- [x] Update CHANGELOG.md.

### DIFF
--- a/packages/router/src/bindings/messaging/auctionRequest.ts
+++ b/packages/router/src/bindings/messaging/auctionRequest.ts
@@ -8,10 +8,9 @@ import {
 } from "@connext/nxtp-utils";
 
 import { ProvidersNotAvailable } from "../../lib/errors";
-import { getNtpTimeSeconds } from "../../lib/helpers";
 import { getOperations } from "../../lib/operations";
 import { getContext } from "../../router";
-import { attemptedAuction, AUCTION_REQUEST_LIMIT, lastAuctionMap } from "../metrics";
+import { attemptedAuction } from "../metrics";
 
 export const auctionRequestBinding = async (
   from: string,
@@ -46,14 +45,6 @@ export const auctionRequestBinding = async (
   // On every new auction broadcast, route to the new auction handler
   logger.info("Received auction request", requestContext, methodContext);
   const { bid, bidSignature } = await newAuction(data, requestContext);
-
-  // check if new auction happens in 5s or not
-  const lastAttemptTime = lastAuctionMap.get(`${bid.user}-${bid.router}`) as number;
-  const currentTime = await getNtpTimeSeconds();
-  if (lastAttemptTime && lastAttemptTime + AUCTION_REQUEST_LIMIT > currentTime) {
-    return;
-  }
-  lastAuctionMap.set(`${bid.user}-${bid.router}`, currentTime);
 
   await messaging.publishAuctionResponse(from, inbox, { bid, bidSignature });
   attemptedAuction.inc({
--- a/packages/router/src/bindings/messaging/auctionRequest.ts
+++ b/packages/router/src/bindings/messaging/auctionRequest.ts
@@ -8,9 +8,10 @@ import {
 } from "@connext/nxtp-utils";
 
 import { ProvidersNotAvailable } from "../../lib/errors";
+import { getNtpTimeSeconds } from "../../lib/helpers";
 import { getOperations } from "../../lib/operations";
 import { getContext } from "../../router";
-import { attemptedAuction } from "../metrics";
+import { attemptedAuction, AUCTION_REQUEST_LIMIT, lastAuctionMap } from "../metrics";
 
 export const auctionRequestBinding = async (
   from: string,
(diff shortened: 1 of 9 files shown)


packages/router/test/bindings/messaging/auctionRequest.spec.ts
```diff
@@ -82,25 +82,53 @@ describe("auctionRequestBinding", () => {
     expect(messagingMock.publishAuctionResponse).to.be.calledOnceWith(from, inbox, { bid, bidSignature });
   });
 
+  it("should work if new auction happens in longer than 5s", async () => {
+    await new Promise((resolve) => {
```
rhlsthrm, line 86: I think we should overwrite the lastAuctionMap in the tests instead of waiting 5 seconds for each test.

packages/router/src/bindings/messaging/auctionRequest.ts
```diff
@@ -41,9 +42,19 @@ export const auctionRequestBinding = async (
     logger.error("No data in auction request", requestContext, methodContext, err);
     return;
   }
+
   // On every new auction broadcast, route to the new auction handler
   logger.info("Received auction request", requestContext, methodContext);
   const { bid, bidSignature } = await newAuction(data, requestContext);
+
+  // check if new auction happens in 5s or not
+  const lastAttemptTime = lastAuctionMap.get(`${bid.user}-${bid.router}`) as number;
```
rhlsthrm, line 51: Doesn't need to be keyed by `user-router`, we should key it by `user-sendingAsset-sendingChain-receivingAsset-receivingChain. We want to avoid a user spamming auctions for a swap combo, but for different swaps it's okay.

rhlsthrm (changes requested):

packages/router/src/bindings/messaging/auctionRequest.ts
```diff
@@ -41,9 +42,19 @@ export const auctionRequestBinding = async (
...
+  const currentTime = await getNtpTimeSeconds();
+  if (lastAttemptTime && lastAttemptTime + AUCTION_REQUEST_LIMIT > currentTime) {
```
rhlsthrm, line 53: I think this code belongs in `operations/auction.ts` since it's a part of the auction business logic.

packages/router/test/bindings/messaging/auctionRequest.spec.ts
wanglonghong: Ok. I agree

wanglonghong:
In commit a3ae18deff0f77ffbba62fbe1e6fd98dbf17a511, I fixed above issues.

rhlsthrm (changes requested):
This looks good. One last thing, let's make the rate limit configurable just in case. Can you add it as a router config item in `config.ts` then default it to `2 seconds`? If they put in 0 seconds then we can skip this process.

wanglonghong:
I agree

wanglonghong:
Fixed in commit b67ebf8bb3beef07b357858cd919b8eab794aa31

packages/router/src/config.ts
```diff
@@ -28,6 +28,7 @@ import contractDeployments from "@connext/nxtp-contracts/deployments.json";
 const MIN_GAS = utils.parseEther("0.1");
 const MIN_RELAYER_FEE = "0"; // relayerFee is in respective chain native asset unit
 const MIN_SUBGRAPH_SYNC_BUFFER = 25;
+export const AUCTION_REQUEST_LIMIT = 5;
```
rhlsthrm, line 31: Sorry this was not clear. We need this to be part of the router config object below, added to `NxtpRouterConfigSchema` and parsed in `getEnvConfig`.

rhlsthrm (changes requested):

rhlsthrm (changes requested):
Sorry, couple more things.
- Let's make the config params in ms instead of seconds to make it consistent with everything else.
- We need to update the docs to reflect the new config: https://github.com/connext/documentation/blob/main/docs/Routers/Reference/configuration.md

wanglonghong:
> Sorry, couple more things.
> 
> * Let's make the config params in ms instead of seconds to make it consistent with everything else.
> * We need to update the docs to reflect the new config: 

No problem

packages/router/src/lib/operations/auction.ts
```diff
@@ -90,8 +91,22 @@ export const newAuction = async (
     });
   }
 
-  // Validate expiry is valid (greater than current time plus a buffer).
   const currentTime = await getNtpTimeSeconds();
+  console.log("config = ", config);
```
rhlsthrm, line 95: Remove console.log

rhlsthrm (changes requested):
Final fix to remove console.log, then we can mark this as ready!

rhlsthrm (approved):
Good to go! Mark it as ready when you're ready and I'll merge it!

PR #367, opened 2021-09-21
[router] Estimates Gas and Includes in fee.

wanglonghong:


rhlsthrm (changes requested):
This looks great so far! Can you add to `deploy.ts` so that we are able to deploy it?

wanglonghong:
> This looks great so far! Can you add to `deploy.ts` so that we are able to deploy it?

I've added to deploy.ts at commit b7123fe1b832878461b286878573fe2331002be2

rhlsthrm (changes requested):
Amazing work on this! The only change request is to move the ABI logic out of the tx service and into the router app itself.

wanglonghong:
> Amazing work on this! The only change request is to move the ABI logic out of the tx service and into the router app itself.

. I updated in commit 9d99665759fbbf3eacfefdae815fe17ce538b93d.

packages/router/src/adapters/contract/contract.ts
```diff
@@ -28,9 +30,20 @@ export const getContractAddress = (chainId: number): string => {
   return nxtpContractAddress;
 };
 
+export const getOracleContractAddress = (chainId: number): string => {
+  const { config } = getContext();
+  const oracleContractAddress = config.chainConfig[chainId]?.priceOracleAddress;
+  if (!oracleContractAddress) {
+    throw new Error(`No price oracle contract exists for chain ${chainId}`);
```
rhlsthrm, line 37: Can we make a named error for this? Use the structure of the other errors and put them in the domain.

packages/router/src/lib/helpers/shared.ts
```diff
@@ -1,8 +1,95 @@
...
+export const getGasPrice = async (chainId: number, requestContext: RequestContext): Promise<BigNumber> => {
+  const { txService } = getContext();
+  const gasPrice = await txService.getGasPrice(chainId, requestContext);
+  return gasPrice;
+};
+
+/**
+ * Gets chain id to take fee from
+ */
+export const getChainIdForGasFee = () => {
```
rhlsthrm, line 93: This probably should return an array of chainIds so we can get gas fees on multiple chains.

packages/router/src/lib/helpers/shared.ts
```diff
@@ -1,8 +1,95 @@
...
+export const calculateGasFeeInReceivingToken = async (
+  sendingAssetId: string,
+  sendingChainId: number,
+  receivingAssetId: string,
+  receivingChainId: number,
+  outputDecimals: number,
+  requestContext: RequestContext,
+): Promise<BigNumber> => {
+  const chaindIdForGasFee = getChainIdForGasFee();
+  if (chaindIdForGasFee != sendingChainId && chaindIdForGasFee != receivingChainId) return constants.Zero;
```
rhlsthrm, line 32: Need to handle the case where there are multiple chainIds for gas fees. Also we should add the gas fee estimation to rinkeby and goerli if we can so we can test this on testnet.

rhlsthrm (changes requested):
Looks good, added some requests. Confused why there are so many txservice diffs? There shouldn't be any change from main branch on txservice right?

wanglonghong:
> Looks good, added some requests. Confused why there are so many txservice diffs? There shouldn't be any change from main branch on txservice right?

I've only added the function to calculate gas price on txservice. others are same.

packages/router/src/adapters/contract/contract.ts
wanglonghong: Yes, we can

packages/router/src/lib/helpers/shared.ts
wanglonghong: got it

packages/contracts/contracts/interfaces/IERC20Extended.sol
```diff
@@ -0,0 +1,95 @@
+pragma solidity ^0.8.4;
```
rhlsthrm, line 1: I think this interface is already part of the openzeppelin contract package we have right?

packages/integration/test/load/txserviceConcurrency.ts
```diff
@@ -1,18 +1,24 @@
 import pino from "pino";
```
rhlsthrm, line 1: Are the changes in this file from your PR?

packages/integration/test/utils/config.ts
```diff
@@ -40,12 +41,14 @@ const DEFAULT_LOCAL_CONFIG = {
       confirmations: 1,
       subgraph: "http://localhost:8010/subgraphs/name/connext/nxtp",
       transactionManagerAddress: "0x8CdaF0CD259887258Bc13a92C0a6dA92698644C0",
+      priceOracleAddress: "0x0000000000000000000000000000000000000000",
```
rhlsthrm, line 44: Do we have a solution for local testnet testing? Does AddressZero mock things out?

packages/router/src/lib/helpers/auction.ts
```diff
@@ -8,3 +8,5 @@ export const AUCTION_EXPIRY_BUFFER = 5 * 60;
  * @returns Expiry time of a given bid in s
  */
 export const getBidExpiry = (currentTime: number) => currentTime + 60 * 5;
+
+export const FEE_CHAIN_IDS = [1, 3, 4, 5];
```
rhlsthrm, line 12: If we do this we need to add to deploy script, right?

rhlsthrm (changes requested):
Still a lot of txservice changes for some reason. You might need to copy the files manually from `main`.

packages/contracts/contracts/interfaces/IERC20Extended.sol
wanglonghong: No. I don't think there is an interface with DECIAML parameters.

packages/integration/test/load/txserviceConcurrency.ts
wanglonghong: Honestly I didn't work on txserviceConcurrency.ts. I just copied and pasted to my repo to resolve conflicts.

packages/integration/test/utils/config.ts
wanglonghong: I think It is impossible to have a local testnet testing.

packages/router/src/lib/helpers/auction.ts
wanglonghong: right

packages/contracts/contracts/interfaces/IERC20Extended.sol
rhlsthrm: You're right, that's very silly.

packages/integration/test/load/txserviceConcurrency.ts
rhlsthrm: Did you copy paste from `main`? It looks like it has changes from main.

packages/integration/test/utils/config.ts
rhlsthrm: Hmm can we do it so that routers do not need to update the config? We can use our deployed priceOracle, and if it's not deployed that means there is no gas fee? That way we can simply use the existence of the contract to check for the chain.

rhlsthrm (approved):
Excellent work! This looks like it is good to go!

Issue #343, opened 2021-09-16
[router] Keep track of prepared/fulfilled/resolved txs in memory until subgraph catches up.


rhlsthrm:
The issue:
- Router gets subgraph data saying tx needs to be prepared (for example).
- Router prepares on the receiver side.
- Subgraph is behind on the sender side.
- Next subgraph loop, the router again gets subgraph data saying tx needs to be prepared.
- Tries to prepare again, will error at the gas estimation step.

Solution:
- On each router operation, put into a cache with the transaction block number.
- Remove from the cache when subgraph synced block is the same as the cached block.

PR #368, opened 2021-09-21
subgraph buffer

sanchaymittal:
closes: #343

### DIFF
--- a/packages/router/src/bindings/contractReader/index.ts
+++ b/packages/router/src/bindings/contractReader/index.ts
@@ -6,7 +6,7 @@ import {
   RequestContextWithTransactionId,
   safeJsonStringify,
 } from "@connext/nxtp-utils";
-import { BigNumber } from "ethers";
+import { BigNumber, providers } from "ethers";
 
 import { getContext } from "../../router";
 import {
@@ -14,7 +14,7 @@ import {
   CrosschainTransactionStatus,
   FulfillPayload,
   PreparePayload,
-  TCrosschainTransactionStatus,
+  Tracker,
 } from "../../lib/entities";
 import { getOperations } from "../../lib/operations";
 import { ContractReaderNotAvailableForChain } from "../../lib/errors";
@@ -39,7 +39,7 @@ import {
 const LOOP_INTERVAL = 15_000;
 export const getLoopInterval = () => LOOP_INTERVAL;
 
-export const handlingTracker: Map<string, TCrosschainTransactionStatus> = new Map();
+export const handlingTracker: Map<string, Tracker> = new Map();
 
 export const bindContractReader = async () => {
   const { contractReader, logger } = getContext();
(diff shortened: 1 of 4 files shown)


packages/router/src/bindings/contractReader/index.ts
```diff
@@ -105,13 +138,13 @@ export const handleSingle = async (
         requestContext,
       });
     }
-    const senderReceipt = await txService.getTransactionReceipt(
+    const senderPerpareReceipt = await txService.getTransactionReceipt(
```
jakekidd, line 141: senderPerpare lol

packages/router/src/bindings/contractReader/index.ts
```diff
@@ -72,21 +84,41 @@ export const handleActiveTransactions = async (transactions: ActiveTransaction<a
...
+      transaction.status === CrosschainTransactionStatus.SenderPrepared ||
+      transaction.status === CrosschainTransactionStatus.SenderExpired
+    ) {
+      chainId = transaction.crosschainTx.invariant.receivingChainId;
+    } else {
+      chainId = transaction.crosschainTx.invariant.sendingChainId;
+    }
+
+    handlingTracker.set(transaction.crosschainTx.invariant.transactionId, {
+      blockNumber: 0,
```
jakekidd, line 102: block number 0?

jakekidd: Oh I see, you set the block number below

If you're going to set a "substitute" or fake initial block number first, can we make it -1 so it's obvious it's just not set yet?

packages/router/src/bindings/contractReader/index.ts
```diff
@@ -72,21 +84,41 @@ export const handleActiveTransactions = async (transactions: ActiveTransaction<a
...
+      chainId,
+    });
+    const res = await handleSingle(transaction, requestContext);
+    if (res) {
+      handlingTracker.set(transaction.crosschainTx.invariant.transactionId, {
+        blockNumber: res.blockNumber,
+        chainId,
+      });
+    } else {
+      handlingTracker.delete(transaction.crosschainTx.invariant.transactionId);
```
jakekidd, line 112: Why are we delaying when there isn't a `res` ?

packages/router/src/bindings/contractReader/index.ts
jakekidd: nit

packages/router/src/bindings/contractReader/index.ts
sanchaymittal: we are delaying but deleting the record if res is undefined.

packages/router/src/bindings/contractReader/index.ts
sanchaymittal: yeah we can do that

packages/router/src/bindings/contractReader/index.ts
jakekidd: I meant deleting lol

jakekidd: What does `res` being undefined mean exactly?

sanchaymittal: handleSingle resolves with undefined if it errors or transaction fails

rhlsthrm (approved):

PR #377, opened 2021-09-26
Contract updates

LayneHaber:
## Checklist

- [ ] Run load tests.
- [ ] Update CHANGELOG.md.


packages/contracts/contracts/interpreters/FulfillInterpreter.sol
```diff
@@ -98,8 +99,9 @@ contract FulfillInterpreter is ReentrancyGuard, IFulfillInterpreter {
...
       callData,
       returnData,
-      success
+      success,
+      isContract
     );
-    return (success, returnData);
+    return (success, isContract, returnData);
   }
 }
```
jakekidd, line 107: nit: do we need newline here?

LayneHaber: We don't need it but I can add one

packages/contracts/contracts/interpreters/FulfillInterpreter.sol
```diff
@@ -98,8 +99,9 @@ contract FulfillInterpreter is ReentrancyGuard, IFulfillInterpreter {
       amount,
       callData,
       returnData,
-      success
+      success,
+      isContract
     );
-    return (success, returnData);
+    return (success, isContract, returnData);
```
jakekidd, line 105: personally I would have preferred to tack new variables on to the end of the return tuple here, but is there a reason it's going in the middle?

packages/contracts/contracts/interfaces/ITransactionManager.sol
```diff
@@ -11,6 +11,7 @@ interface ITransactionManager {
     address receivingChainTxManagerAddress;
     address user;
     address router;
+    address initiator; // msg.sender of sending side
```
jakekidd, line 14: I probs missed a beat somewhere, but won't initiator always be user essentially?

packages/contracts/contracts/interfaces/IFulfillInterpreter.sol
```diff
@@ -11,7 +11,8 @@ interface IFulfillInterpreter {
     uint256 amount,
     bytes callData,
     bytes returnData,
-    bool success
+    bool success,
+    bool isContract
```
jakekidd, line 15: another nit: are event objects ordered tuples, or indexed by keys? This ordering could match TransactionFulfilled below, where `returnData` is the last thing in the tuple per se, so that's why I ask

packages/contracts/contracts/TransactionManager.sol
```diff
@@ -110,13 +110,9 @@ contract TransactionManager is ReentrancyGuard, ProposedOwnable, ITransactionMan
     uint256 chain = chainId;
     if (chain == 0) {
       // If not provided, pull from block
-      assembly {
-        _chainId := chainid()
-      }
-    } else {
-      // Use provided override
-      _chainId = chain;
+      chain = block.chainid;
```
jakekidd, line 113: if chain == 0 couldn't that also be mainnet eth? sry not familiar with api of where this is pulled from

packages/contracts/contracts/TransactionManager.sol
```diff
@@ -503,11 +499,22 @@ contract TransactionManager is ReentrancyGuard, ProposedOwnable, ITransactionMan
       // on the sending chain
       require(msg.sender == txData.router, "#F:016");
 
+      // Validate the user has signed
+      require(recoverFulfillSignature(txData.transactionId, relayerFee, txData.receivingChainId, txData.receivingChainTxManagerAddress, signature) == txData.user, "#F:022");
```
jakekidd, line 503: I'm confused why this has been moved out from under the 'validation scope' above - although I'm not familiar enough with code conventions around sol to know why. do you mind explaining? what is that 'validation scope' exactly? if this is a validation step, shouldn't it also be under that scope?

jakekidd (commented):
have some questions, thanks :)

packages/contracts/contracts/interpreters/FulfillInterpreter.sol
LayneHaber: they're both booleans, so keeping the same data types next to each other

packages/contracts/contracts/interfaces/ITransactionManager.sol
LayneHaber: unless they're using a smart contract

packages/contracts/contracts/TransactionManager.sol
```diff
@@ -336,6 +332,9 @@ contract TransactionManager is ReentrancyGuard, ProposedOwnable, ITransactionMan
 
     // First determine if this is sender side or receiver side
     if (invariantData.sendingChainId == _chainId) {
+      // Check the sender is correct
+      require(msg.sender == invariantData.initiator, "#P:039");
```
jakekidd, line 336: this likely isn't feasible/useful but, is there any way we could return in the revert error the "bad initiator" here? would that be useful to know?

jakekidd: just realized this is `msg.sender`, lol dumb q

packages/contracts/contracts/TransactionManager.sol
LayneHaber: no chain == 0 means its uninitiated via the constructor, and so should be taken from the block. the reason we have the option to override the chain is because some chains don't have unique chain ids

packages/contracts/contracts/TransactionManager.sol
LayneHaber: It's moved out of the combined validation scope to strengthen the domain separation enforcement. You can see now we do the same recovery on both chains, but on the sending chain we use what's in the transaction data, whereas on the receiving chain we use what we can pull from the chain itself. the 'validation scope' is actually just a trick around the stack too deep issues -- by placing things in brackets you create a new scope that doesn't impact the local stack. i grouped all validation there where it should be performed the exact same across both chains

packages/contracts/contracts/TransactionManager.sol
LayneHaber: the error code maps to `INITIATOR_MISMATCH`

packages/contracts/contracts/TransactionManager.sol
```diff
@@ -503,11 +499,22 @@ contract TransactionManager is ReentrancyGuard, ProposedOwnable, ITransactionMan
...
 
     } else {
-      (success, returnData) = _receivingChainFulfill(txData, relayerFee, callData);
+      // Validate the user has signed, using domain of contract
+      require(recoverFulfillSignature(txData.transactionId, relayerFee, _chainId, address(this), signature) == txData.user, "#F:022");
+
+      // Sanity check: fee <= amount. Allow `=` in case of only 
+      // wanting to execute 0-value crosschain tx, so only providing 
+      // the fee amount
+      require(relayerFee <= txData.amount, "#F:023");
```
jakekidd, line 515: is it possible to execute a 0-value tx where the relayer gets the entire amount? wouldn't there be other participants that the fee would be paid out to, thereby detracting more portions from the amount? what happens if there are multiple relayers exactly? if this relayer is taking the full amount as the fee, how does the next relayer get their fee?

packages/contracts/contracts/interpreters/FulfillInterpreter.sol
```diff
@@ -73,7 +73,8 @@ contract FulfillInterpreter is ReentrancyGuard, IFulfillInterpreter {
     // Check if the callTo is a contract
     bool success;
     bytes memory returnData;
-    if (Address.isContract(callTo)) {
+    bool isContract = Address.isContract(callTo);
+    if (isContract) {
```
jakekidd, line 77: so we don't execute the call data if it's not targeting a contract then? who executes the call data then, exactly? sorry i'm fairly certain this is a really basic q

packages/contracts/contracts/interpreters/FulfillInterpreter.sol
jakekidd: thank you lol

packages/contracts/contracts/interpreters/FulfillInterpreter.sol
jakekidd: fair 100%

packages/contracts/contracts/interfaces/IFulfillInterpreter.sol
jakekidd: total nit, feel free to ignore/resolve this if you'd like. 100% your call

packages/contracts/contracts/TransactionManager.sol
jakekidd: some chains don't have unique chain ids? woah I did not know that

also, what would eth be then, if not chainid 0?

jakekidd (approved):

packages/contracts/contracts/TransactionManager.sol
jakekidd: Ah understood, thanks for explaining!

packages/contracts/contracts/TransactionManager.sol
LayneHaber: This is on the receiver side -- the router fee is already implicitly included because the amount it is prepared with on the receiving chain is less than the amount the user prepared with on the sending chain. That difference is the router fee.

The relayer fee is for someone who is submitting the transactions to chain. There can only be one person who claims the fee, since the transaction can only be executed once. The relayer could get the entire amount from the transaction, and this would make sense in the case where you want to execute calldata on a chain you don't have gas on. You wouldn't want your funds there, and including a relayer fee creates an incentive for someone to submit the tx on your behalf.

packages/contracts/contracts/interfaces/ITransactionManager.sol
```diff
@@ -103,6 +105,7 @@ interface ITransactionManager {
     bytes signature,
     bytes callData,
     bool success,
+    bool isContract,
```
rhlsthrm, line 108: Not understanding this one yet.

rhlsthrm: OK I think I get it.

rhlsthrm (approved):
Looks good to me. Sad we couldn't get the prepare relay sig to work, but we can always have a different approach for that.

PR #386, opened 2021-09-28
update chain config to potentially take an array of subgraph uris

### LINKED ISSUE #385: [utils/subgraph] Fallback subgraph
- [ ] update ChainConfig to take an array of subgraph endpoints (similar to the providers entry) --> this must be done in a backwards compatible way
- [ ] create a FallbackSubgraph that looks at the latest block of all the subgraphs, and uses the most in sync one
- [ ] runs the query on the most in sync subgraph

there may be some difficulty with how the clients of the subgraphs should get integrated -- right now both router and sdk have their own subgraph client (which is silly) but i think just being able to rotate between different instantiations would be the ideal

jakekidd:
Fixes #385 

## Checklist

- [ ] Run load tests.
- [ ] Update CHANGELOG.md.


packages/router/src/config.ts
```diff
@@ -51,7 +51,7 @@ export const TChainConfig = Type.Object({
   providers: Type.Array(Type.String()),
   confirmations: Type.Number({ minimum: 1 }),
   defaultInitialGas: Type.Optional(TIntegerString),
-  subgraph: Type.String(),
+  subgraph: Type.Array(Type.String()),
```
LayneHaber, line 54: This will not be backwards compatible, you should either handle the type difference or add a new property called `subgraphs` thats an optional array (then convert the value in subgraph to a `string[]`)

packages/utils/src/fallbackSubgraph.ts
```diff
@@ -0,0 +1,85 @@
...
+};
+
+export class FallbackSubgraph {
+  private readonly subgraphs: SubgraphSyncRecord[];
+
+  constructor(
+    private readonly logger: Logger,
+    private readonly chainId: number,
+    private readonly uris: string[],
+    private readonly allowUnsynced: number = 1,
```
LayneHaber, line 23: what does `allowUnsynced` mean here?

LayneHaber: If it's follow-on from the router/sdk configs it means the number of blocks a subgraph is allowed to be behind by before we consider it truly out of sync, and `1` is not a great default -- maybe like `25`?

packages/router/src/adapters/subgraph/graphqlsdk.ts
```diff
@@ -889,41 +882,56 @@ export const GetAssetBalanceDocument = gql`
...
     },
     GetTransactions(variables?: GetTransactionsQueryVariables, requestHeaders?: Dom.RequestInit["headers"]): Promise<GetTransactionsQuery> {
-      return withWrapper((wrappedRequestHeaders) => client.request<GetTransactionsQuery>(GetTransactionsDocument, variables, {...requestHeaders, ...wrappedRequestHeaders}), 'GetTransactions');
+      return withWrapper((wrappedRequestHeaders) => client.request<GetTransactionsQuery>(GetTransactionsDocument, variables, {...requestHeaders, ...wrappedRequestHeaders}), "GetTransactions");
     },
     GetAssetBalance(variables: GetAssetBalanceQueryVariables, requestHeaders?: Dom.RequestInit["headers"]): Promise<GetAssetBalanceQuery> {
-      return withWrapper((wrappedRequestHeaders) => client.request<GetAssetBalanceQuery>(GetAssetBalanceDocument, variables, {...requestHeaders, ...wrappedRequestHeaders}), 'GetAssetBalance');
+      return withWrapper((wrappedRequestHeaders) => client.request<GetAssetBalanceQuery>(GetAssetBalanceDocument, variables, {...requestHeaders, ...wrappedRequestHeaders}), "GetAssetBalance");
```
jakekidd: MARK: Real change in this file is here. All of the above is just Prettier eslint formatting

packages/router/src/adapters/subgraph/graphqlsdk.ts
```diff
@@ -804,126 +988,120 @@ export const GetReceiverTransactionsDocument = gql`
...
+      return withWrapper(
+        (wrappedRequestHeaders) =>
+          client.request<GetAssetBalanceQuery>(GetAssetBalanceDocument, variables, {
+            ...requestHeaders,
+            ...wrappedRequestHeaders,
+          }),
+        "GetAssetBalance",
+      );
+    },
+    UpdateSyncStatus(
```
jakekidd, line 1088: MARK: Only real change in this file is here: `UpdateSyncStatus`; all other changes are eslint/Prettier formatting.

packages/router/src/adapters/subgraph/subgraph.ts
```diff
@@ -69,25 +69,10 @@ const setSyncRecord = async (chainId: number, requestContext: RequestContext) =>
...
-        subgraphBlockNumber,
-        chainId,
-      });
-      record = { synced: false, latestBlock: realBlockNumber, syncedBlock: subgraphBlockNumber };
-    } else {
-      record = { synced: true, latestBlock: realBlockNumber, syncedBlock: subgraphBlockNumber };
-    }
-    synced[chainId] = record;
-    return record;
+    return await sdk.UpdateSyncStatus(realBlockNumber);
```
jakekidd, line 75: This call doesn't actually return the sync records like it used to. Additionally, all logging is done internally.

Is that what we want? Or would we like to return sync records?

packages/utils/src/fallbackSubgraph.ts
```diff
@@ -0,0 +1,80 @@
+import { GraphQLClient } from "graphql-request";
+import { Variables, RequestDocument } from "graphql-request/dist/types";
+import { Headers } from "graphql-request/dist/types.dom";
+
+import { Logger } from "./logger";
+import { createLoggingContext } from "./request";
+
+export type SubgraphSyncRecord = {
```
jakekidd, line 8: This seemed like a fine way to organize things, but if we need to separate the sync record from the client, I can change the type to be organized such that the sync record is nested in another type that holds the client.
i.e.
```ts
type Subgraph = {
  client: GraphQLClient;
  uri: string;
  syncRecord: SubgraphSyncRecord;
}
```
or something like that. Lmk @LayneHaber

packages/router/src/adapters/subgraph/index.ts
```diff
@@ -39,10 +38,10 @@ export const getSdks = (): Record<number, Sdk> => {
 };
 
 export const subgraphContractReader = (): ContractReader => {
-  const { config } = getContext();
+  const { config, logger } = getContext();
   Object.entries(config.chainConfig).forEach(([chainId, { subgraph }]) => {
-    const client = new GraphQLClient(subgraph);
-    sdks[parseInt(chainId)] = getSdk(client);
+    const chainIdNumber = parseInt(chainId);
+    sdks[chainIdNumber] = getSdk(new FallbackSubgraph(logger, chainIdNumber, subgraph));
```
LayneHaber, line 44: The `getSdk` function is generated by the `codegen` command, using the graph schema and the queries outlined in `queries.ts`. It won't be something that's easy to meddle with. [Here's](https://github.com/connext/nxtp/blob/8a05946b82bfd80fdd174a192cfdfeed910ea614/packages/sdk/src/subgraph/subgraph.ts#L207) [some](https://github.com/connext/nxtp/blob/8a05946b82bfd80fdd174a192cfdfeed910ea614/packages/sdk/src/subgraph/subgraph.ts#L325-L329) [examples](https://github.com/connext/nxtp/blob/8a05946b82bfd80fdd174a192cfdfeed910ea614/packages/router/src/adapters/subgraph/subgraph.ts#L347-L349) of how the clients are used.

It would be a good idea to explicitly make the `FallbackSubgraph` extend the `GraphQLClient` class, but I think this is likely the best way to go about it

packages/router/src/config.ts
jakekidd, line 54: @LayneHaber changed this to array type, but below I'm checking to see if the subgraph property passed in is a string type - and it handles 'coercing' it into an array, per se. Does that work? It looks like we only do validation at the end there, just wanted to double check that this is functional.

jakekidd: I was originally doing the latter, will reimplement

packages/utils/src/fallbackSubgraph.ts
jakekidd: Gotcha, I actually meant to take the default out, and I'll do so now. It should come straight from config, no default should be set here.

packages/router/src/config.ts
jakekidd: So just to clarify however, I actually handle parsing the item passed into chainConfig.subgraph below, checking whether it's a string type, then coercing it to array per se if so. All of that is done before validation.

Why is it that that isn't backwards compat? Will it really error if I pass a string into subgraph, or won't it go through this coercing/formatting process first?

packages/router/src/adapters/subgraph/index.ts
jakekidd: This *is* the best way to go about it? The problem has particularly to do with the `GetBlockNumberQuery` being locked up in the `graphqlsdk.ts` file. If I try to make the get block number call to subgraph internal to the fallback provider, I end up with an antipattern of sorts

packages/router/src/config.ts
LayneHaber: Didn't realize the validation was done after the coercing, this will be fine

packages/utils/src/fallbackSubgraph.ts
LayneHaber: I think this is fine, also the real time is the `Sdk<T>` type below

packages/utils/src/fallbackSubgraph.ts
```diff
@@ -0,0 +1,80 @@
...
+  latestBlock: number;
+  syncedBlock: number;
+}
+
+type Sdk<T> = {
+  client: T;
+  record: SubgraphSyncRecord;
+};
+
+export class FallbackSubgraph<T extends { GetBlockNumber: () => Promise<any>; }> {
```
LayneHaber, line 15: Should really define the type it's extending, since it should also be used in the `Sdk<T>` type above

packages/utils/src/fallbackSubgraph.ts
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  public getSynced(): T {
+    const synced = this.synced;
+    if (synced.length === 0) {
+      throw new Error("No subgraphs available / in-sync!");
+    }
+    // Get a random client to ensure we are varying which subgraph client we're using.
+    return synced[Math.floor(Math.random() * synced.length)].client;
+  }
```
LayneHaber, line 50: why not preferentially get the most in sync client? then select randomly if there are multiple clients that are at the same processing block

LayneHaber (changes requested):
some smol changes around the types and subgraph selection, but otherwise lgtm!

packages/utils/src/fallbackSubgraph.ts
jakekidd: Ahhh gotcha, will do

packages/utils/src/fallbackSubgraph.ts
jakekidd: sure, i like that

rhlsthrm (commented):
I was thinking to do this simpler where we assume the subgraph is synced and only change the subgraph URL if it comes back out of sync. This should work also though, but I'm a little nervous about the complexity. This sync step happens on each loop iteration right?

packages/router/src/adapters/subgraph/index.ts
jakekidd: > This _is_ the best way to go about it?
This sounded really sassy but I actually was just trying to clarify loll

packages/utils/src/fallbackSubgraph.ts
jakekidd: ohh woops, I implemented a weighted system, must have misremembered this comment lol.

Should I go with the weighted system, preferring *more* in sync..? Hmmm, I think actually the way you described it would be best

jakekidd:
> I was thinking to do this simpler where we assume the subgraph is synced and only change the subgraph URL if it comes back out of sync. This should work also though, but I'm a little nervous about the complexity. This sync step happens on each loop iteration right?

We could certainly do it that way, but the problem there is that we wouldn't really have all the information to select the most in-sync subgraph URL. So we could end up with a subgraph that's lagging 8-10 blocks behind but consistently staying within what we require, while another unused resource of ours is actually totally in-sync, unbeknownst to us. By checking sync status each cycle, we ensure we're using the most insync subgraph each cycle as well.

If there's tradeoffs there, lmk

packages/utils/src/fallbackSubgraph.ts
```diff
@@ -0,0 +1,117 @@
...
+    for (let i = 0; i < synced.length; i++) {
+      const sdk = synced[i];
+      const difference = sdk.record.latestBlock - sdk.record.syncedBlock;
+      if (difference < bestSync) {
+        mostInSyncClients = [sdk.client];
+        bestSync = difference;
+      } else if (difference === bestSync) {
+        mostInSyncClients.push(sdk.client);
+      }
+    }
```
LayneHaber, line 65: can this be done in a `.sort()` method?

LayneHaber (approved):
one comment, but otherwise lgtm

packages/sdk/test/unit/subgraph.spec.ts
```diff
@@ -112,7 +112,7 @@ const convertMockedToActiveTransaction = (
   };
 };
 
-describe("Subgraph", () => {
+describe.only("Subgraph", () => {
```
jakekidd, line 115: oops

PR #387, opened 2021-09-29
update: yarn version berry

sanchaymittal:
- chg: yarn version to berry
- update lock and yarn version
- add: workplace plugin
- update: typechain
- chg: ignore catch unknon type errors/
- update: typechain import
- fix: type errors


rhlsthrm:
Why are we doing this? Node 16 I think has some issues with some of our 3rd part libraries.

sanchaymittal:
Not bumping it to 16.x, just tried cause it's failing the docker build

PR #399, opened 2021-10-04
Estimage gas on meta tx request

wanglonghong:
## Checklist

- [ ] Run load tests.
- [ ] Update CHANGELOG.md.


wanglonghong:
> Great work! Very close to being done here. Can we add a function to `sdk.ts` which is called `estimateFulfillFee` that can be called by the fulfill function itself? This way we can expose a method that clients use to get the fee.

implemented in commit c9ce5887154094ae8e00600b40f34296da8ef680

PR #412, opened 2021-10-06
[router] Hotfix: prevent error in active tx retrieval from blocking other chains

jakekidd:
## The Problem
In `getActiveTransactions` in `packages/router/src/adapters/subgraph/subgraph.ts`,  if any one of the anonymous async methods to which we map in the `Promise.all` throws, then we will short circuit all other methods for all other chains. It will cause `getActiveTransactions` to throw the error - so we won't return anything.

## The Solution

I added a try/catch inside the anonymous methods in the mapping in the `Promise.all` block. We log the errors when they occur in the catch block, and put them in a `Map<string, Error>` object indexed by chainId. Currently I'm just using that `Map` to check whether the # errors that occurred === # chains we support; in which case, we throw the error.

This will enable us to continue supporting the working chains even when 1 chain is consistently failing for some reason.

## Checklist

- [ ] Run load tests.
- [ ] Update CHANGELOG.md.

### DIFF
--- a/packages/router/src/adapters/subgraph/subgraph.ts
+++ b/packages/router/src/adapters/subgraph/subgraph.ts
@@ -107,212 +107,229 @@ export const getActiveTransactions = async (_requestContext?: RequestContext): P
 
   // get local context
   const sdks = getSdks();
+  const errors: Map<string, Error> = new Map();
   const allChains = await Promise.all(
     Object.entries(sdks).map(async ([cId, sdk]) => {
-      const chainId = parseInt(cId);
+      try {
+        const chainId = parseInt(cId);
 
-      const chainConfig = config.chainConfig[chainId];
-      if (!chainConfig) {
-        throw new NoChainConfig(chainId);
-      }
-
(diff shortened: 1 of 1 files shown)


packages/router/src/adapters/subgraph/subgraph.ts
```diff
@@ -107,212 +107,233 @@ export const getActiveTransactions = async (_requestContext?: RequestContext): P
...
           if (!_sdk) {
+            // if receiving SDK doesnt exist, cancel all the txs
             logger.warn(
-              "No contract reader available for receiver chain, filtering txs",
-              requestContext,
+              "No contract reader available for receiver chain, marking sender tx for cancellation",
+              { ...requestContext, transactionId: senderTx.transactionId } as RequestContext<string>,
               methodContext,
-              { receivingChain: cId },
+              { sendingChain: senderTx.chainId, receivingChain: senderTx.receivingChainId },
```
jakekidd, line 155: idk why this change is part of this diff - I did not make it

rhlsthrm:
Yes, can we just remake your changes on top of main?

PR #414, opened 2021-10-06
Slightly modified the subgraph to be able to calculate the time taken for a swap

0xakshay:
## The Problem

Currently, it's not possible to calculate the time taken for a swap using the current subgraphs because there is no timestamp for when the transaction is fulfilled. 
There's no timestamp for when the transaction is canceled either, this might be needed in the data analysis. 

## The Solution

Added fulfillTimestamp and cancelTimestamp to the schema. 
Mapped these two variables with the right data on mapping.ts 

## Checklist

- Deployed and tested with the Arbitrum contract.

### DIFF
--- a/packages/subgraph/schema.graphql
+++ b/packages/subgraph/schema.graphql
@@ -61,11 +61,13 @@ type Transaction @entity {
   fulfillCaller: Bytes
   fulfillTransactionHash: Bytes
   fulfillMeta: Bytes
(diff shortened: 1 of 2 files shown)


rhlsthrm:
I like this idea! Thanks!

rhlsthrm (approved):

rhlsthrm (approved):

PR #445, opened 2021-10-15
VAMM Integration

### LINKED ISSUE #437: [router] Adjusting Liquidity Updates
With the VAMMs, removing and adding liquidity has to be done carefully as to not impact prices for a given transaction. Also has to be done in the proper ratios. Tasks:

- [ ] Update `remove-liquidity` endpoint to halt transaction prepare processing while liquidity ratios are adjusted
- [ ] Update `add-liquidity` methods on the frontend to add transactions simultaneously
- [ ] UI update to add liquidity -- should be able to input the amounts for both chains, validate that the prices are okay compared with what exists

LayneHaber:
## The Problem

Fixes #437 

- Need to display the price impact when adding liquidity to a router
- Need to revert if price impact is too high when removing liquidity
- Need to provide override methods for both ^^ to handle initial cases that prevents the router from responding to bids or preparing transfers when the liquidity adjustments are being made.

## The Solution

Remaining tasks:
- [ ] Fix UI to have the virtual AMM per-chain
- [ ] Add max price impact to config

## Checklist

- [ ] Run load tests.
- [ ] Update CHANGELOG.md.


packages/router/src/lib/operations/auction.ts
```diff
@@ -185,8 +173,75 @@ export const newAuction = async (
...
+  logger.info("Got balances of router", requestContext, methodContext, {
+    routerBalances: routerBalances.map((balance) => balance.toString()),
+    sendingChainIdx,
+    receivingChainIdx,
+  });
+
+  const [senderLiquidity, receiverLiquidity] = await Promise.all([
+    contractReader.getAssetBalance(sendingAssetId, sendingChainId),
+    contractReader.getAssetBalance(receivingAssetId, receivingChainId),
+  ]);
```
LayneHaber, line 227: these will have different decimals than the `routerBalances` values

packages/router/src/lib/operations/auction.ts
```diff
@@ -185,8 +173,75 @@ export const newAuction = async (
...
+  });
+
   // getting the swap rate from the receiver side config
-  let amountReceived = await getReceiverAmount(amount, inputDecimals, outputDecimals);
+
+  let amountReceived = await getReceiverAmount(
+    amount,
+    inputDecimals,
+    outputDecimals,
+    routerBalances,
```
LayneHaber, line 240: we should probably have the conversion to 18 decimals from whatever it has onchain happen in only one place, so in the `getReceiverAmount` function only

packages/router/src/lib/operations/auction.ts
```diff
@@ -220,13 +275,11 @@ export const newAuction = async (
 
   amountReceived = amountReceivedInBigNum.sub(gasFeeInReceivingToken).toString();
 
-  const balance = await contractReader.getAssetBalance(receivingAssetId, receivingChainId);
-  logger.info("Got asset balance", requestContext, methodContext, { balance: balance.toString() });
-  if (balance.lt(amountReceived)) {
+  if (receiverLiquidity.lt(amountReceived)) {
```
LayneHaber, line 278: decimals not converted for liquidity

packages/router/src/lib/operations/prepare.ts
```diff
@@ -78,14 +80,88 @@ export const prepare = async (
...
+      );
+
+      receivingChainIdx = pool.assets.findIndex(
+        (a) =>
+          getAddress(a.assetId) === getAddress(invariantData.receivingAssetId) &&
+          a.chainId === invariantData.receivingChainId,
+      );
+      swapPoolIndex = index;
+    }
+  });
```
LayneHaber, line 113: lets make a helper for this, we have the exact same (i believe) code in `newAuction`

packages/router/src/lib/operations/prepare.ts
```diff
@@ -78,14 +80,88 @@ export const prepare = async (
...
+  const swapPool = config.swapPools[swapPoolIndex];
+  const routerBalances = await Promise.all(
+    swapPool.assets.map(async (asset) => {
+      const assetLiquidity = await contractReader.getAssetBalance(asset.assetId, asset.chainId);
+      let assetDecimals = chainData.get(asset.chainId.toString())?.assetId[asset.assetId]?.decimals;
+      if (!assetDecimals) {
+        assetDecimals = await txService.getDecimalsForAsset(asset.chainId, asset.assetId);
+      }
+      // convert asset liquidity into 18 decimal value.
+      const res = assetLiquidity.mul(BigNumber.from(10).pow(18 - assetDecimals));
```
LayneHaber, line 137: same comment as above here re: keeping decimal conversions in one function

packages/router/src/lib/operations/prepare.ts
```diff
@@ -78,14 +80,88 @@ export const prepare = async (
...
+      return res;
+    }),
   );
 
-  let receiverAmount = await getReceiverAmount(senderAmount, inputDecimals, outputDecimals);
+  let inputDecimals = chainData.get(invariantData.sendingChainId.toString())?.assetId[invariantData.sendingAssetId]
+    ?.decimals;
+  if (!inputDecimals) {
+    inputDecimals = await txService.getDecimalsForAsset(invariantData.sendingChainId, invariantData.sendingAssetId);
+  }
```
LayneHaber, line 146: lets make a helper for this too

packages/router/src/lib/operations/prepare.ts
```diff
@@ -78,14 +80,88 @@ export const prepare = async (
...
+
+  let receiverAmount = await getReceiverAmount(
+    senderAmount,
+    inputDecimals,
+    outputDecimals,
+    routerBalances,
+    sendingChainIdx,
+    receivingChainIdx,
+    config.maxPriceImpact,
+  );
```
LayneHaber, line 164: if the `PriceImpactTooHigh` error is thrown, we should mark that error as cancellable, so that the router will cancel the users transaction automatically if it cannot prepare

packages/router/src/lib/helpers/prepare.ts
```diff
@@ -33,39 +35,73 @@ export const validBidExpiry = (bidExpiry: number, currentTime: number) => bidExp
...
  */
-export const getSwapRate = async (): Promise<string> => {
-  return "1";
+export const getSwapRate = async (
+  amount: BigNumber,
+  balances: BigNumber[],
+  indexIn: number,
+  indexOut: number,
+): Promise<BigNumber> => {
+  const amountOut = getAmountsOut(
```
LayneHaber, line 49: will the `getAmountsOut` function properly return only integers?

packages/router/src/lib/helpers/prepare.ts
```diff
@@ -33,39 +35,73 @@ export const validBidExpiry = (bidExpiry: number, currentTime: number) => bidExp
...
- * @returns The swapRate, determined by the AMM
+ * @returns The swapRate, determined by the stableSwap
  *
  * @remarks
- * TODO: getSwapRate using AMM
+ * TODO: getSwapRate using stableSwap
  */
-export const getSwapRate = async (): Promise<string> => {
-  return "1";
+export const getSwapRate = async (
```
LayneHaber, line 43: this doesn't necessarily return the swap *rate* but instead returns the swapped amount, right? if so, we should rename the function so that people (*cough* me *cough*) don't get confused

packages/utils/src/stableMath.ts
```diff
@@ -0,0 +1,188 @@
...
+  convertedBalances[tokenIndexIn] = convertedBalances[tokenIndexIn].plus(tokenAmountIn);
+
+  const finalBalanceOut: BigNumber = _getTokenBalanceGivenInvariantAndAllOtherBalances(
+    convertedAmp,
+    convertedBalances,
+    lastInvariant,
+    tokenIndexOut,
+  );
+
+  return convertedBalances[tokenIndexOut].minus(finalBalanceOut).minus(BIG_ONE).toFixed(0).toString();
```
LayneHaber, line 142: ahh nice, disregard above comment since this should convert to an integer string!

LayneHaber (commented):
can't request changes since i created the PR but, there's one high priority change (the liquidity vs. balances decimal issues) and some other low-er priority ones (lower default, helper functions, converting to 18 out of onchain decimals in one place)

this looks awesome though! are there any other remaining tasks?

packages/router/src/config.ts
```diff
@@ -28,6 +28,7 @@ import contractDeployments from "@connext/nxtp-contracts/deployments.json";
 const MIN_GAS = utils.parseEther("0.1");
 const DEFAULT_RELAYER_FEE_THRESHOLD = "10"; // relayerFee is in respective chain native asset unit
 const MIN_SUBGRAPH_SYNC_BUFFER = 25;
+const DEFAULT_MAX_PRICE_IMPACT = 20; // max price impact in percentage
```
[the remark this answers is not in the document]
wanglonghong: I thought defaultMaxPriceImpact means the max value of price impact that we can allow. not default price impact.  maxPriceImpact of Uniswap and PancakeSwap is around 50%. larger than ours. 

`advanced` you mean here that we should have a UI option in frontend side so users are able to change price impact. right?

packages/router/src/lib/helpers/prepare.ts
```diff
@@ -4,16 +4,18 @@ import {
...
 import { BigNumber } from "ethers";
+import { PriceImpactTooHigh } from "../errors/auction";
 
 import { AmountInvalid } from "../errors/prepare";
 
 const ROUTER_FEE = "0.05"; // 0.05%
 const EXPIRY_DECREMENT = 3600 * 24;
 const ONE_DAY_IN_SECONDS = 3600 * 24;
 const ONE_WEEK_IN_SECONDS = 3600 * 24 * 7;
+const AMPLIFICATION = "85";
```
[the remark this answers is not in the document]
wanglonghong: Right now, it is a constant value. I also think it could be configurable. router/src/config.ts. is it ok to read this value from env file or config json just like other configs?

packages/router/src/lib/helpers/prepare.ts
```diff
@@ -33,39 +35,73 @@ export const validBidExpiry = (bidExpiry: number, currentTime: number) => bidExp
...
-  const amountAfterSwapRate = calculateExchangeWad(BigNumber.from(amount), inputDecimals, swapRate, outputDecimals);
 
+  const inputAmount = BigNumber.from(amount).mul(BigNumber.from(10).pow(18 - sendingDecimals));
+  // 1. swap rate from stableMath
+  const amountAfterSwapRate = await getSwapRate(BigNumber.from(inputAmount), routerBalances, sendingIdx, receivingIdx);
+
+  // check price impact
+  const deltaPrice = amountAfterSwapRate.gt(inputAmount)
+    ? amountAfterSwapRate.sub(inputAmount)
+    : BigNumber.from(inputAmount).sub(amountAfterSwapRate);
```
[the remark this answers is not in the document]
wanglonghong: No reason at all. It could probably be a reason that I like to use low-level coding. lol. Thanks. I will absolutely use .abs().

packages/router/src/lib/operations/auction.ts
```diff
@@ -185,8 +173,75 @@ export const newAuction = async (
...
+
+    return existSwap;
+  });
+  if (!allowedSwap) {
+    throw new SwapInvalid(sendingChainId, sendingAssetId, receivingChainId, receivingAssetId, {
+      methodContext,
+      requestContext,
+    });
+  }
+
```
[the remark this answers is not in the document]
wanglonghong: I got you. let me fix

packages/router/src/lib/operations/auction.ts
wanglonghong: yeah. no problem. routerBalances are used to calculate swapRate using stableMath so its decimal is 18. but receiverLiquidity is used to check if receiver liquidity is enough to process request or not so its decimal should be equal to receiving asset decimals, any mistakes here?

packages/router/src/lib/operations/auction.ts
wanglonghong: sorry?

packages/router/src/lib/operations/auction.ts
wanglonghong: getReceiverAmount returns the value with receiving asset decimals even if routerBalances has 18 decimals.  receiverLiquidity and amountReceived have same decimals right now.

packages/router/src/lib/operations/prepare.ts
wanglonghong: ok. will do!

packages/router/src/lib/operations/prepare.ts
wanglonghong: yeah!

packages/router/src/lib/helpers/prepare.ts
wanglonghong: absolutely!

packages/router/src/lib/helpers/prepare.ts
wanglonghong: right. will rename!.

packages/utils/src/stableMath.ts
wanglonghong: :)

packages/router/src/lib/operations/prepare.ts
wanglonghong: so you mean we have to have helper function?

packages/router/src/lib/operations/prepare.ts
wanglonghong: ok. I got you

packages/router/src/config.ts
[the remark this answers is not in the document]
LayneHaber: yeah i think 20% is still a little high for us -- but open to leaving it as is.

and re: `advanced` yes, exactly

packages/router/src/lib/helpers/prepare.ts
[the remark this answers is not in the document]
LayneHaber: yep!

packages/router/src/lib/helpers/prepare.ts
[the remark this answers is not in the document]
LayneHaber: no pressure! who knows maybe the low level stuff is more performant, but `abs` is a bit easier for me to read personally

packages/router/src/lib/operations/auction.ts
LayneHaber: Ahh my mistake -- I thought you were comparing it to `routerBalances` for some reason. Must've misread.

Still, I think we should keep all conversions to and from 18 decimals within the `getReceiverAmount` function to avoid these types of confusions. I was totally wrong about the error here, but since it only has to be 18 in all cases within that function, we should encapsulate all decimal-based manipulation to that function

packages/router/src/lib/operations/auction.ts
LayneHaber: wow highlighted the incorrect bit of code 😅 

Same as I mentioned above -- there's only one function that requires decimal manipulation, so lets make sure we keep the manipulation isolated to that function

packages/router/src/lib/operations/auction.ts
LayneHaber: Correct :)

packages/router/src/lib/operations/prepare.ts
LayneHaber: nope, [this](https://github.com/connext/nxtp/pull/445#discussion_r747927783) idea, I guess "above" was used too loosely here!

LayneHaber (commented):
A couple of issues:
- You have to track the pending balances by asset and chain, not just by chain
- You can't clear the pending balances at the top of each loop because there's a race between when it's cleared, when auctions access it, and when its considered fully developed. Instead, you should only update this at the end of each loop. Also, clearing this means any data about transactions that are prepared on the sending and receiving side is removed (since those transactions aren't returned in the `getActiveTransaction` function)
- We are missing an instance of calculating pending balances. We should *add* to the router balance when the transaction is prepared by the router, and we should *subtract* from the pending balance when a transaction is fulfilled by the router. 

Pending balance should be thought of as any funds that are not accounted by the subgraph liquidity values, but are logically owned by the router. There are a couple of relevant cases:
1. Transfer is `SenderPrepared` -- this will *increase* the routers pending balance on the receiving side to account for the deduction by the subgraph in the tracked liquidity. there are no changes to the sending side liquidity (router is logically not owner of those funds)
2. Transfer is `ReceiverFulfilled` -- this will *increase* the routers pending balance on the sending side to account for the claimable asset.
3. Transfer is `ReceiverPrepared` (see below) -- this will *increase* routers pending balance on the receiving side.

We also have to update the `getActiveTransaction` function so that it will return transactions where both sides have prepared (i.e. add a `ReceiverPrepared` status) so we can properly account for this balance in the pending balances within the processing loop. We didn't have this status previously because no router action is required, but now this status keeps this calculation stateless.

packages/router/src/bindings/contractReader/index.ts
```diff
@@ -180,6 +186,22 @@ export const handleSingle = async (
...
+      }
+      const sendingDecimals = await getDecimalsForAsset(
+        transaction.crosschainTx.invariant.sendingChainId,
+        transaction.crosschainTx.invariant.sendingAssetId,
+      );
+      const sendingAmountInEther = BigNumber.from(transaction.crosschainTx.sending.amount).mul(18 - sendingDecimals);
+      pendingLiquidityMap.set(
+        transaction.crosschainTx.invariant.sendingChainId,
+        lastLiquidityInBigNum.add(sendingAmountInEther),
+      );
```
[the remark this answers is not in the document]
wanglonghong: oh yeah. you are right

packages/router/src/adapters/subgraph/subgraph.ts
```diff
@@ -116,19 +116,18 @@ export const getActiveTransactions = async (_requestContext?: RequestContext): P
...
             routerId: routerAddress.toLowerCase(),
             receivingChainId: chainId,
             status: SdkTransactionStatus.Prepared,
             expiry_lt: Math.floor(Date.now() / 1000),
           }),
         );
-        if ((allReceiverExpired.router?.transactions.length ?? 0) > 0) {
-          logger.debug("Got receiver expired", requestContext, methodContext, {
+        if ((allReceiverPrepared.router?.transactions.length ?? 0) > 0) {
+          logger.debug("Got receiver prepared", requestContext, methodContext, {
```
LayneHaber, line 129: are we properly handling cases with receiver expired? `ReceiverExpired` means the router has to take the action of cancelling that tx. This query seems to get the expired + prepared txs only, so maybe the variable is misnamed?

packages/router/src/adapters/subgraph/subgraph.ts
```diff
@@ -387,31 +434,7 @@ export const getActiveTransactions = async (_requestContext?: RequestContext): P
...
-                    cancelHash: expTx.cancelTransactionHash,
-                    fulfillHash: expTx.fulfillTransactionHash,
-                  },
-                },
-              },
-              status: CrosschainTransactionStatus.ReceiverExpired,
-            } as ActiveTransaction<"ReceiverExpired">;
-          });
-        return filterUndefined.concat(receiverNotConfigured).concat(remainingReceiverExpired);
+        return filterUndefined.concat(receiverNotConfigured);
```
LayneHaber, line 437: seems like this edgecase handling got removed

packages/router/src/bindings/contractReader/index.ts
```diff
@@ -100,6 +101,47 @@ export const handleActiveTransactions = async (
...
+        tmpLiquidityMap.set(liquidityKey, prevLiquidity.add(pendingAmount));
+      } else {
+        tmpLiquidityMap.set(
+          transaction.crosschainTx.invariant.receivingChainId
+            .toString()
+            .concat("-")
+            .concat(transaction.crosschainTx.invariant.receivingAssetId),
+          BigNumber.from(pendingAmount),
+        );
+      }
```
LayneHaber, line 128: could make this entire if block:

```ts
tmpLiquidityMap.set(liquidityKey, (prevLiquidity ?? constants.Zero).add(pendingAmount))
```

packages/router/src/bindings/contractReader/index.ts
```diff
@@ -100,6 +101,47 @@ export const handleActiveTransactions = async (
...
+    } else if (transaction.status === CrosschainTransactionStatus.ReceiverFulfilled) {
+      tmpLiquidityMap.set(
+        transaction.crosschainTx.invariant.sendingChainId
+          .toString()
+          .concat("-")
+          .concat(transaction.crosschainTx.invariant.sendingAssetId),
+        BigNumber.from(transaction.crosschainTx.sending?.amount),
+      );
```
LayneHaber, line 136: this will overwrite everything that exists for that asset, and only save it for that transaction when it should be cumulative

packages/router/src/bindings/contractReader/index.ts
```diff
@@ -109,6 +151,9 @@ export const handleActiveTransactions = async (
       transaction.crosschainTx.invariant.transactionId,
     );
 
+    // if transfer is receiver prepared, just no router action required
+    if (transaction.status === CrosschainTransactionStatus.ReceiverPrepared) continue;
+
```
LayneHaber, line 156: probably a bit more logically consistent to add this status-based check to the `handleSingle` function

LayneHaber:
Marked everything as resolved before submitting this review btw

ishan-verma:
Just one change that would be really helpful:

Currently one amplification value is applied to all pools but it would be nice if we could add amplification param for each coin/pool. 

instead of this config file structure 

```
"swapPools": [
    {
      "name": "DAI",
      "assets": [
        {
          "chainId": 137,
          "assetId": "0x8f3Cf7ad23Cd3CaDbD9735AFf958023239c6A063"
        },
        {
          "chainId": 250,
          "assetId": "0x6B175474E89094C44Da98b954EedeAC495271d0F"
        },
        {
          "chainId": 42161,
          "assetId": "0xDA10009cBd5D07dd0CeCc66161FC93D7c9000da1"
        },
        {
          "chainId": 43114,
          "assetId": "0xd586E7F844cEa2F87f50152665BCbc2C279D8d70"
        }
      ]
    },
    {
      "name": "USDC",
      "assets": [
        {
          "chainId": 137,          
          "assetId": "0x2791Bca1f2de4661ED88A30C99A7a9449Aa84174"
        },
        {
          "chainId": 250,
          "assetId": "0x04068DA6C83AFCFA0e13ba15A6696662335D5B75"
        },
        {
          "chainId": 42161,
          "assetId": "0xFF970A61A04b1cA14834A43f5dE4533eBDDB5CC8"
        },
        {
          "chainId": 43114,
          "assetId": "0xA7D7079b0FEaD91F3e65f86E8915Cb59c1a4C664"
        }
      ]
    },
    {
      "name": "USDT",
      "assets": [

        {
          "chainId": 137,
          "assetId": "0xc2132D05D31c914a87C6611C10748AEb04B58e8F"
        },
        {
          "chainId": 250,
          "assetId": "0x049d68029688eAbF473097a2fC38ef61633A3C7A"
        },
        {          
          "chainId": 42161,
          "assetId": "0xFd086bC7CD5C481DCC9C85ebE478A1C0b69FCbb9"
        },
        {
          "chainId": 43114,
          "assetId": "0xc7198437980c041c805A1EDcbA50c1Ce5db95118"
        }
      ]
    }
  ],
  "cleanupMode": false,
  "allowedVAMM": true,
  "amplification": 85,
  "maxPriceImpact": 90
}
```
we could have something like this 
```
"swapPools": [
    {
      "name": "DAI",
      "amplification": 1000,
      "assets": [
        {
          "chainId": 137,
          "assetId": "0x8f3Cf7ad23Cd3CaDbD9735AFf958023239c6A063"
        },
        {
          "chainId": 250,
          "assetId": "0x6B175474E89094C44Da98b954EedeAC495271d0F"
        },
        {
          "chainId": 42161,
          "assetId": "0xDA10009cBd5D07dd0CeCc66161FC93D7c9000da1"
        },
        {
          "chainId": 43114,
          "assetId": "0xd586E7F844cEa2F87f50152665BCbc2C279D8d70"
        }
      ]
    },
    {
      "name": "USDC",
      "amplification": 205,
      "assets": [
        {
          "chainId": 137,          
          "assetId": "0x2791Bca1f2de4661ED88A30C99A7a9449Aa84174"
        },
        {
          "chainId": 250,
          "assetId": "0x04068DA6C83AFCFA0e13ba15A6696662335D5B75"
        },
        {
          "chainId": 42161,
          "assetId": "0xFF970A61A04b1cA14834A43f5dE4533eBDDB5CC8"
        },
        {
          "chainId": 43114,
          "assetId": "0xA7D7079b0FEaD91F3e65f86E8915Cb59c1a4C664"
        }
      ]
    },
    {
      "name": "USDT",
      "amplification": 185,
      "assets": [
        {
          "chainId": 137,
          "assetId": "0xc2132D05D31c914a87C6611C10748AEb04B58e8F"
        },
        {
          "chainId": 250,
          "assetId": "0x049d68029688eAbF473097a2fC38ef61633A3C7A"
        },
        {          
          "chainId": 42161,
          "assetId": "0xFd086bC7CD5C481DCC9C85ebE478A1C0b69FCbb9"
        },
        {
          "chainId": 43114,
          "assetId": "0xc7198437980c041c805A1EDcbA50c1Ce5db95118"
        }
      ]
    }
  ],
  "cleanupMode": false,
  "allowedVAMM": true,
  "maxPriceImpact": 90
}
```

PR #486, opened 2021-11-02
Handle fulfill relay

rhlsthrm:
## The Problem

`allowFulfillRelay = false` was not respected.

## Checklist

- [ ] Run load tests.
- [ ] Update CHANGELOG.md.

### DIFF
--- a/packages/router/src/config.ts
+++ b/packages/router/src/config.ts
@@ -246,14 +246,15 @@ export const getEnvConfig = (crossChainData: Map<string, any> | undefined): Nxtp
       nxtpConfig.chainConfig[chainId].priceOracleAddress = res?.address;
     }
 
-    if (chainConfig.minGas === undefined) {
+    if (!chainConfig.minGas) {
       nxtpConfig.chainConfig[chainId].minGas = MIN_GAS.toString();
     }
-    if (chainConfig.relayerFeeThreshold === undefined) {
+
+    if (!chainConfig.relayerFeeThreshold) {
       nxtpConfig.chainConfig[chainId].relayerFeeThreshold = +DEFAULT_RELAYER_FEE_THRESHOLD;
     }
 
-    if (chainConfig.allowFulfillRelay === undefined) {
+    if (chainConfig.allowFulfillRelay === undefined || chainConfig.allowFulfillRelay === null) {
       nxtpConfig.chainConfig[chainId].allowFulfillRelay = true;
     }
 
--- a/packages/router/src/config.ts
+++ b/packages/router/src/config.ts
@@ -246,14 +246,14 @@ export const getEnvConfig = (crossChainData: Map<string, any> | undefined): Nxtp
       nxtpConfig.chainConfig[chainId].priceOracleAddress = res?.address;
     }
 
-    if (!chainConfig.minGas) {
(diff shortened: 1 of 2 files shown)


packages/router/src/config.ts
```diff
@@ -246,14 +246,14 @@ export const getEnvConfig = (crossChainData: Map<string, any> | undefined): Nxtp
       nxtpConfig.chainConfig[chainId].priceOracleAddress = res?.address;
     }
 
-    if (!chainConfig.minGas) {
+    if (chainConfig.minGas === undefined) {
```
LayneHaber, line 249: won't work for `null` -- should do `typeof` probably

LayneHaber (changes requested):

packages/router/src/config.ts
rhlsthrm: It won't be null though, it's parsed from JSON or env vars.

jakekidd: technically you can pass null in to JSON so safest thing is `chainConfig.minGas === undefined || chainConfig.minGas === null`

wanglonghong: I think we don't need to change first two conditions. because their types are string and number, not boolean.
- if (!chainConfig.minGas) 
- if (!chainConfig.relayerFeeThreshold)

But we have to absolutely revise allowFulfillRelay part. because type of allowFulFillRelay variable is boolean. so it should be `if (!chainConfig.allowRelayFulfill === undefined)`

wanglonghong: if (!chainConfig.minGas)

packages/router/src/config.ts
```diff
@@ -246,14 +246,14 @@ export const getEnvConfig = (crossChainData: Map<string, any> | undefined): Nxtp
...
       nxtpConfig.chainConfig[chainId].minGas = MIN_GAS.toString();
     }
-    if (!chainConfig.relayerFeeThreshold) {
+    if (chainConfig.relayerFeeThreshold === undefined) {
```
wanglonghong, line 252: if (!chainConfig.relayerFeeThreshold)

packages/router/src/config.ts
```diff
@@ -246,14 +246,14 @@ export const getEnvConfig = (crossChainData: Map<string, any> | undefined): Nxtp
...
       nxtpConfig.chainConfig[chainId].relayerFeeThreshold = +DEFAULT_RELAYER_FEE_THRESHOLD;
     }
 
-    if (!chainConfig.allowFulfillRelay) {
+    if (chainConfig.allowFulfillRelay === undefined) {
```
wanglonghong, line 256: Correct!

packages/router/src/config.ts
rhlsthrm: > I think we don't need to change first two conditions. because their types are string and number, not boolean.

Correct, I removed the change on the first two.

> so it should be if (!chainConfig.allowRelayFulfill === undefined)

It would need to be `chainConfig.allowFulfillRelay === undefined` because you only want to set the default if it's undefined. We don't need that extra !.

wanglonghong: Correct. `!` was my mistake.

PR #487, opened 2021-11-03
[sdk/router] fix fallback subgraph in-sync requirements

jakekidd:
## The Problem

Subgraph failing on one chain due to all the subgraph providers being out of sync is causing failure on all chains.

This seems to be occurring in the sdk, not in router - bc router's calls are all wrapped in try/catch per chain.

Error:
```
{"level":50,"time":1635867900140,"pid":1,"hostname":"80993d7e403d","name":"0x29A519e21d6A97cdB82270b69c98bAc6426CDCf9","requestContext":{"id":"st3wTsHgQ6e9OkZ2nP44LQ/7241201","origin":"getActiveTransactions"},"methodContext":{"id":"st3wTsHgQ6e9OkZ2nP44LQ/7241200","name":"getActiveTransactions"},"chainId":"250","error":{"message":"All subgraphs out of sync! Cannot handle active transactions for chain 56","type":"Error","context":{},"stack":"Error: All subgraphs out of sync! Cannot handle active transactions for chain 56\n    at FallbackSubgraph.request (/home/node/node_modules/@connext/nxtp-utils/dist/fallbackSubgraph.js:37:19)\n    at /home/node/router/dist/adapters/subgraph/subgraph.js:144:42\n    at Array.map (<anonymous>)\n    at /home/node/router/dist/adapters/subgraph/subgraph.js:135:79\n    at runMicrotasks (<anonymous>)\n    at processTicksAndRejections (internal/process/task_queues.js:95:5)\n    at async Promise.all (index 3)\n    at async Object.getActiveTransactions (/home/node/router/dist/adapters/subgraph/subgraph.js:78:23)\n    at async Timeout._onTimeout (/home/node/router/dist/bindings/contractReader/index.js:21:28)"},"msg":"Error getting active transactions for chain 250"}
```

## The Solution

Some adjustments were made to some faulty bit of logic for fallback subg, as well as to the SDK and router logic.

No changes to CHANGELOG as this should be a part of fallback subg

## Checklist

- [ ] Run load tests.
- [x] Update CHANGELOG.md.

### DIFF
--- a/packages/sdk/src/subgraph/subgraph.ts
+++ b/packages/sdk/src/subgraph/subgraph.ts
@@ -4,6 +4,7 @@ import {
   FallbackSubgraph,
   jsonifyError,
   Logger,
+  NxtpError,
   RequestContext,
   TransactionData,
   VariantTransactionData,
@@ -190,6 +191,16 @@ export class Subgraph {
   async getActiveTransactions(_requestContext?: RequestContext): Promise<ActiveTransaction[]> {
     const { requestContext, methodContext } = createLoggingContext(this.getActiveTransactions.name, _requestContext);
 
+    // For each chain, update subgraph sync status
+    await this.updateSyncStatus();
+    const inSyncChains: number[] = Object.entries(this.sdks).filter(([, sdk]) => sdk.inSync).map(([chainId]) => parseInt(chainId));
+    if (inSyncChains.length === 0) {
+      // If no chains are in-sync, return an empty array.
+      // TODO: Or throw an error - all chains out of sync?
+      this.logger.warn("Cannot get active transactions; all subgraphs for all chains are out of sync!", requestContext, methodContext);
+      return [];
+    }
+
     // Step 1: handle any already listed as active transactions.
     // This is important to make sure the events are properly emitted
     // when sdks remain online for the duration of the transaction.
@@ -200,6 +211,9 @@ export class Subgraph {
     // Get all ids organized in an object with keyed on their receiving chain id
     const idsBySendingChains: Record<string, string[]> = {};
     [...this.activeTxs.entries()].forEach(([id, tx]) => {
+      if (!inSyncChains.includes(tx.crosschainTx.invariant.sendingChainId)) {
+        return;
+      }
       if (!idsBySendingChains[tx.crosschainTx.invariant.sendingChainId]) {
         idsBySendingChains[tx.crosschainTx.invariant.sendingChainId] = [id];
       } else {
@@ -207,8 +221,7 @@ export class Subgraph {
       }
     });
 
-    // For each chain, update subgraph sync status
-    await this.updateSyncStatus();
+    
 
     // Gather matching sending-chain records from the subgraph that will *not*
     // be handled by step 2 (i.e. statuses are *not* prepared)
@@ -228,6 +241,7 @@ export class Subgraph {
 
         const { transactions } = await subgraph.request<GetTransactionsQuery>((client) =>
           client.GetTransactions({ transactionIds: ids }),
+          false,
         );
         if (transactions.length === 0) {
           return;
@@ -243,6 +257,9 @@ export class Subgraph {
 
     // Get corresponding receiver transaction records
     nonPreparedSendingTxs.forEach((transaction) => {
+      if (!inSyncChains.includes(transaction.receivingChainId)) {
+        return;
+      }
       if (!correspondingReceiverTxIdsByChain[transaction.receivingChainId]) {
         correspondingReceiverTxIdsByChain[transaction.receivingChainId] = [transaction.transactionId];
       } else {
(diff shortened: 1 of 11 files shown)


packages/sdk/src/subgraph/subgraph.ts
```diff
@@ -190,6 +191,16 @@ export class Subgraph {
...
 
+    // For each chain, update subgraph sync status
+    await this.updateSyncStatus();
+    const inSyncChains: number[] = Object.entries(this.sdks).filter(([, sdk]) => sdk.inSync).map(([chainId]) => parseInt(chainId));
+    if (inSyncChains.length === 0) {
+      // If no chains are in-sync, return an empty array.
+      // TODO: Or throw an error - all chains out of sync?
+      this.logger.warn("Cannot get active transactions; all subgraphs for all chains are out of sync!", requestContext, methodContext);
+      return [];
+    }
```
LayneHaber, line 202: active transactions should return as many as possible transactions, even if the subgraph is out of sync. similar to historical transactions

LayneHaber: cc @rhlsthrm want to make sure this is desired behavior. seems like this is pretty risk-free in the sdk but in the router could lead to some undesired actions on subgraph loops

packages/sdk/src/subgraph/subgraph.ts
```diff
@@ -329,12 +346,11 @@ export class Subgraph {
 
     // Step 2: handle any not-listed active transactions (i.e. sdk has
     // gone offline at some point during the transactions)
-    const errors: Map<string, Error> = new Map();
+    const errors: Map<number, Error> = new Map();
     const txs = await Promise.all(
-      Object.keys(this.sdks).map(async (c) => {
+      inSyncChains.map(async (chainId) => {
```
LayneHaber, line 351: this logic may need to change for the same reasoning as the above comment

jakekidd:
Just ran this with test-ui, working

packages/sdk/src/subgraph/subgraph.ts
jakekidd: This would result in potential duplicates, though, which means wasted gas.
Ex.
getActiveTransactions -> new sender prepare (at block N) -> handle, add to our local active txs -> handled -> clear from local active txs.
subg goes out of sync, stuck at block N -> should be considered out of sync but we ignore it -> get the same sender prepare as before (at block N) -> try to handle -> P019

but maybe we don't care about these edge cases too much..?

packages/sdk/src/subgraph/subgraph.ts
jakekidd: Gotcha, I think I'll wait for @rhlsthrm's feedback on this one.

packages/sdk/src/subgraph/subgraph.ts
rhlsthrm: Yes I agree with @LayneHaber I think we need to return all active transactions even if it's out of sync.

> subg goes out of sync, stuck at block N -> should be considered out of sync but we ignore it -> get the same sender prepare as before (at block N) -> try to handle -> P019

This flow does not happen on the SDK, the SDK doesn't handle prepares by preparing. On the subgraph it will just show up as active.

packages/sdk/src/subgraph/subgraph.ts
rhlsthrm: Agreed, we need to return all the active transactions regardless of out of sync. In fact on the SDK I don't think we need to care about out-of-sync at all in this subgraph loop. We only need to:
1. Not send auction requests involving a particular chain if all subgraphs are out of sync.
2. Choose the "in sync" subgraph out of all the fallbacks.
3. Once we have the subgraph chosen, just treat it as in-sync.

What are the cases we need to worry about on the SDK?
- Sender stops syncing
  - In this case, we won't see a sender prepare on the active transactions. That's fine, just continue to monitor and poll and make sure we are using the synced subgraph (if any) on each loop iteration.
- Receiver stops syncing
  - Will not see the receiver prepare after the router does it. That's fine, just continue to monitor and poll and make sure we are using the synced subgraph (if any) on each loop iteration.

We don't actually need to change any of the active tx logic here, all this needs to do is choose the correct subgraph.

This might be true for the router as well. I think this only needs to affect the initial auction methods.

rhlsthrm (changes requested):
Might need a restructure based on my latest comments.

packages/sdk/src/subgraph/subgraph.ts
jakekidd: Okay, I'm going to make it the default for the `request` method that sync is *not* required

jakekidd:
validated test ui is working for me here on latest commit

packages/utils/src/fallbackSubgraph.ts
```diff
@@ -111,6 +187,41 @@ export class FallbackSubgraph<T extends SdkLike> {
...
+        avgExecTime,
+        reliability,
+      };
+    });
+    // Order the subgraphs based on these metrics:
+    // 1. Lag, which is the difference between the latest block and the subgraph's latest block.
+    // 2. CPS, which is the number of calls per second the subgraph has been making (averaged over last N calls).
+    // 3. Reliability, which is how often RPC calls to that subgraph are successful / total calls out of last N calls.
+    // 4. Average execution time, which is the average execution time of the last N calls.
+    sdks.forEach((sdk) => {
```
rhlsthrm, line 221: Cool stuff!

rhlsthrm (approved):

PR #502, opened 2021-11-05
created stable math

liu-zhipeng:
## Checklist

- [ ] Run load tests.
- [ ] Update CHANGELOG.md.


packages/utils/package.json
```diff
@@ -26,6 +26,7 @@
     "ajv": "^8.6.2",
     "ajv-formats": "^2.1.1",
     "axios": "^0.21.1",
+    "bignumber.js": "^9.0.1",
```
LayneHaber, line 29: IIRC this is a different `BigNumber` library than [ethers uses](https://github.com/ethers-io/ethers.js/blob/b1458989761c11bf626591706aa4ce98dae2d6a9/package.json#L95), with slightly different capabilities (this lib can handle decimals, while the `BN` lib cannot), but the same interface. Mixing these two classes has historically been problematic with types, and it would be ideal if they could all use the same library

LayneHaber: If the decimals functionality is needed for handling approximations, we can leave it but everything leaving these functions should be parsed to an integer string

packages/utils/src/stableMath.ts
```diff
@@ -0,0 +1,209 @@
...
+    // n = number of tokens                                                                      //
+    *********x************************************************************************************/
+    // D(n+1) = D(n) - (f(Dn)/f'(Dn)) = (AnnS + nDp)Dn / [(Ann-1)Dn + (n+1)Dp]
+    // We support rounding up or down.
+
+    let sum = BIG_ZERO;
+    let numTokens = balances.length;
+    for (let i = 0; i < numTokens; i++) {
+        sum = sum.plus(balances[i])
+    }
```
LayneHaber, line 30: we should use a `reduce` statement instead:

```js
const sum = balances.reduce((prev, curr) => prev.add(curr), BIG_ZERO)
```

packages/utils/src/stableMath.ts
```diff
@@ -0,0 +1,209 @@
...
+const BIG_TEN = new BigNumber(10)
+
+
+
+// Computes the invariant given the current balances, using the Newton-Raphson approximation.
+// The amplification parameter equals: A n^(n-1)
+export const _calculateInvariant = (
+  amplificationParameter: BigNumber,
+  balances: BigNumber[]
+) :BigNumber => {
```
LayneHaber, line 14: see first comment re: types -- return value should be parsed to an integer string, or we should use the ethers big number library

packages/utils/src/stableMath.ts
```diff
@@ -0,0 +1,209 @@
...
+    let D = sum;    // Dn+1
+    let Ann = new BigNumber(amplificationParameter).times(numTokens);  // An
+    
+    for (let i = 0; i < 255; i++) {
+        let D_P = D;
+
+        // calculate derivative of f(D)
+        for(let j = 0; j < numTokens; j++) {
+          D_P = D_P.multipliedBy(D).dividedBy(new BigNumber(balances[j]).times(numTokens));
+        }
```
LayneHaber, line 46: could put this into a `reduce` statement as well:

```js
D_P = balances.reduce((prev, curr) => D_P.multipliedBy(D).dividedBy(new BigNumber(curr)).times(numTokens)), D)
```

packages/utils/src/stableMath.ts
```diff
@@ -0,0 +1,209 @@
...
+        
+        if (D.gt(Dprev)) {
+          if (D.minus(Dprev).lte(BIG_ONE)) {
+            break;
+          }
+        } else {
+          if(Dprev.minus(D).lte(BIG_ONE)) {
+            break;
+          }
+        }
```
LayneHaber, line 63: Seems like you can take the absolute value of `D.minus(Dprev).abs()` and it'll be a bit easier to read

packages/utils/src/stableMath.ts
LayneHaber: If this function will just be used internally (i.e. no need for the router to call `_calculateInvariant` directly, you can leave the return type but not export the function

packages/utils/src/stableMath.ts
```diff
@@ -0,0 +1,209 @@
...
+
+const BIG_ZERO = new BigNumber(0)
+const BIG_ONE = new BigNumber(1)
+const BIG_TEN = new BigNumber(10)
+
+
+
+// Computes the invariant given the current balances, using the Newton-Raphson approximation.
+// The amplification parameter equals: A n^(n-1)
+export const _calculateInvariant = (
```
LayneHaber, line 11: we should only prefix functions with `_` if it will not be exported

packages/utils/src/stableMath.ts
```diff
@@ -0,0 +1,209 @@
...
+        }
+    }
+
+    return D;
+}
+
+
+// This function calculates the balance of a given token (tokenIndex)
+// given all the other balances and the invariant
+export const _getTokenBalanceGivenInvariantAndAllOtherBalances = (
```
LayneHaber, line 72: See above comment about `_` prefixes

packages/utils/src/stableMath.ts
```diff
@@ -0,0 +1,209 @@
...
+  amp: BigNumber,
+  balances: BigNumber[],
+  invariant: BigNumber,
+  tokenIndex: number
+): BigNumber => {
```
LayneHaber, line 77: See above comments about return types

packages/utils/src/stableMath.ts
```diff
@@ -0,0 +1,209 @@
...
+    c = c.multipliedBy(invariant).dividedBy(_x.multipliedBy(numTokens))
+  }
+      
+  c = c.multipliedBy(invariant).dividedBy(Ann.multipliedBy(numTokens))
+  let b = S_.plus(invariant.dividedBy(Ann))
+
+  let y_prev = BIG_ZERO
+  let y = invariant
+
+  for(let i = 0; i < 255; i++) {
```
LayneHaber, line 102: We should add more comments around why 255 is chosen here

packages/utils/src/stableMath.ts
```diff
@@ -0,0 +1,209 @@
...
+    
+    if (y.gt(y_prev)) {  
+      if(y.minus(y_prev).lte(BIG_ZERO)) {
+        break
+      }
+    } else {
+      if(y_prev.minus(y).lte(BIG_ZERO)) {
+        break
+      }
+    }
```
LayneHaber, line 114: Same comment re: `abs()` function for readability

packages/utils/src/stableMath.ts
```diff
@@ -0,0 +1,209 @@
...
+  // getAmountsOut token x for y - polynomial equation to solve                                                //
+  // ay = amount out to calculate                                                                              //
+  // by = balance token out                                                                                    //
+  // y = by - ay (finalBalanceOut)                                                                             //
+  // D = invariant                                               D                     D^(n+1)                 //
+  // A = amplification coefficient               y^2 + ( S - ----------  - D) * y -  ------------- = 0         //
+  // n = number of tokens                                    (A * n^n)               A * n^2n * P              //
+  // S = sum of final balances but y                                                                           //
+  // P = product of final balances but y                                                                       //
+  **************************************************************************************************************/
```
LayneHaber, line 147: These are super helpful comment blocks, nice work

packages/utils/src/stableMath.ts
```diff
@@ -0,0 +1,209 @@
...
+  
+  let convertedBalances = balances.map((balance) => new BigNumber(balance))
```
LayneHaber, line 149: this can be a `const`

packages/utils/src/stableMath.ts
```diff
@@ -0,0 +1,209 @@
...
+  const convertedAmp = new BigNumber(amp)
+
+  const lastInvariant = _calculateInvariant(convertedAmp, convertedBalances);
+
+  convertedBalances[tokenIndexIn] = convertedBalances[tokenIndexIn].plus(tokenAmountIn);
+
+  let finalBalanceOut: BigNumber = _getTokenBalanceGivenInvariantAndAllOtherBalances(
```
LayneHaber, line 156: can also be a `const`

LayneHaber (changes requested):
There are some comments in here regarding types and readability nit picks, but well done! would ideally like those addressed before merging into main, but safe to merge into #445 as is.

Would also be a good idea to run `lint` on the utils package before merging

LayneHaber:
Good stuff in here :)

packages/utils/package.json
wanglonghong: yeah, BN is definitely different from BigNumber of ethers. but ethers BigNumber has a disadvantage when processing float numbers. so we discussed this with @rhlsthrm before using it. I think we need to definitely use bignumber.js. so the results would be more correct.

LayneHaber: Yeah, makes sense. One of the biggest annoyances.

We should just make sure that functions that take in this type from the interface aren't exported, and that the results are converted to integer strings when they return from the exported functions

LayneHaber (approved):
Looks good!

PR #558, opened 2021-11-25
Router contract integration

rhlsthrm:
Part 1/2 for #581 

- [x] Finish functions to send to Gelato.
- [x] Finish meta tx functionality in router.
- [x] Change prepare in router to handle receiver side prepares (i.e. metatx).
- [x] Test on Rinkeby by deploying RouterFactory and Router.sol


packages/router/src/adapters/contract/contract.ts
```diff
@@ -209,16 +249,51 @@ export const prepare = async (
...
+    if (isChainSupportedByGelato(chainId)) {
+      try {
+        await gelatoSend(
+          chainId,
+          nxtpContractAddress,
+          encodedData,
+          chainId === txData.sendingChainId ? txData.sendingAssetId : txData.receivingAssetId,
+          "0",
+        );
+      } catch (err) {}
```
jakekidd, line 262: We should have a general enclosure for try catch that resorts to using txservice as a backup

wanglonghong (approved):
looks good

packages/contracts/contracts/Router.sol
```diff
@@ -0,0 +1,220 @@
+// SPDX-License-Identifier: UNLICENSED
```
LayneHaber, line 1: need to improve comments and code coverage here and in the factory before merging:

```
  Router.sol              |        0 |        0 |        0 |        0 |... 203,207,218 |
  RouterFactory.sol       |     6.25 |        0 |    16.67 |     6.25 |... 63,70,71,75 |
  ```

File: packages/contracts/contracts/Router.sol [hunk 3]
```diff
@@ -0,0 +1,220 @@
...
+pragma solidity ^0.8.4;
+
+import "./interfaces/ITransactionManager.sol";
+import "./lib/LibAsset.sol";
+import "@openzeppelin/contracts/access/Ownable.sol";
+import "@openzeppelin/contracts/utils/cryptography/ECDSA.sol";
+
+contract Router is Ownable {
```
LayneHaber, line 9: should we do `ProposedOwnable` instead?

packages/contracts/contracts/Router.sol
```diff
@@ -0,0 +1,220 @@
...
+  address public immutable routerFactory;
+
+  ITransactionManager public transactionManager;
+
+  uint256 private chainId;
```
LayneHaber, line 14: This has to match the transaction manager's `chainId`, not necessarily the canonical chain of this contract

packages/contracts/contracts/Router.sol
```diff
@@ -0,0 +1,220 @@
...
+
+  address public recipient;
+
+  address public routerSigner;
+
+  struct SignedPrepareData {
+    ITransactionManager.PrepareArgs args;
+    address routerRelayerFeeAsset;
+    uint256 routerRelayerFee;
+    uint256 chainId; // For domain separation
```
LayneHaber, line 24: if you want to include proper domain separation, you also need to include the address of this contract. same to all the other signed type structs here

packages/contracts/contracts/Router.sol
```diff
@@ -0,0 +1,220 @@
...
+  function removeRelayerFee(uint256 amount, address assetId) external onlyOwner {
+    // Sanity check: nonzero amounts
+    require(amount > 0, "#RC_RRF:002");
+
+    // Transfer funds from contract
+    LibAsset.transferAsset(assetId, payable(recipient), amount);
+
+    // Emit event
+    emit RelayerFeeRemoved(assetId, amount, msg.sender);
+  }
```
LayneHaber, line 108: If someone compromises the owner, they could easily call `removeLiquidity` then `removeRelayerFee`. I don't think there's anyway around this, but it's important we spell that out somewhere in the documentation of the contracts

packages/contracts/contracts/Router.sol
```diff
@@ -0,0 +1,220 @@
...
+    if (msg.sender != routerSigner) {
+      SignedRemoveLiquidityData memory payload = SignedRemoveLiquidityData({
+        amount: amount,
+        assetId: assetId,
+        chainId: chainId
+      });
+
+      address recovered = recoverSignature(abi.encode(payload), signature);
+      require(recovered == routerSigner, "#RC_RL:040");
+    }
```
LayneHaber, line 124: Do you need to handle router relayer fees here?

packages/contracts/contracts/Router.sol
```diff
@@ -0,0 +1,220 @@
...
+        chainId: chainId
+      });
+
+      address recovered = recoverSignature(abi.encode(payload), signature);
+      require(recovered == routerSigner, "#RC_P:040");
+
+      // Send the relayer the fee
+      if (routerRelayerFee > 0) {
+        LibAsset.transferAsset(routerRelayerFeeAsset, payable(msg.sender), routerRelayerFee);
+      }
```
LayneHaber, line 149: Also need to require that the amount stored in the contract is greater than the relayer fee

packages/contracts/contracts/Router.sol
```diff
@@ -0,0 +1,220 @@
...
+    }
+
+    return
+      transactionManager.prepare{value: LibAsset.isNativeAsset(args.invariantData.sendingAssetId) ? msg.value : 0}(
+        args
+      );
+  }
```
LayneHaber, line 156: If the router sent any funds to the contract, it would need to be `receivingAssetId` , but all router liquidity is stored on chain already

should probably have some sanity checks around the chain id of this signer and the chain id of the transaction you're preparing

LayneHaber: Could also add more sanity checks around the router specified in the args and the calls

packages/contracts/contracts/Router.sol
```diff
@@ -0,0 +1,220 @@
...
+        chainId: chainId
+      });
+
+      address recovered = recoverSignature(abi.encode(payload), signature);
+      require(recovered == routerSigner, "#RC_F:040");
+
+      // Send the relayer the fee
+      if (routerRelayerFee > 0) {
+        LibAsset.transferAsset(routerRelayerFeeAsset, payable(msg.sender), routerRelayerFee);
+      }
```
LayneHaber, line 178: Sanity checks

packages/contracts/contracts/Router.sol
```diff
@@ -0,0 +1,220 @@
...
+        chainId: chainId
+      });
+
+      address recovered = recoverSignature(abi.encode(payload), signature);
+      require(recovered == routerSigner, "#RC_C:040");
+
+      // Send the relayer the fee
+      if (routerRelayerFee > 0) {
+        LibAsset.transferAsset(routerRelayerFeeAsset, payable(msg.sender), routerRelayerFee);
+      }
```
LayneHaber, line 204: sanity checks

packages/contracts/src/errors.ts
```diff
@@ -45,6 +45,9 @@ export const Errors: ErrorsType = {
   "037": "NO_PROPOSAL",
   "038": "NO_OWNERSHIP_CHANGE",
   "039": "INITIATOR_MISMATCH",
+  "040": "INVALID_ROUTER_SIGNATURE",
```
LayneHaber, line 48: why not use `"022": "INVALID_SIGNATURE",`?

packages/contracts/src/errors.ts
```diff
@@ -45,6 +45,9 @@ export const Errors: ErrorsType = {
...
+  "041": "ROUTER_SIGNER_EMPTY",
```
LayneHaber, line 49: why not use `"001": "ROUTER_EMPTY",`?

packages/contracts/src/errors.ts
```diff
@@ -76,6 +79,14 @@ export const ErrorsPrefix: ErrorsPrefixType = {
...
   "#PNO": "proposeNewOwner",
   "#RO": "renounceOwnership",
+  "#RC_ARF": "routerContract_addRelayerFee",
+  "#RC_RRF": "routerContract_removeRelayerFee",
+  "#RC_RL": "routerContract_removeLiquidity",
+  "#RC_P": "routerContract_prepare",
+  "#RC_F": "routerContract_fulfill",
+  "#RC_C": "routerContract_cancel",
+  "#RF_I": "routerFactory_init",
+  "#RF_CR": "routerFactory_createRouter",
```
LayneHaber, line 89: This is using a slightly different error code format, i see the benefits but i'm a sucker for consistency.

packages/contracts/src/tasks/routerFactory_create.ts
```diff
@@ -0,0 +1,27 @@
...
+    console.log("recipient: ", recipient);
+
+    console.log("namedAccounts: ", namedAccounts);
+
+    const routerFactoryDeployment = await deployments.get("RouterFactory");
+    const routerFactoryAddress = routerFactoryDeployment.address;
+
+    console.log("routerFactoryAddress: ", routerFactoryAddress);
+
+    const routerFactory = await ethers.getContractAt("RouterFactory", routerFactoryAddress);
```
LayneHaber, line 19: can we check that the router factory didn't already deploy a contract at that address before we try to recreate it?

packages/contracts/test/routerFactory.spec.ts
```diff
@@ -0,0 +1,109 @@
...
+import { deployContract, assertReceiptEvent } from "./utils";
+
+// import types
+import { RouterFactory } from "../typechain";
+
+import { getContractError } from "../src";
+
+const { AddressZero } = constants;
+const createFixtureLoader = waffle.createFixtureLoader;
+describe("Router Factory Contract", function () {
```
LayneHaber, line 21: coverage is broken, and needs to be improved for the factory and router contracts

packages/router/src/adapters/contract/index.ts
```diff
@@ -1,47 +1,85 @@
...
-    fulfill,
-    cancel,
-    removeLiquidity,
+    prepareTransactionManager,
+    prepareRouterContract,
+    fulfillTransactionManager,
+    fulfillRouterContract,
+    cancelRouterContract,
+    cancelTransactionManager,
+    removeLiquidityTransactionManager,
```
LayneHaber, line 82: can't you do an internal check to see if you should use the `prepareTransactionManager` over the `prepareRouterContract` for each of these functions so it would just show up as `prepare` and handle the checks internally?

packages/router/src/bindings/fastify/index.ts
```diff
@@ -47,7 +47,7 @@ export const bindFastify = () =>
           return res.code(401).send("Unauthorized to perform this operation");
         }
         try {
-          const result = await contractWriter.removeLiquidity(
+          const result = await contractWriter.removeLiquidityTransactionManager(
```
LayneHaber, line 50: what if they have a contract based router?

packages/router/src/bindings/fastify/index.ts
```diff
@@ -72,7 +72,7 @@ export const bindFastify = () =>
         }
         try {
           const senderTx = await prepareCancel({ senderChainId, user, transactionId });
-          const result = await contractWriter.cancel(
+          const result = await contractWriter.cancelTransactionManager(
```
LayneHaber, line 75: what if they have a contract based router?

packages/router/src/lib/helpers/shared.ts
```diff
@@ -152,6 +192,50 @@ export const calculateGasFeeInReceivingTokenForFulfill = async (
...
+
+  const assetIdOnMainnet = await getMainnetEquivalent(assetId, chainId);
+  const tokenPricingChain = assetIdOnMainnet ? 1 : chainId;
+  const tokenPricingAssetId = assetIdOnMainnet ? assetIdOnMainnet : assetId;
+
+  const nativeAssetIdOnMainnet = await getMainnetEquivalent(constants.AddressZero, chainId);
+  const nativeTokenPricingChain = nativeAssetIdOnMainnet ? 1 : chainId;
+  const nativeTokenPricingAssetId = nativeAssetIdOnMainnet ? nativeAssetIdOnMainnet : constants.AddressZero;
+
+  const gasFeeRes = await txService.calculateGasFee(
```
LayneHaber, line 224: tx service will also need to know if they should account for bump in estimates due to transaction manager vs. router contract calls (these should have different estimates)

packages/router/src/lib/helpers/shared.ts
```diff
@@ -161,6 +245,160 @@ export const getTokenPriceFromOnChain = async (
...
+      //   (_user, _router, _transactionId, args, _success, _isContract, _returnData, _caller, event) => {
+      //     fulfillEvt.post({ event, args });
+      //   },
+      // );
+      // contract.on("TransactionCancelled", (_user, _router, _transactionId, args, _caller, event) => {
+      //   cancelEvt.post({ event, args });
+      // });
+    }
+  });
+};
```
LayneHaber, line 279: not sure i understand why we need this

LayneHaber: or why we need it in the router instead of txservice

packages/router/src/lib/operations/auction.ts
```diff
@@ -229,31 +230,34 @@ export const newAuction = async (
...
+    // Log if gas is low, but above min
+    const LOW_GAS = parseEther("0.1");
+    if (senderBalance.lt(LOW_GAS) || receiverBalance.lt(LOW_GAS)) {
+      logger.warn("Router has low gas", requestContext, methodContext, {
+        sendingChainId,
+        receivingChainId,
+        senderBalance: formatEther(senderBalance),
+        receiverBalance: formatEther(receiverBalance),
+      });
+    }
```
LayneHaber, line 253: This should be configured as some amount above the configured

packages/router/src/adapters/contract/index.ts
LayneHaber: i think this may be in place in the `operations`?

packages/router/src/lib/operations/cancel.ts
```diff
@@ -96,17 +102,65 @@ export const cancel = async (
...
+  let receipt: providers.TransactionReceipt;
+
+  if (isRouterContract) {
+    routerRelayerFeeAsset = utils.getAddress(
+      config.chainConfig[invariantData.receivingChainId].routerContractRelayerAsset || AddressZero,
+    );
+    const relayerFeeAssetDecimal = await txService.getDecimalsForAsset(
+      invariantData.receivingChainId,
+      invariantData.receivingAssetId,
+    );
```
LayneHaber, line 114: isn't this the wrong asset?

packages/router/src/lib/operations/metaTx.ts
```diff
@@ -0,0 +1,267 @@
...
+  // receiver fulfill, done directly on contract using user's sig broadcast
+  if (type === MetaTxTypes.Fulfill) {
+    const { txData, signature, relayerFee, callData } = data as MetaTxFulfillPayload;
+    // Send to tx service
+    logger.info("Sending fulfill tx", requestContext, methodContext, { signature });
+
+    let outputDecimals = chainData.get(txData.receivingChainId.toString())?.assetId[txData.receivingAssetId]?.decimals;
+    if (!outputDecimals) {
+      outputDecimals = await txService.getDecimalsForAsset(txData.receivingChainId, txData.receivingAssetId);
+    }
```
LayneHaber, line 76: really should just make a helper for this, we use this same logic everywhere

packages/router/src/lib/operations/metaTx.ts
```diff
@@ -0,0 +1,267 @@
...
+        signature: signature,
+        relayerFee: relayerFee,
+        callData: callData,
+      },
+      requestContext,
+    );
+
+    logger.info("Method complete", requestContext, methodContext, { transactionHash: receipt.transactionHash });
+  } else {
+    // router contract methods
```
LayneHaber, line 113: any router can also participate in relaying other routers contract txs? dont we want to offload all of this to gelato (that has proper fallbacks), or *only* relay if gelato fails

packages/router/src/lib/operations/metaTx.ts
```diff
@@ -0,0 +1,267 @@
...
+      } = data as MetaTxRouterContractPreparePayload;
+
+      const routerRelayerFee = await calculateGasFee(
+        chainId,
+        relayerFeeAsset,
+        relayerFeeAssetDecimal,
+        "prepare",
+        requestContext,
+        methodContext,
+      );
```
LayneHaber, line 137: all of these gas fees need to have the adjustments for the costs of calling router prepare vs direct contract prepare

packages/router/src/lib/operations/prepare.ts
```diff
@@ -74,11 +78,11 @@ export const prepare = async (
...
 
-  const routerAddress = await wallet.getAddress();
+  const signerAddress = await wallet.getAddress();
   const recovered = recoverAuctionBid(bid, bidSignature);
-  if (recovered !== routerAddress) {
+  if (recovered !== signerAddress) {
     // cancellable error
-    throw new AuctionSignerInvalid(routerAddress, recovered, { methodContext, requestContext });
+    throw new AuctionSignerInvalid(signerAddress, recovered, { methodContext, requestContext });
   }
```
LayneHaber, line 86: should you also check if the bid has the right router address in it if you are using a contract router? otherwise signer could change and all previously accepted bids would fail here

packages/router/src/lib/operations/prepare.ts
```diff
@@ -200,20 +206,78 @@ export const prepare = async (
...
-      expiry: receiverExpiry,
-      bidSignature,
-      encodedBid,
+  if (isRouterContract) {
+    const routerRelayerFeeAsset = utils.getAddress(
+      config.chainConfig[invariantData.receivingChainId].routerContractRelayerAsset ?? AddressZero,
+    );
+    const relayerFeeAssetDecimal = await txService.getDecimalsForAsset(
+      invariantData.receivingChainId,
+      invariantData.receivingAssetId,
```
LayneHaber, line 216: wrong asset

packages/router/src/router.ts
```diff
@@ -52,10 +55,19 @@ export const makeRouter = async () => {
     context.wallet = context.config.mnemonic
       ? Wallet.fromMnemonic(context.config.mnemonic)
       : new Web3Signer(context.config.web3SignerUrl!);
+    context.signerAddress = await context.wallet.getAddress();
+    context.isRouterContract = context.config.routerContractAddress ? true : false;
```
LayneHaber, line 59: `!!context.config.routerContractAddress` ?

packages/txservice/src/chainreader.ts
```diff
@@ -352,17 +349,16 @@ export class ChainReader {
...
-  private async calculateGasFee(
+  public async calculateGasFee(
     chainId: number,
     gasPriceChainId: number,
     assetId: string,
     nativeTokenChainId: number,
     nativeTokenAssetId: string,
     decimals: number,
-    method: "prepare" | "fulfill",
+    method: "prepare" | "fulfill" | "cancel",
```
LayneHaber, line 359: needs to also have a flag to know if this is a router contract or direct transaction manager prepare, and should have different estimates for both

packages/utils/src/gasEstimates.ts
```diff
@@ -1,6 +1,8 @@
 export const GAS_ESTIMATES = {
```
LayneHaber, line 1: see previous comments

LayneHaber (changes requested):

packages/contracts/contracts/Router.sol
jakekidd: Shouldn't it match both

packages/contracts/contracts/Router.sol
jakekidd: The way "around it" is to have a sep signer as the remover/recipient of the liquidity, which it seems like we have. Enables routers to at least optionally have a diff signer that they keep even more securely (e.g. don't use it on their router's server, etc).

packages/contracts/contracts/Router.sol
jakekidd: Imo that should be a separate step. We want this operation to be as small as possible to enable liquidity removal at more frequent intervals for regular dividend harvesting. Just my .02

jakekidd: Could have a separate method, like `removeAllLiquidity` or `cleanOutLiquidity` that does all of that, but that adds complexity to this contract and we could just do that in a sep. helper contract or further iteration

packages/contracts/contracts/Router.sol
```diff
@@ -0,0 +1,220 @@
...
+  function prepare(
+    ITransactionManager.PrepareArgs calldata args,
+    address routerRelayerFeeAsset,
+    uint256 routerRelayerFee,
+    bytes calldata signature
+  ) external payable returns (ITransactionManager.TransactionData memory) {
+    if (msg.sender != routerSigner) {
+      SignedPrepareData memory payload = SignedPrepareData({
+        args: args,
+        routerRelayerFeeAsset: routerRelayerFeeAsset,
```
jakekidd, line 138: Won't the relayer fee *always* be in receiving token?

packages/contracts/contracts/Router.sol
jakekidd: shouldn't all of these checks be in the txmanager contract prepare method?

packages/contracts/contracts/Router.sol
jakekidd: Technically I don't think we need this. If the amount stored in the contract is < relayer fee then the relayer would never call this method - they would simulate the call, see that it would reject onchain, and then refuse to call. Same difference, as a require statement would have the same effect.

packages/contracts/contracts/Router.sol
LayneHaber: depends on the underlying chain, some return `1` and it's not mainnet (i.e. Harmony)

packages/contracts/contracts/Router.sol
LayneHaber: if they compromise the owner key though, they can just remove all the relayer fee and liquidity but yeah that's just an effect of permissioning

packages/contracts/contracts/Router.sol
LayneHaber: You have to pay the relayer for each transaction though?

packages/contracts/contracts/Router.sol
LayneHaber: sure it would fail if you tried to transfer more than what existed on the contract, but require statements return much better error messages

packages/contracts/contracts/Router.sol
LayneHaber: yeah i guess they are, but doesn't mean im not a fan of excessive sanity checks in solidity

packages/contracts/contracts/Router.sol
LayneHaber: in the router contract you could specify what asset and just park fees on the contract directly. the relayer fee from the transaction manager would actually be sent to this contract, so we should make sure that we can easily handle that case

packages/contracts/contracts/Router.sol
ArjunBhuptani: Making the remove recipient changeable via owner defeats the whole purpose of restricting the remove process in the first place.

Our current situation:
1. Remove liquidity can be called only by router (safe)
2. Router key is stored via web3signer (safe)
3. We need to expose some way to allow the router to call remove liquidity so we create an admin key/endpoint for it. This endpoint can send liquidity *anywhere* (unsafe)

With `recipient` changeable by owner:
1. Remove liquidity only callable by router key (safe)
2. Router key stored via web3signer (safe)
3. Admin key endpoint exists but funds can only go to address specified as part of setup (safe)
4. Deployer key which controls changing recipient now needs to be stored securely and recipient-changing functions now need to be exposed by admin endpoint (unsafe).

In particular, the above case creates an additional security assumption for router operators who are using a custody solution and expect to enforce that transactions go from custody solution to contract and back *without* any additional need tor keys/security beyond the existing custody system.

Instead of the above, we should explicitly *not* allow changing the recipient once contract is deployed. If routers need to change the recipient, they must then withdraw to the prespecified recipient address, and deploy a new wrapper contract, thus delegating the security entirely to the `recipient` (which is typically going to be a custody system).

ArjunBhuptani: P.s. doing it this way more closely emulates the UX for DeFi liquidity provision which is the most expected user flow and has a lot more tooling build around it

packages/contracts/contracts/Router.sol
sanchaymittal: Router.sol              |    73.33 |    52.94 |    66.67 |    73.91 |... 104,107,148 |
  RouterFactory.sol       |      100 |     87.5 |      100 |      100 |                |
  
  the coverage beforeEach was broken for both contracts. Here is the updated coverage.

packages/contracts/contracts/Router.sol
sanchaymittal: transferring ownership in constructor only, which is going to be two steps process under ProposedOwnable

packages/contracts/contracts/Router.sol
sanchaymittal: here we are getting the chainId from RouterFactory, In RouterFactory we are getting it from TransactionManager only.

packages/contracts/contracts/Router.sol
sanchaymittal: address on all the chains are going to be the same, so it won't help in domain separation.

packages/router/src/lib/operations/prepare.ts
sanchaymittal: even if we check the router contract address, bids will fail at a later stage for relayers.

packages/router/src/lib/operations/prepare.ts
sanchaymittal: It's on router side, so it should be receiving assetId, right?

packages/router/src/router.ts
sanchaymittal: it throws `Redundant double negation` when I tried using it.

packages/txservice/src/chainreader.ts
sanchaymittal: right! we have flag in getContext() called isRouterContract. Need to add those estimates

sanchaymittal: updated

packages/utils/src/gasEstimates.ts
sanchaymittal: pending... Need to fetch this add to gasEstimates @rhlsthrm

packages/contracts/contracts/Router.sol
rhlsthrm: I think we do need to handle relayer fee here @sanchaymittal

packages/contracts/contracts/Router.sol
rhlsthrm: I think that check would add more gas right? Isn't it better to just let it fail.

packages/contracts/contracts/Router.sol
rhlsthrm: I think this should be a super light wrapper contract and not do any of the extra validation on the solidity side tbh.

packages/contracts/contracts/Router.sol
rhlsthrm: I don't think that's necessary. This contract shouldn't have extra checks beyond the underlying contract.

packages/router/src/adapters/contract/index.ts
rhlsthrm: Yes this is definitely in place but it's not that easy because you need to send relayer fees in some cases so there's more business logic that should be in operations.

packages/router/src/bindings/fastify/index.ts
rhlsthrm: Not handled yet, can deal with this later.

packages/router/src/bindings/fastify/index.ts
rhlsthrm: Not handled yet, can deal with this later. Trying to reduce scope still.

packages/router/src/lib/helpers/shared.ts
rhlsthrm: Good call this is true. cc @sanchaymittal

packages/router/src/lib/operations/auction.ts
rhlsthrm: Why "some amount above"?

packages/router/src/lib/operations/cancel.ts
rhlsthrm: Oops, this needs to depend on the chain they want to cancel on cc @sanchaymittal!

packages/router/src/lib/operations/metaTx.ts
rhlsthrm: Can this go into the txService?

packages/router/src/lib/operations/metaTx.ts
rhlsthrm: Yes, it uses Gelato first then falls back here. This is the listening side. This was all quite complicated to architect and build lol.

packages/contracts/contracts/Router.sol
jakekidd: @LayneHaber what would the error message look like w/o the require? Would it not be clear that it failed for that reason? I think that holds the answer

packages/contracts/contracts/Router.sol
jakekidd: agreed ^

packages/contracts/src/errors.ts
jakekidd: agreed, we should either use invalid signature in all cases or remove 022 and replace it with invalid router sig / invalid user sig

obvi former is easier

packages/contracts/src/errors.ts
jakekidd: also agreed

packages/contracts/src/errors.ts
jakekidd: yeah I don't like it. With error codes, won't we know what contract the error comes from? Even if there's a call to another contract within that method (in the case of Router.sol)? if that's the case, this naming seems redundant

packages/router/src/lib/operations/metaTx.ts
jakekidd: chaindata should get integrated into txservice imo, then we can combine that with the caching so it's just one call here `getDecimalsForAsset`

packages/contracts/contracts/Router.sol
jakekidd: Wouldn't it just be `ERC20: transfer amount exceeds balance` ? In which case I think it's descriptive enough

PR #564, opened 2021-11-26
[txservice] Surface error types

jakekidd:
## The Problem

Too much TransactionServiceFailure

## The Solution

No more TransactionServiceFailure

## Checklist

- [ ] Run load tests.
- [ ] Update CHANGELOG.md.

### DIFF
--- a/packages/txservice/src/dispatch.ts
+++ b/packages/txservice/src/dispatch.ts
@@ -17,8 +17,11 @@ import {
   TransactionReplaced,
   TransactionReverted,
   TimeoutError,
-  TransactionServiceFailure,
   TransactionBackfilled,
+  InitialSubmitFailure,
+  TransactionProcessingError,
+  NotEnoughConfirmations,
+  MaxAttemptsReached,
 } from "./error";
 import { ChainConfig, TransactionServiceConfig } from "./config";
 import { ChainRpcProvider } from "./provider";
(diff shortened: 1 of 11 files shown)


packages/txservice/src/dispatch.ts
```diff
@@ -570,26 +568,24 @@ export class TransactionDispatch extends ChainRpcProvider {
...
+        throw new TransactionProcessingError(TransactionProcessingError.reasons.DidNotThrowRevert, method, {
           chainId: this.chainId,
           receipt,
           transaction: transaction.loggable,
         });
       } else if (receipt.confirmations < 1) {
         // Again, should never occur.
-        throw new TransactionServiceFailure("Receipt did not have any confirmations, should have timed out!", {
-          method,
+        throw new TransactionProcessingError(TransactionProcessingError.reasons.InsufficientConfirmations, method, {
```
LayneHaber, line 585: Not a huge fan of the `TransactionProcessingError` either because there's no distinct type matching the distinct reasons. Willing to let it slide iff it's only used for cases where the transaction mining/confirmation breaks

LayneHaber: ideally each error that would have a unique code assigned to it, should also have a unique type

packages/txservice/src/error.ts
```diff
@@ -220,35 +220,104 @@ export class TransactionKilled extends TransactionError {
...
+  constructor(public readonly context: any = {}) {
+    super(
+      "Transaction never submitted: exceeded maximum iterations in initial submit loop.",
+      context,
+      InitialSubmitFailure.type,
+    );
+  }
+}
+
+// These errors should essentially never happen; they are only used within the block of sanity checks.
```
LayneHaber, line 291: willing to let it slide then

LayneHaber (approved):

packages/txservice/src/error.ts
jakekidd: yeah. i think this was the original intention with TransactionServiceFailure: make a grouping of errors used in sanity checks only that we could look out for to indicate extremely unexpected misbehavior. Realistically, these checks exist just to *eliminate* the possibility of a future problem being one of those cases

This error "grouping" feels a lot more specific to that imo

PR #567, opened 2021-11-29
[txservice] Sync providers

jakekidd:
This is a rehash of #482 

I took out the block listener stuff.

UPDATE:
now includes the changes from chainreader integration:
* expand chainreader to include fee calc, price oracle calc
* integrate chainreader into sdk
* use chain reader fee calc in router
* expose getBalance and getDecimalsForAsset methods in sdk
* add NxtpSdk static create method

## The Problem(s)

Just like subgraph, providers sometimes are out of sync or behave weirdly otherwise, giving bad txcount values, gas estimates, gas, etc.

- we need to track how “in-sync” providers are (i.e. their relative “lag”)
- we should never consult a provider that is out-of-sync (rn this is hardcoded as 30 or 40 blocks behind)
- we should consult providers in order of how in-sync they are

Also, neverthrow is annoying. It bogs down implementations, and isn't doing anything anymore

## The Solution

- synchronize providers (just get the block number and compare them) every 60 seconds (this value is configurable)}
- replace wrapper method for `ChainRpcProvider` with execute, which basically hands off the most in-sync provider (and if it fails, it calls the next one, etc).
- added a typed `ProviderCache`. each cached value has ttl (there is also btl, blocks to live, but we can't use this without a block listener).
- caching value `blockPeriod` locally to make wait periods (for polling) match the efficiency/speed (blocks per second/minute/whatever) of the chain.
- removed usage of neverthrow (errors were being thrown 9/10 times anyway).
- added some additional/improved error parsing

## Checklist

- [x] Run load tests.
- [x] Update CHANGELOG.md.


packages/txservice/src/chainreader.ts
```diff
@@ -6,6 +6,9 @@ import { ReadTransaction } from "./types";
 import { ChainRpcProvider } from "./provider";
 import { ConfigurationError, ProviderNotConfigured } from "./error";
 
+// TODO: I do not like that this is generally a passthrough class now - all it handles is the mapping. We should
+// probably just expose a provider getter method and have the consumer call that to access the target ChainRpcProvider
+// directly.
```
LayneHaber, line 11: would you want to include this in this PR or a separate one?

packages/txservice/src/config.ts
```diff
@@ -60,20 +60,26 @@ export const ProviderConfigSchema = Type.Object({
...
   defaultInitialGas: Type.Optional(TIntegerString),
   // Gas station URL, if any, to retrieve current gas price from. If gas station is down or otherwise fails,
   // we'll use the RPC provider's gas as a backup.
-  // Gas station should return a "rapid" gas price within the response.data. See ethereum gasnow API as example.
-  gasStations: Type.Array(Type.String()),
+  // Gas station should return a "rapid" gas price within the response.data.
+  gasStations: Type.Optional(Type.Array(Type.String())),
+  // An integer value by which we will inflate the gas LIMIT that is returned by the provider (flat increase).
+  // Use this if your provider is returning low values and you're getting "out of gas" call exceptions.
+  gasLimitInflation: Type.Optional(Type.Integer()),
```
LayneHaber, line 82: Flat increase in GWEI? or a percentage increase as a buffer?

packages/txservice/src/config.ts
```diff
@@ -109,12 +115,22 @@ const TransactionServiceConfigSchema = Type.Object({
...
 
   /// CHAINS
   // Configuration for each chain that this txservice will be supporting.
   chains: Type.Record(TIntegerString, ChainConfigSchema),
+
+  /// DEBUGGING / DEVELOPMENT
+  // WARNING: Please do not alter these configuration values; they should be used for development and/or debugging
+  // purposes only, and can greatly affect performance.
+  // Whether to log rpc calls.
+  debug_logRpcCalls: Type.Boolean(),
```
LayneHaber, line 133: why the `_`? not very javascript of u

packages/txservice/src/dispatch.ts
```diff
@@ -314,105 +315,123 @@ export class TransactionDispatch extends ChainRpcProvider {
...
+            this.getGasPrice(requestContext),
+            this.determineNonce(attemptedNonces),
+          ]);
+          let { nonce, backfill, transactionCount } = nonceInfo;
+
+          // TODO: Remove hardcoded (exposed gasLimitInflation config var should replace this).
+          let gas = new Gas(gasPrice, gasLimit);
+          if (this.chainId === 42161) {
+            gas = new Gas(gas.baseValue, BigNumber.from(10_000_000));
+          }
```
LayneHaber, line 353: oof wow dont love the hardcoded chain id exceptions, but totally understand why they happen. adding a comment here would be good (i already can't remember, and i know this isn't from this PR)

packages/txservice/src/dispatch.ts
```diff
@@ -455,14 +479,35 @@ export class TransactionDispatch extends ChainRpcProvider {
...
+    try {
+      const response = await this.sendTransaction(transaction);
+      // Add this response to our local response history.
+      if (transaction.hashes.includes(response.hash)) {
+        // Duplicate response? This should never happen.
+        throw new TransactionProcessingError(TransactionProcessingError.reasons.DuplicateHash, method, {
+          chainId: this.chainId,
+          response,
+          transaction: transaction.loggable,
+        });
```
LayneHaber, line 493: this would be such a weird error so if there's *anything* else that should be logged (request context/method context maybe) along with the error it should be included in the error context

packages/txservice/src/dispatch.ts
```diff
@@ -516,11 +548,32 @@ export class TransactionDispatch extends ChainRpcProvider {
...
+      const receipt = await this.confirmTransaction(transaction, 1);
+
+      // Sanity checks.
+      if (receipt.status === 0) {
+        // This should never occur. We should always get a TransactionReverted error in this event.
+        throw new TransactionProcessingError(TransactionProcessingError.reasons.DidNotThrowRevert, method, {
+          chainId: this.chainId,
+          receipt,
+          transaction: transaction.loggable,
+        });
```
LayneHaber, line 563: same as above -- if there's anything else you think should be logged here include it in the error context

packages/txservice/src/dispatch.ts
```diff
@@ -641,9 +662,11 @@ export class TransactionDispatch extends ChainRpcProvider {
...
-    const confirmResult = await this.confirmTransaction(transaction, this.confirmationsRequired, timeout);
-    if (confirmResult.isErr()) {
-      // No other errors should occur during this confirmation attempt. This could occur during a reorg.
+    let receipt: providers.TransactionReceipt;
+    try {
+      receipt = await this.confirmTransaction(transaction, this.confirmationsRequired, timeout);
+    } catch (error) {
+      // No other errors should normally occur during this confirmation attempt. This could occur during a reorg.
       throw new NotEnoughConfirmations(
         this.confirmationsRequired,
```
LayneHaber, line 671: we should log the error with `jsonifyError` just in case

packages/txservice/src/dispatch.ts
```diff
@@ -675,56 +691,72 @@ export class TransactionDispatch extends ChainRpcProvider {
...
       chainId: this.chainId,
       receipt: {
-        gasUsed: receipt.gasUsed.toString(),
-        transactionHash: receipt.transactionHash,
-        blockHash: receipt.blockHash,
+        transactionHash: transaction.receipt.transactionHash,
+        confirmations: transaction.receipt.confirmations,
+        blockNumber: transaction.receipt.blockNumber,
       },
+      transactionExecutionTime: Date.now() - transaction.timestamp,
```
LayneHaber, line 704: nice, i like this

packages/txservice/src/dispatch.ts
```diff
@@ -675,56 +691,72 @@ export class TransactionDispatch extends ChainRpcProvider {
...
     const { requestContext, methodContext } = createLoggingContext(this.bump.name, transaction.context);
-    if (transaction.attempt >= Transaction.MAX_ATTEMPTS) {
-      const error = new MaxAttemptsReached(transaction.attempt, {
-        gasPrice: `${utils.formatUnits(transaction.gas.price, "gwei")} gwei`,
-        transaction: transaction.loggable,
-      });
-      throw error;
-    } else if (transaction.bumps >= transaction.hashes.length) {
+    if (
+      transaction.bumps >= transaction.hashes.length ||
```
LayneHaber, line 718: shouldn't it be `>` rather than `>=` ? dont see why having bumps `==` hashes is a problem (seems like you would want a new hash for each bump)

packages/txservice/src/error.ts
```diff
@@ -356,14 +350,23 @@ export const parseError = (error: any): NxtpError => {
     return new TransactionReverted(TransactionReverted.reasons.GasExceedsAllowance, undefined, context);
   } else if (
     message.match(
-      /another transaction with same nonce|same hash was already imported|transaction nonce is too low|nonce too low|already known/,
+      /another transaction with same nonce|same hash was already imported|transaction nonce is too low|nonce too low|oldnonce/,
```
LayneHaber, line 353: should there be a space? have we seen this string? what about `already known` -- why take it out?

packages/txservice/src/provider.ts
```diff
@@ -159,24 +179,68 @@ export class ChainRpcProvider {
...
+      const _receipts: Promise<providers.TransactionReceipt | null>[] = transaction.responses.map(async (response) => {
+        try {
+          return await this.getTransactionReceipt(response.hash);
+        } catch (error) {
+          errors.push(error);
+          return null;
+        }
+      });
+      // Wait until all the 'receipts' (or errors) have been pushed to the list.
+      const receipts = await Promise.all(_receipts);
```
LayneHaber, line 209: why not just filter out the `null` return values here rather than handling it below?

packages/txservice/src/provider.ts
```diff
@@ -187,17 +251,17 @@ export class ChainRpcProvider {
...
    */
-  public readTransaction(tx: ReadTransaction): ResultAsync<string, TransactionError> {
-    return this.resultWrapper<string>(true, async () => {
+  public async readTransaction(tx: ReadTransaction): Promise<string> {
+    return this.execute<string>(false, async (provider: SyncProvider) => {
       try {
-        return await this.provider.call(tx);
+        if (this.signer) {
+          this.signer.connect(provider);
+          return await this.signer.call(tx);
```
LayneHaber, line 261: should we just do `await this.signer.connect(provider).call(tx)` ?

packages/txservice/src/provider.ts
```diff
@@ -216,167 +280,133 @@ export class ChainRpcProvider {
...
+    return this.execute(false, async (provider: SyncProvider) => {
+      // This call will prepare the transaction params for us (hexlify tx, etc).
+      const args = provider.prepareRequest("estimateGas", { transaction });
+      const result = await provider.send(args[0], args[1]);
+      try {
+        return BigNumber.from(result).add(gasLimitInflation ? BigNumber.from(gasLimitInflation) : 0);
+      } catch (error) {
+        throw new GasEstimateInvalid(result, {
+          error: error.message,
+        });
```
LayneHaber, line 297: would this function cause any problems with the error message returned from potentially reverted transactions when you estimate gas?

packages/txservice/src/provider.ts
```diff
@@ -461,128 +492,197 @@ export class ChainRpcProvider {
...
+      return this.cache.data.transactionCount;
     }
 
-    return this.resultWrapper<number>(true, async () => {
-      const value = await this.signer!.getTransactionCount(blockTag);
-      this.cachedTransactionCount = { value, timestamp: Date.now() };
-      return value;
+    return this.execute<number>(true, async (provider: SyncProvider) => {
+      const transactionCount = await provider.getTransactionCount(await this.signer!.getAddress(), blockTag);
+      this.cache.set({ transactionCount });
```
LayneHaber, line 533: how long does this live?

packages/txservice/src/provider.ts
```diff
@@ -461,128 +492,197 @@ export class ChainRpcProvider {
...
+    // If we need a signer, check to ensure we have one.
+    if (needsSigner) {
+      this.checkSigner();
+    }
+    if (!this.blockPeriod) {
+      // This block should only be called once.
+      const currentBlock = await this.getBlock("latest");
+      const previousBlock = await this.getBlock(currentBlock.parentHash);
+      this.blockPeriod = currentBlock.timestamp - previousBlock.timestamp;
+    }
```
LayneHaber: nice, so presumably this is the cache period?

packages/txservice/src/provider.ts
```diff
@@ -461,128 +492,197 @@ export class ChainRpcProvider {
...
+    // Shuffle isn't applied to lead provider - instead, we just guarantee that it's in the first position.
+    syncedProviders.forEach((p) => {
+      p.priority =
+        p.lag -
+        (this.leadProvider && p.url === this.leadProvider.url ? 1 : Math.random()) -
+        p.cps / this.config.maxProviderCPS -
+        p.reliability * 2 +
+        p.avgExecTime;
+    });
+    return syncedProviders.sort((a, b) => a.priority - b.priority);
```
LayneHaber, line 698: do we want to do something similar with the unsynced providers being used at the back of the list, or simply *only* include the synced providers?

LayneHaber (commented):
this looks good, but we should make sure to run load tests on mainnet so that we can ensure there aren't any notable degradations

packages/txservice/src/chainreader.ts
jakekidd: separate for now. ran some ideas by rahul, we'll catch up on this in standup tmrw

packages/txservice/src/config.ts
jakekidd: yo great q

packages/txservice/src/config.ts
jakekidd: lmao I actually made a note of this in original PR. I just wanted to communicate that it's a no-touchy thing. Like, it's literally to turn on/off console logs in SyncProvider...

lmk if I should remove

packages/txservice/src/dispatch.ts
jakekidd: Yeah but it looks like it might be outdated? not sure I will look into this

packages/txservice/src/error.ts
jakekidd: no space, this is from matic providers I believe

already known is what happens when Fallback Provider spams every provider with a new sent TX
but by the time it spams 1 or more of the providers they are like 'I already seen it dude'

AlreadyKnown is now a sep. error and it's only used internally bc it shouldn't be raised to consumer level.

jakekidd: see ServerError.reasons.TransactionAlreadyKnown below

packages/txservice/src/provider.ts
jakekidd: yes

packages/txservice/src/provider.ts
jakekidd: it's not a perfect system, but in reality the error parsing happens twice - once in SyncProvider and once in ChainRpcProvider.execute (the new wrapper). This is bc we want to do that regardless of whether the error comes from the block in ChainRpcProvider (some logic fail) or from ethers/provider itself.

So either way, the error will be parsed the same way it always has. And when it's called a second time, we'll recognize it as an NxtpError and just pass it up as-is

packages/txservice/src/provider.ts
jakekidd: 2 seconds, see the hardcoded TTL in the constructor lol

we can make this configurable/based on block period later

packages/txservice/src/provider.ts
jakekidd: yeah could be. I can add a TODO for this

jakekidd: also thinking that putting this setup in execute is tacky.

Going to move this to a sep method and call it in constructor (without await obvi)... and just leave the default as 1 sec until it gets set.

packages/txservice/src/provider.ts
jakekidd: sync is all relative here. technically speaking I could be running with a single provider that's 300 blocks behind but i wouldn't know it, bc I have no other provider to compare it to.

as it is, lag is the most influential factor here by far. being 10 blocks behind another provider basically guarantees that provider will be chosen after the one that's ahead.

Buuut if you'd like, I could choose to do what I did with fallback subg. Make the ..MAX_LAG value effectively a line in the sand, and have all providers that are lagging < than that in front and concatenate all laggy providers.

So we guarantee, regardless of any metrics, that we always try those "in-sync" providers first

packages/txservice/src/provider.ts
jakekidd: genius

packages/txservice/src/config.ts
jakekidd: I was thinking we could have a suite of `debug_` options

debug_listProviders
debug_tryThisMethod
debug_logEveryTimeYouBreathe

etc etc

jakekidd: Ik we want to debug log everything but the console will have a literal stroke if I turn on debug_logRpcCalls

packages/txservice/src/dispatch.ts
jakekidd: nope not outdated, same thing as it was previously lol

packages/txservice/src/dispatch.ts
jakekidd: the response is everything here
also `transaction.loggable`
all we would need is hash in this case.

i've never seen this happen, it's just to enforce that transaction.hashes is a set of unique hashes... i imagine you could only break this by unit testing improperly

packages/txservice/src/dispatch.ts
jakekidd: will do

packages/txservice/src/config.ts
jakekidd: this is a flat increase in gas lol. gwei is gas price, this is for gas amount.

for example, if we want to inflate by 10_000_000 on arbitrum

jakekidd: not related to gas price

and in fact i'm going to change "defaultInitialGas" above to reflect that

packages/txservice/src/dispatch.ts
jakekidd: it actually should logically be `===`. The `>` is just my paranoia showing.

we bump *before* sending an additional tx, not after. and we only bump after sending the initial one.

- send initial (hashes length 1)
- bump (hashes length 1, bumps 1)
- send (2, 1)
- bump (2, 2)
...

dont want to bump if we've already bumped for the current hash.

jakekidd:
attaching these load tests... they were run on the branch that was merged into this one, so equivalent to commit fb217da

txs 500
[log_output_23.txt](https://github.com/connext/nxtp/files/7631119/log_output_23.txt)

PR #570, opened 2021-12-01
Subgraph updates

LayneHaber:
## The Problem

- Need to track investment made by routers

## The Solution

- Add convenience fields to analytics subgraph to make it easy to calculate ROI via:
```sh
roi = 100% * [sum_across_chains(locked + amount - supplied) / sum_across_chains(supplied)]
```

## Checklist

- [ ] Run load tests.
- [ ] Update CHANGELOG.md.


packages/subgraph/src/v1-analytics/mapping.ts
```diff
@@ -257,6 +297,7 @@ export function handleTransactionFulfilled(event: TransactionFulfilled): void {
       hourlyMetric.liquidity = assetBalance.amount;
     }
 
+    hourlyMetric.sendingTxCount = hourlyMetric.receivingTxCount.plus(BigInt.fromI32(1));
```
wanglonghong: why receivintTxCount + 1 for sendingTxCount???

I think it should be like this
`hourlyMetric.sendingTxCount = hourlyMetric.sendingTxCount .plus(BigInt.fromI32(1));`

packages/subgraph/src/v1-analytics/mapping.ts
```diff
@@ -291,8 +332,52 @@ export function handleTransactionCancelled(event: TransactionCancelled): void {
...
     }
     assetBalance.amount = assetBalance.amount.plus(transaction.amount);
+    assetBalance.locked = assetBalance.locked.sub(transaction.amount);
     assetBalance.save();
   }
+
+  // update metrics
+  let timestamp = event.block.timestamp.toI32();
+
+  let hour = timestamp / 3600; // rounded
```
jakekidd, line 346: these timestamps are in seconds (not ms)?

packages/subgraph/src/v1-analytics/mapping.ts
```diff
@@ -291,8 +332,52 @@ export function handleTransactionCancelled(event: TransactionCancelled): void {
...
+  let hourStartTimestamp = hour * 3600;
```
jakekidd, line 347: lol isn't this the same value as `timestamp`

packages/subgraph/src/v1-analytics/mapping.ts
```diff
@@ -291,8 +332,52 @@ export function handleTransactionCancelled(event: TransactionCancelled): void {
...
+    hourlyMetric.volume = BigInt.fromI32(0);
+    hourlyMetric.liquidity = BigInt.fromI32(0);
+    hourlyMetric.txCount = BigInt.fromI32(0);
+    hourlyMetric.sendingTxCount = BigInt.fromI32(0);
+    hourlyMetric.receivingTxCount = BigInt.fromI32(0);
+    hourlyMetric.cancelTxCount = BigInt.fromI32(0);
+  }
+
+  let day = timestamp / 86400; // rounded
+  let dayStartTimestamp = day * 86400;
```
jakekidd, line 365: this math isn't right - we should be flooring down to the beginning of the day... right?
so
```ts
dayStartTimestamp = day - (timestamp % 86400);
```

packages/subgraph/src/v1-analytics/schema.graphql
```diff
@@ -90,6 +92,12 @@ type HourlyMetric @entity {
   liquidity: BigInt!
   # hourly number of transactions per asset
```
jakekidd, line 93: if my impression is correct, the more data we store, the more costly the subgraph... so couldn't this value, `txCount` be calculated as the sum of both the `sending` and the `receiving` ?

packages/subgraph/src/v1-analytics/mapping.ts
LayneHaber: the timestamps from the block are, yeah

packages/subgraph/src/v1-analytics/mapping.ts
LayneHaber: yes

packages/subgraph/src/v1-analytics/mapping.ts
```diff
@@ -194,19 +220,26 @@ export function handleTransactionFulfilled(event: TransactionFulfilled): void {
...
     assetBalance.amount = assetBalance.amount.plus(transaction.amount);
-    assetBalance.save();
+  } else {
+    // router releases locked liquidity on receiver fulfill
+    assetBalance.locked = assetBalance.locked.sub(transaction.amount);
   }
+  assetBalance.save();
 
   // update metrics
   let timestamp = event.block.timestamp.toI32();
```
jakekidd, line 245: Ik you didn't write this code, but see my comment about calculating the `dayStartTimestamp` and compare that to the `hourStartTimestamp` below

https://github.com/connext/nxtp/pull/570/files#r759804355

packages/subgraph/src/v1-analytics/schema.graphql
LayneHaber: the more costly in terms of time per request (eventually perhaps for the actual query when it becomes a fully decentralized network, but we're a bit far off from that). i don't think the latency here matters so much, but you could count it as the sending and receiving. i prefer the convenience/clarity here though

packages/subgraph/src/v1-analytics/mapping.ts
LayneHaber: i think the timestamp is really just to show when we were able to start collecting this data. we can't really floor it down to the beginning of the day because this should be used to show how much of the day these metrics are missing (they are only updated as the subgraph gets new entities)

packages/subgraph/src/v1-analytics/mapping.ts
LayneHaber: https://github.com/connext/nxtp/pull/570#discussion_r759807650

packages/subgraph/src/v1-analytics/mapping.ts
jakekidd: I'm still missing something here. bc according to this `dayStartTimestamp === timestamp`

You are just dividing by 86400 then multiplying by 86400 again

packages/subgraph/src/v1-analytics/schema.graphql
jakekidd: gotcha kk nw

packages/subgraph/src/v1-analytics/mapping.ts
LayneHaber: jk i was wrong, wang was right. rounding makes them different

jakekidd: apparently it is not, the hour value is rounded

jakekidd (approved):
lgtm

packages/subgraph/src/v1-analytics/mapping.ts
```diff
@@ -31,9 +31,16 @@ export function handleLiquidityAdded(event: LiquidityAdded): void {
...
     assetBalance.router = router.id;
     assetBalance.amount = new BigInt(0);
+    assetBalance.supplied = new BigInt(0);
+    assetBalance.locked = new BigInt(0);
   }
   // add new amount
   assetBalance.amount = assetBalance!.amount.plus(event.params.amount);
+
+  // add invested
+  assetBalance.supplied = assetBalance.supplied.plus(event.params.amount);
```
jakekidd, line 41: nit: needs an explicit unwrap above and none here?

LayneHaber: i dont think it does -- in fact i think the one above is unnecessary since its assigned in the case where `assetBalance` does not exist

packages/subgraph/src/v1-analytics/mapping.ts
```diff
@@ -161,11 +174,24 @@ export function handleTransactionPrepared(event: TransactionPrepared): void {
...
+    assetBalance.locked = new BigInt(0);
+  }
+
   // router is providing liquidity on receiver prepare
   if (chainId == transaction.receivingChainId) {
     let assetBalanceId = transaction.receivingAssetId.toHex() + "-" + event.params.router.toHex();
     let assetBalance = AssetBalance.load(assetBalanceId);
     assetBalance.amount = assetBalance.amount.minus(transaction.amount);
+    assetBalance.locked = assetBalance.locked.add(transaction.amount);
     assetBalance.save();
```
jakekidd, line 195: we don't save unless chainId == transaction.receivingChainId ..?

jakekidd: if this is the case, the whole "get asset balance" above should maybe be in this block

packages/subgraph/src/v1-analytics/mapping.ts
```diff
@@ -291,8 +332,52 @@ export function handleTransactionCancelled(event: TransactionCancelled): void {
       assetBalance.assetId = transaction.receivingAssetId;
       assetBalance.router = event.params.router.toHex();
       assetBalance.amount = new BigInt(0);
+      assetBalance.supplied = new BigInt(0);
+      assetBalance.locked = new BigInt(0);
```
jakekidd, line 336: `assetBalance.locked` will be set to `-transaction.amount` in this case.

the below `assetBalance.locked = assetBalance.locked.sub(transaction.amount);` gets executed after this block...

jakekidd: is that expected behavior?

jakekidd: unless it's like, an unsigned int and cant below 0 lol

LayneHaber: it shouldn't ever fall into the case where its not defined, ill use a force unwrap to show that instead.

packages/subgraph/src/v1-analytics/mapping.ts
```diff
@@ -291,8 +332,52 @@ export function handleTransactionCancelled(event: TransactionCancelled): void {
...
+  let timestamp = event.block.timestamp.toI32();
+
+  let hour = timestamp / 3600; // rounded
+  let hourStartTimestamp = hour * 3600;
+
+  let hourIDPerAsset = hour.toString() + "-" + transaction.receivingAssetId.toHex();
+
+  let hourlyMetric = HourlyMetric.load(hourIDPerAsset.toString());
+  if (hourlyMetric === null) {
+    hourlyMetric = new HourlyMetric(hourIDPerAsset.toString());
```
wanglonghong, line 353: let hourIDPerAsset = hour.toString().concat("-").concat(transaction.receivingAssetId.toHex());

  let hourlyMetric = HourlyMetric.load(hourIDPerAsset);
  if (hourlyMetric === null) {
    hourlyMetric = new HourlyMetric(hourIDPerAsset);
    
    I think it could be like this as well. hourIDPerAsset is string object right now. any reason to use `toString` here?

packages/subgraph/src/v1-analytics/mapping.ts
```diff
@@ -291,8 +332,52 @@ export function handleTransactionCancelled(event: TransactionCancelled): void {
...
+  }
+
+  let day = timestamp / 86400; // rounded
+  let dayStartTimestamp = day * 86400;
+
+  let dayIDPerAsset = day.toString() + "-" + transaction.receivingAssetId.toHex();
+
+  let dayMetric = DayMetric.load(dayIDPerAsset.toString());
+  if (dayMetric === null) {
+    dayMetric = new DayMetric(dayIDPerAsset.toString());
```
wanglonghong, line 371: same here to above hourIDPerAsset.

wanglonghong (changes requested):

packages/subgraph/src/v1-analytics/mapping.ts
LayneHaber: ahh yep!

packages/subgraph/src/v1-analytics/mapping.ts
LayneHaber: ah wow actually this made me catch that i wasn't handling the difference in sending/receiving asset balances properly in the id. but yeah, we only save if its the receiving chain cus thats the only time we update on prepare

wanglonghong (approved):

rhlsthrm (approved):

PR #596, opened 2021-12-07
[utils] Subgraph health

### LINKED ISSUE #587: Subgraph health endpoint caller utility
Since the subgraph health endpoint is an atypical endpoint that calls a static graph url and only needs to specify which subgraph by name, we should just make a utility with a hardcoded query string for this.

jakekidd:
## The Problem

There's a really neat API for getting subgraph "health" on the graph. We should use this in place of or as a preference to the current method for determining subgraph sync status

## The Solution

Fixes #587 

Using this subg health method as the preferred way of getting subgraph health... with the current method (using sep. provider block) as a backup option in `FallbackSubgraph.sync()`

You can also import and call this method directly from utils.

It uses axios to call with query directly, because this query will *not* go to our actual subgraphs, but to the graph API itself.

## Checklist

- [ ] remove old subg health query?
- [ ] Run load tests.
- [x] Update CHANGELOG.md.


packages/utils/src/fallbackSubgraph.ts
```diff
@@ -31,28 +34,28 @@ type Sdk<T extends SdkLike> = {
...
 
-  private readonly sdks: Sdk<T>[];
+  private readonly subgraphs: Subgraph<T>[];
 
   /**
    * Returns boolean representing whether at least one available subgraph is in sync.
    */
   public get inSync(): boolean {
-    return this.sdks.some((sdk) => sdk.record.synced);
+    return this.subgraphs.some((sdk) => sdk.record.synced);
```
wanglonghong, line 50: > `return this.subgraphs.some((subgraph) => subgraph.record.synced);`

how about this change?

packages/utils/src/fallbackSubgraph.ts
```diff
@@ -31,28 +34,28 @@ type Sdk<T extends SdkLike> = {
...
   }
 
   /**
    * Returns boolean indicating whether we've synchronized the subgraphs at all (indicating
    * whether the records are in fact representative).
    */
   public get hasSynced(): boolean {
-    return this.sdks.some((sdk) => sdk.record.syncedBlock !== -1 && sdk.record.latestBlock !== -1);
+    return this.subgraphs.some((sdk) => sdk.record.syncedBlock !== -1 && sdk.record.latestBlock !== -1);
```
wanglonghong, line 58: > `return this.subgraphs.some((subgraph) => subgraph.record.syncedBlock !== -1 && subgraph.record.latestBlock !== -1);`

same as above

packages/utils/src/fallbackSubgraph.ts
```diff
@@ -73,18 +76,22 @@ export class FallbackSubgraph<T extends SdkLike> {
...
       record: {
-        synced: true,
-        latestBlock: -1,
-        syncedBlock: -1,
-        // Setting maxLag + 1 as default to ensure we don't use the subgraph in this current state
-        // by virtue of this metric (sync() must be called first).
-        lag: this.maxLag + 1,
         // Typically used for logging, distinguishing between which subgraph is which, so we can monitor
         // which ones are most in sync.
         uri: uri.replace("https://", "").split(".com")[0],
```
wanglonghong, line 88: `uri: uri.replace("https://", "").split(".com")[0],`

I think it could be updated later with regex so it will look better. especially for different types of domains such as https://thegraph.io.

rhlsthrm: Wow, yes this is pretty bad we should not be using `.com` for things.

packages/utils/src/fallbackSubgraph.ts
wanglonghong: done in commit: f8143453b7eb20adc03da959fd87e9554714bbcb

packages/utils/src/fallbackSubgraph.ts
wanglonghong: fixed in commit: f8143453b7eb20adc03da959fd87e9554714bbcb

PR #599, opened 2021-12-09
[txservice] Expose txservice config; parse and apply defaults internally

jakekidd:
## The Problem

- txservice config unexposed, unable to access except by named property in router
- parsing code for passing in the chain config for txs is redundant and in many places

## The Solution

- Expose config for txservice
- Parse config internally in txservice
- Specify config vars for txservice for all chains by passing in by using "all" or "default" under "chainConfig" in the config.json
- Any config vars except for providers may be passed into "all"/"default": and they will be overwritten by config vars belonging to specific chains.
- Simplified config passed to children objects in txservice
- better naming for txservice config vars (mostly around "gasPrice", etc)

## Checklist

- [ ] Make sure backwards compat!
- [ ] Run load tests.
- [ ] **ACTUALLY UPDATE DOCUMENTATION TO SHOW CONFIG**
- [ ] Update CHANGELOG.md.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Next Release
 
+- [txservice] Update / expose config vars
+
 # 0.0.99
 
 - [txservice] Fix error parsing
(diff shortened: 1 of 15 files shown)


jakekidd:
Unfortunately I don't think the "all" / "default" feature will work as-is. Some other peripherals may throw errors due to that string not being a chain id

PR #600, opened 2021-12-09
[router] Hotfix: handle setting gas price vs eip-1559 gas fee based on type

jakekidd:
## The Problem

- web3signer was presuming we were sending a legacy transaction
- ethers was secretly converting it to a eip-1559 tx whenever possible

## The Solution

- added check for the type of tx in web3signer (0 = legacy; 2 = eip-1559)
- force all txs to be legacy for now, bc ethers conversion to eip-1559 is really gas-inefficient.

## Checklist

- [ ] Run load tests.
- [x] Update CHANGELOG.md.

### DIFF
--- a/packages/router/src/adapters/web3signer/index.ts
+++ b/packages/router/src/adapters/web3signer/index.ts
@@ -54,16 +54,27 @@ export class Web3Signer extends Signer {
 
   public async signTransaction(transaction: providers.TransactionRequest): Promise<string> {
     const tx = await utils.resolveProperties(transaction);
-    const baseTx: utils.UnsignedTransaction = {
-      to: tx.to || undefined,
-      nonce: tx.nonce ? BigNumber.from(tx.nonce).toNumber() : undefined,
-      gasLimit: tx.gasLimit || undefined,
-      gasPrice: tx.gasPrice || undefined,
-      data: tx.data || undefined,
-      value: tx.value || undefined,
-      chainId: tx.chainId || undefined,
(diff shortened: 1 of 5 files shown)


packages/router/src/adapters/web3signer/index.ts
```diff
@@ -54,16 +54,27 @@ export class Web3Signer extends Signer {
...
+      {
+        to: tx.to || undefined,
+        nonce: tx.nonce ? BigNumber.from(tx.nonce).toNumber() : undefined,
+        gasLimit: tx.gasLimit || undefined,
+        data: tx.data || undefined,
+        value: tx.value || undefined,
+        chainId: tx.chainId || undefined,
+      },
+      // If an EIP-1559 transaction, use the EIP-1559 specific fields.
+      tx.type && tx.type === 2
```
LayneHaber, line 67: this tuple condition will be problematic if there *is* a 0 type assigned (the first `tx.type` is always false). maybe do `typeof tx.type`?

LayneHaber (changes requested):
quick logic adjustment

packages/router/src/adapters/web3signer/index.ts
jakekidd: great catch, i'll just check if it's === null or === undefined

jakekidd: or just check `tx.type === 2` without the first part lol

rhlsthrm (approved):

PR #607, opened 2021-12-11
[txservice] Code coverage improvements; separation of interests; config fixes

jakekidd:
- Separate out classes from types, put all of those into a shared/ directory (could rename this to utils?)
- More code coverage!
- Expose txservice config; parse and apply defaults internally
  - Specify config vars for txservice for all chains by passing in by using "all" or "default" under "chainConfig" in the config.json
  - Any config vars except for providers may be passed into "all"/"default": and they will be overwritten by config vars belonging to specific chains.
  - Simplified config passed to children objects in txservice
  - better naming for txservice config vars (mostly around "gasPrice", etc)
  
NOTE: Unfortunately I don't think the "all" / "default" feature in config will work as-is. Some other peripherals may throw errors due to that string not being a chain id number in chainConfig

## Checklist

- [x] Make sure backwards compat!
- [ ] Run load tests.
- [ ] **ACTUALLY UPDATE DOCUMENTATION TO SHOW CONFIG CHANGES**
- [x] Update CHANGELOG.md.


packages/router/src/config.ts
```diff
@@ -104,7 +104,6 @@ export const getDeployedMulticallContract = (chainId: number): { address: string
 export const TChainConfig = Type.Object({
   providers: Type.Array(Type.String()),
   confirmations: Type.Number({ minimum: 1 }),
```
rhlsthrm, line 106: Shouldn't we have the exposure for the tx service config in here?

rhlsthrm (approved):

rhlsthrm:
Not fully sure how the config changes map into here but we need this code coverage so I'm going to merge this.

PR #618, opened 2021-12-15
Undefined in logs

LayneHaber:
## The Problem

- `config.chainConfig` may not have `chainId`

## The Solution

- Handle undefined case

## Checklist

- [ ] Run load tests.
- [x] Update CHANGELOG.md.

### DIFF
--- a/packages/router/src/bindings/prices/index.ts
+++ b/packages/router/src/bindings/prices/index.ts
@@ -22,10 +22,10 @@ export const bindPrices = async () => {
         const cachingAssetIdOnMainnet = await getMainnetEquivalent(asset.assetId, asset.chainId);
         const cachingTokenChainId = cachingAssetIdOnMainnet ? 1 : asset.chainId;
         const cachingTokenAssetId = cachingAssetIdOnMainnet ? cachingAssetIdOnMainnet : asset.assetId;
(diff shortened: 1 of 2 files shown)


packages/router/src/bindings/prices/index.ts
```diff
@@ -22,10 +22,10 @@ export const bindPrices = async () => {
         const cachingAssetIdOnMainnet = await getMainnetEquivalent(asset.assetId, asset.chainId);
         const cachingTokenChainId = cachingAssetIdOnMainnet ? 1 : asset.chainId;
         const cachingTokenAssetId = cachingAssetIdOnMainnet ? cachingAssetIdOnMainnet : asset.assetId;
-        if (!chainAssetMap.get(cachingTokenChainId)) {
+        if (!chainAssetMap.has(cachingTokenChainId)) {
```
jakekidd, line 25: why was `get` not working before? wouldn't it have returned undefined if didn't exist?

no possibility of it returning 0 so idk if i see the problem lol, lmk

jakekidd: just curious. `has` is better ofc

jakekidd: oh bc it's an array

jakekidd (approved):
lgtm

PR #629, opened 2021-12-17
[sdk] Throw valid auction errors as-is

jakekidd:
## The Problem

We were snuffing out `NoBids` errors with `UnknownAuctionError`

## The Solution

- Added property to `AuctionError` "isAuctionError" (equal to true)
- Wrapped the whole segment of the `getTransferQuote` method in a try/catch, instead of multiple try/catches
- When we catch an error, we check `error.isAuctionError` to see if we should throw the error as-is.

## Checklist

- [ ] Run load tests.
- [x] Update CHANGELOG.md.

### DIFF
--- a/packages/sdk/src/sdkBase.ts
+++ b/packages/sdk/src/sdkBase.ts
@@ -32,6 +32,7 @@ import {
   getReceiverAmount,
   ChainData,
   StatusResponse,
+  AuctionPayload,
 } from "@connext/nxtp-utils";
 import { Interface } from "ethers/lib/utils";
 import { abi as TransactionManagerAbi } from "@connext/nxtp-contracts/artifacts/contracts/TransactionManager.sol/TransactionManager.json";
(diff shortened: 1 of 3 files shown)


packages/sdk/src/sdkBase.ts
```diff
@@ -788,25 +791,28 @@ export class NxtpSdkBase {
           return data;
         }),
       );
+
+      const validBids = receivedBids.filter((x) => typeof x !== "string") as AuctionResponse[];
+      const invalidBids = receivedBids.filter((x) => typeof x === "string") as string[];
+
+      if (validBids.length === 0) {
+        throw new NoValidBids(transactionId, payload, invalidBids.join(","), receivedBids);
```
rhlsthrm, line 799: This error also needs to be `isAuctionError = true` right?

rhlsthrm (changes requested):

packages/sdk/src/sdkBase.ts
jakekidd: it should inherit that property

rhlsthrm (approved):

rhlsthrm:
Pls fix conflicts.

LayneHaber (changes requested):
Looks good, but needs unit tests

sanchaymittal:

PR #634, opened 2021-12-17
[sdk] fees breakdown at transferQuote

sanchaymittal:
closes: #604 
- add: fees break down on transferQuote response
- update: receiverAmount calc on estimateFee at sdk

## Checklist

- [ ] Run load tests.
- [ ] Update CHANGELOG.md.

### DIFF
--- a/packages/sdk/src/sdkBase.ts
+++ b/packages/sdk/src/sdkBase.ts
@@ -402,11 +402,7 @@ export class NxtpSdkBase {
     });
 
     // calculate router fee
-    const { receivingAmount: receiverAmount, routerFee } = await getReceiverAmount(
-      amount,
-      inputDecimals,
-      outputDecimals,
-    );
+    const { receivingAmount: swapAmount, routerFee } = await getReceiverAmount(amount, inputDecimals, outputDecimals);
 
     // calculate gas fee
     const gasFee = await this.estimateFeeForRouterTransfer(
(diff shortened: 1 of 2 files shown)


packages/sdk/src/sdkBase.ts
```diff
@@ -436,12 +432,14 @@ export class NxtpSdkBase {
       methodContext,
     );
 
-    const totalFee = gasFee.add(relayerFee).add(routerFee);
+    const totalGasFee = gasFee.add(relayerFee).add(gasFee);
+    const totalFee = totalGasFee.add(routerFee);
+    const receiverAmount = BigNumber.from(swapAmount).sub(totalGasFee);
```
jakekidd, line 437: Ok, so the routerFee has already been applied to the returned `receivingAmount` (here called `swapAmount`), so we only need to apply the totalGasFee. Right?

And I guess `totalFee` is just used for logging below?

packages/sdk/src/sdkBase.ts
```diff
@@ -806,7 +804,7 @@ export class NxtpSdkBase {
       return BigNumber.from(a.bid.amountReceived).gt(b.bid.amountReceived) ? -1 : 1;
     })[0];
 
-    return { ...chosen, metaTxRelayerFee };
+    return { ...chosen, totalFee, metaTxRelayerFee, routerFee };
```
jakekidd, line 807: Shouldn't the *actual* total fee and router fee returned here come from the bid itself? Ideally, it should be the same as our internal estimate above. But we should definitely be returning what the bid is really offering, just in case it is different for some reason.

jakekidd: In fact, it could easily be different if gas price or token price has changed slightly since we did the estimate.

jakekidd (commented):
Left some comments to address, ty

jakekidd (approved):

PR #655, opened 2021-12-22
[txservice] Check that transaction exists when mine timeout occurs.

jakekidd:
## The Problem

Problem assessment thread: https://discord.com/channels/454734546869551114/860482380321652766/922567480080232500

TL;DR:
- A gas spike happened, we couldn't bump over hard maximum and resubmit.
- A tx "disappeared", meaning the blockade nonce had no transaction.... and since we had bumped over maximum gas, we threw an error. What we should have done is resubmitted at the same gas price since the transaction didn't actually exist on chain!*
- A reboot fixed it, but we shouldn't rely on reboots, especially if we have 64 other txs in the queue.

*(*Idk how it's possible, but it happens. Maybe providers don't broadcast a tx properly, a reorg occurs, geth pending tx pool overflows, who knows? But if I check the tx hash on tenderly, I can see that that tx does not exist, even though the logs say it *was* sent to chain.)*

## The Solution

Timeout handling logic was improved. Some of this logic was pre-existing, but here's the whole flow for timeouts now:
- If mine OperationTimeout occurs, we now check to make sure the transaction *does* indeed exist on chain.
  - If it does not exist:
    - Check to make sure the signer's onchain *mined* transaction count (nonce) hasn't surpassed this transaction's nonce. If it has, this indicates the transaction was replaced by another tx, and we throw TransactionBackfilled (should I make this TransactionReplaced?).
    - If this nonce is still the blockade, we should resubmit immediately *without* bumping gas, as the transaction does not exist.
      (- If for some reason the transaction *does actually exist* we will get back a BadNonce error, which is now treated the same as an OperationTimeout.)
  - If transaction *does* exist...
    - and it has 1+ confirmations? Circle back to mine() and get the receipt!
    - and it has 0 confirmations? Bump and resubmit.

**Main change here: we now only bump gas price before resubmit if we can confirm the transaction does in fact exist on chain and hasn't received any confirmations.**

I also added more unit test code coverage for this method.

## Checklist

- [ ] Handle the transaction replaced code branch/path in unit tests for this method
- [ ] Run load tests.
- [ ] Update CHANGELOG.md.

### DIFF
--- a/packages/txservice/src/chainreader.ts
+++ b/packages/txservice/src/chainreader.ts
@@ -66,7 +66,7 @@ export class ChainReader {
    * @returns Encoded hexdata representing result of the read from the chain.
    */
   public async readTx(tx: ReadTransaction): Promise<string> {
-    return await this.getProvider(tx.chainId).readTransaction(tx);
+    return await this.getProvider(tx.chainId).readContract(tx);
   }
 
   /**
--- a/packages/txservice/src/chainreader.ts
+++ b/packages/txservice/src/chainreader.ts
@@ -18,7 +18,7 @@ import {
   getDeployedPriceOracleContract,
   getPriceOracleInterface,
 } from "./shared";
-import { ChainRpcProvider } from "./provider";
+import { RpcProviderAggregator } from "./rpcProviderAggregator";
 
 // TODO: Rename to BlockchainService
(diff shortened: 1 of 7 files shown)


rhlsthrm (approved):
Glossed over some of the nitty details but I think it looks good.

PR #664, opened 2021-12-23
[sdk] Resort to using router network if gelato fails; set default useRelayers = false

jakekidd:
## The Problem

- for loop retries 3 times... but if it fails the first time, gelatoSuccess is false, so it just throws right away lol
- default relayer fee is 0, and useRelayers default is true... we should not be using relayers if we use the default value of "0"
- we should fall back to using router network if gelato fails

## Checklist

- [ ] Run load tests.
- [ ] Update CHANGELOG.md.

### DIFF
--- a/packages/sdk/src/sdkBase.ts
+++ b/packages/sdk/src/sdkBase.ts
@@ -1000,8 +1000,8 @@ export class NxtpSdkBase {
     params: Omit<TransactionPreparedEvent, "caller">,
     fulfillSignature: string,
     decryptedCallData: string,
-    relayerFee = "0",
-    _useRelayers = false,
+    relayerFee: string,
+    useRelayers = true,
   ): Promise<{
     transactionResponse?: { transactionHash: string; chainId: number };
     transactionRequest?: providers.TransactionRequest;
@@ -1011,7 +1011,6 @@ export class NxtpSdkBase {
       undefined,
       params.txData.transactionId,
     );
-    const useRelayers = _useRelayers && BigNumber.from(relayerFee).gt(0);
     this.logger.info("Method started", requestContext, methodContext, { params, useRelayers });
     const transactionId = params.txData.transactionId;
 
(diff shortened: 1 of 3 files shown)


packages/sdk/src/sdkBase.ts
```diff
@@ -999,7 +998,7 @@ export class NxtpSdkBase {
     fulfillSignature: string,
     decryptedCallData: string,
     relayerFee = "0",
-    useRelayers = true,
+    _useRelayers = false,
```
LayneHaber, line 1001: how hot is the hot fix? i dont want to change the default behavior here without hearing from @rhlsthrm

packages/sdk/src/sdkBase.ts
```diff
@@ -1063,42 +1062,53 @@ export class NxtpSdkBase {
...
+                ? new FulfillTimeout(txData.transactionId, FULFILL_TIMEOUT, params.txData.receivingChainId, {
+                    requestContext,
+                    methodContext,
+                  })
+                : e;
+            }
           } catch (err) {
             this.logger.error("Error using Gelato Relayer", requestContext, methodContext, jsonifyError(err), {
-              attemptNum: ii + 1,
+              attemptNum: i + 1,
```
LayneHaber, line 1084: do you want to have any delay logic here?

LayneHaber (commented):
Hesitant about changing the default behavior here because I'm not sure if we want to introduce that type of change rn

packages/sdk/src/sdkBase.ts
jakekidd: Did I take that out? woops, should be re-added

packages/sdk/src/sdkBase.ts
jakekidd: more of a moderate warmfix imo

jakekidd: yes this is 100% for @rhlsthrm to pick up from here

PR #672, opened 2021-12-24
Verify contracts

LayneHaber:
## The Problem

- Can't verify contracts from `hardhat` task

## The Solution

- Add the right modules
- Update the router contract to attempt to auto-verify when a router is created

## Checklist

- [ ] Run load tests.
- [ ] Update CHANGELOG.md.


rhlsthrm:
Hardhat deploy has a built-in verifier. We only can't verify router contracts through it because they are not part of our deployments. Have you verified that the normal `hardhat etherscan-verify` through the deploy script works? I don't want to break that functionality.

LayneHaber:
I wasn't able to get it to work as a one-off verify some contract (i.e. verify the `FulfillInterpreter` or a deployed `Router`). If I'm doing it wrong it would be great if you can try that in case I'm not doing it correctly

rhlsthrm:
Yeah, there doesn't seem to be a way to one-off verify it. However, there is a flatten tool and it makes it pretty easy to verify one-off contracts.

LayneHaber:
Ill close in favor of that then

rhlsthrm:
So I actually realized we can totally get this in, because the tasks are named differently. This will actually help things.

PR #686, opened 2021-12-29
add: loss calculation

sanchaymittal:
## Checklist

- [ ] Run load tests.
- [ ] Update CHANGELOG.md.

### DIFF
--- a/packages/router/src/adapters/subgraph/subgraph.ts
+++ b/packages/router/src/adapters/subgraph/subgraph.ts
@@ -9,7 +9,7 @@ import {
   SubgraphSyncRecord,
   VariantTransactionData,
 } from "@connext/nxtp-utils";
-import { BigNumber, constants } from "ethers/lib/ethers";
+import { BigNumber, constants, utils } from "ethers";
 
 import { getContext } from "../../router";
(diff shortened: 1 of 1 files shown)


packages/router/src/adapters/subgraph/subgraph.ts
```diff
@@ -638,17 +639,30 @@ export const getCancelledFulfilledTransactions = async (routerAddress: string, c
...
   });
+
   const data = {
     records,
     miniData,
   };
   // writeFileSync("./logs.json", JSON.stringify(data));
   console.log(data);
+
+  totalLoss.forEach(async (value, key) => {
```
rhlsthrm, line 660: You need to `await` this otherwise totalLoss won't be updated. Do something like 
```
await Promise.all(totalLoss.map(async (value, key) => {...}))
```

rhlsthrm (changes requested):
Pending thoughts on comment.

packages/router/src/adapters/subgraph/subgraph.ts
LayneHaber: Also remove the console.log

PR #687, opened 2021-12-29
Check should be handled within the chain reader

LayneHaber:
## The Problem

- `metrics` helper added a check for address of price oracle

## The Solution

- check should be handled within the chain service where the oracle is queried

## Checklist

- [ ] Run load tests.
- [ ] Update CHANGELOG.md.


wanglonghong (approved):
looks good

PR #692, opened 2022-01-03
Gas checks

LayneHaber:
## The Problem

- Not properly checking min relayer fee on contracts

## The Solution

- Check if it's a contract or if it's an EOA

## Checklist

- [ ] Run load tests.
- [ ] Update CHANGELOG.md.

### DIFF
--- a/packages/router/src/lib/operations/auction.ts
+++ b/packages/router/src/lib/operations/auction.ts
@@ -245,8 +245,8 @@ export const newAuction = async (
     config.chainConfig[receivingChainId].routerContractRelayerAsset === constants.AddressZero;
 
   const [senderBalance, receiverBalance] = await Promise.all([
-    checkSender ? txService.getBalance(sendingChainId, routerAddress) : BigNumber.from(LOW_GAS),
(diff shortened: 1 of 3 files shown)


packages/router/src/lib/operations/auction.ts
```diff
@@ -234,34 +234,40 @@ export const newAuction = async (
...
-      txService.getBalance(sendingChainId, routerAddress),
-      txService.getBalance(receivingChainId, routerAddress),
-    ]);
-    logger.debug("Got balances", requestContext, methodContext, {
-      senderBalance: senderBalance.toString(),
-      receiverBalance: receiverBalance.toString(),
-    });
+  // Check for minimum gas on both chains
+  // TODO: check for proper values for different assets
+  const LOW_GAS = parseEther("0.1");
```
rhlsthrm, line 239: Shouldn't this be taken from the config? https://github.com/connext/nxtp/blob/main/packages/router/src/config.ts#L347

rhlsthrm: Never mind, this is just for logs.

rhlsthrm (approved):

PR #697, opened 2022-01-04
logs bump timeout

sanchaymittal:
- update: tracker type
- add: allSenderedTracker
- add: logs
- chg: bump timeout

## Checklist

- [ ] Run load tests.
- [ ] Update CHANGELOG.md.

### DIFF
--- a/packages/router/src/adapters/subgraph/subgraph.ts
+++ b/packages/router/src/adapters/subgraph/subgraph.ts
@@ -18,6 +18,7 @@ import {
   SingleChainTransaction,
   CrosschainTransactionStatus,
   ExpressiveAssetBalance,
+  ActiveTransactionsTracker,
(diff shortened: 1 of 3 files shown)


LayneHaber:
I'm not sure what the purpose of this PR is outside of bumping the interval timeout

LayneHaber (commented):
Need changelog updates, also need to know the impetus for this change and how adding a sender prepared tracker helps solve that. 

Additionally, more unit tests would be helpful here as well

LayneHaber (changes requested):
Need changelog updates, also need to know the impetus for this change and how adding a sender prepared tracker helps solve that.

Additionally, more unit tests would be helpful here as well

rhlsthrm:
Don't think we need this anymore, it was just for debug.

PR #698, opened 2022-01-04
Proper subg health URL

rhlsthrm:
## Checklist

- [ ] Run load tests.
- [ ] Update CHANGELOG.md.

### DIFF
--- a/packages/utils/src/fallbackSubgraph.ts
+++ b/packages/utils/src/fallbackSubgraph.ts
@@ -48,6 +48,8 @@ export class FallbackSubgraph<T extends SubgraphSdk> {
 
   private readonly subgraphs: Subgraph<T>[];
 
+  private latestSync = 0;
+
   /**
    * Returns boolean representing whether at least one available subgraph is in sync.
    */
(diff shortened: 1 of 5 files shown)


packages/utils/src/fallbackSubgraph.ts
```diff
@@ -180,6 +184,10 @@ export class FallbackSubgraph<T extends SubgraphSdk> {
    * @returns Subgraph sync records for each subgraph.
    */
   public async sync(_latestBlock?: number): Promise<SubgraphSyncRecord[]> {
+    // If the latest sync was within 5s, do not requery
+    if (Date.now() - this.latestSync < 5_000) {
```
jakekidd, line 188: make this a const at the top? And I feel like this should be higher. Like, half a quarter of our max lag value (which is in blocks, so hmmm) or something...

Would be cool if we could pass in like, 1/4 * maxLag * blockLengthMs (and get the blockLengthMs from chain reader, which determines that value on init). If no time for that, can we leave a TODO?

packages/utils/src/subgraphHealth.ts
```diff
@@ -1,6 +1,18 @@
 import axios from "axios";
 
-const SUBGRAPH_HEALTH_URL = "https://api.thegraph.com/index-node/graphql";
+// TODO get from chainData
+const GET_SUBGRAPH_HEALTH_URL = (name: string): string | undefined => {
```
jakekidd, line 4: nit: change the input param from `name` to `url` bc we actually input the url

The `name` value is something like "nxtp-bsc-v1-runtime"

packages/utils/src/subgraphHealth.ts
```diff
@@ -38,9 +51,21 @@ type SubgraphHealth = {
...
+
+const cache: { [url: string]: { record: SubgraphHealth; timestamp: number } } = {};
+export const getSubgraphHealth = async (subgraphName: string, url: string): Promise<SubgraphHealth | undefined> => {
+  const healthUrl = GET_SUBGRAPH_HEALTH_URL(url);
+  if (!healthUrl) {
+    return undefined;
+  }
+
+  // Use cache value if exist and within timeout
+  if (cache[url]?.record && Date.now() - cache[url].timestamp < 5_000) {
```
jakekidd, line 63: wait wtf? we're caching twice? this is redundant with the cache check in the fallback subg

LayneHaber: okay will only cache at top level

packages/utils/src/fallbackSubgraph.ts
```diff
@@ -85,28 +87,30 @@ export class FallbackSubgraph<T extends SubgraphSdk> {
...
-      },
-    }));
+    this.subgraphs = sdks.map(({ client, url }) => {
+      return {
+        url,
+        client,
+        record: {
+          // Typically used for logging, distinguishing between which subgraph is which, so we can monitor
+          // which ones are most in sync.
+          uri: getHostnameFromRegex(url) ? getHostnameFromRegex(url)!.split(".").slice(0, -1).join(".") : url,
```
jakekidd, line 97: lol i know this isn't your code but can we please change this check's redundant method call?
`uri: url ? getHostnameFromRegex(url).split(".").slice(0, -1).join(".") : url`

packages/utils/src/fallbackSubgraph.ts
LayneHaber: will leave as a todo and make a const

packages/utils/src/fallbackSubgraph.ts
```diff
@@ -180,13 +188,17 @@ export class FallbackSubgraph<T extends SubgraphSdk> {
    * @returns Subgraph sync records for each subgraph.
    */
   public async sync(_latestBlock?: number): Promise<SubgraphSyncRecord[]> {
+    // If the latest sync was within 5s, do not requery
```
jakekidd, line 191: lol super nit but this comment should be
`// If the latest sync was within SYNC_CACHE_TTL, do not requery`
haha

jakekidd (approved):

jakekidd (approved):

PR #699, opened 2022-01-04
Use skip polling

LayneHaber:
## The Problem

- `skipPolling` requires the UI to trigger polling when there are active transactions

## The Solution

- start polling when a transaction is `prepare`d, stop polling when there are 0 active transactions and the final tx is completed

## Checklist

- [ ] Run load tests.
- [ ] Update CHANGELOG.md.


rhlsthrm:
Closing this for now. Skip polling needs to actually not poll at all.

PR #700, opened 2022-01-05
[router] Cache auction bids; prevent overbidding by checking current outstanding liquidity

jakekidd:
## Checklist

- [ ] Run load tests.
- [x] Update CHANGELOG.md.

### DIFF
--- a/packages/router/src/adapters/cache/auction.ts
+++ b/packages/router/src/adapters/cache/auction.ts
@@ -1,12 +1,6 @@
 import { BigNumber } from "ethers";
 
-export type CachedBid = {
-  assetId: string;
-  chainId: number;
-  amountReceived: BigNumber;
-  expiry: number;
-  transactionId: string;
-};
+import { CachedBid } from "../../lib/entities/cache";
 
 // TODO: This could use some optimization.
 // - For one, we should be storing via Redis (and move this to adapters) to keep the state persistent across restarts.
--- a/packages/router/src/adapters/cache/auction.ts
+++ b/packages/router/src/adapters/cache/auction.ts
@@ -4,7 +4,6 @@ import { CachedBid } from "../../lib/entities/cache";
 
 // TODO: This could use some optimization.
 // - For one, we should be storing via Redis (and move this to adapters) to keep the state persistent across restarts.
-// - Store by transactionId in a nested Map (instead of an Array)?
 // - We could also optimize by comparing user asks to actual funding in their wallet, and only consider
 //   promised liquidity actually outstanding if the user *can* go through with the bid.
 /**
@@ -12,11 +11,11 @@ import { CachedBid } from "../../lib/entities/cache";
  * and provide a quick calculation endpoint to determine outstanding liquidity.
  */
 export class AuctionCache {
-  private readonly bids: Map<string, CachedBid[]> = new Map();
+  private readonly auctions: Map<string, Map<string, CachedBid>> = new Map();
 
   /**
    * Get the sum of all liquidity promsied across all bids for a given assetId/chainId
-   * combination.
+   * pair.
    *
    * @param chainId - chainId of the bid
    * @param assetId - assetId of the bid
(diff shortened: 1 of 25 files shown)


packages/router/src/adapters/cache/auction.ts
```diff
@@ -0,0 +1,54 @@
...
+// TODO: This could use some optimization.
+// - For one, we should be storing via Redis (and move this to adapters) to keep the state persistent across restarts.
+// - Store by transactionId in a nested Map (instead of an Array)?
+// - We could also optimize by comparing user asks to actual funding in their wallet, and only consider
+//   promised liquidity actually outstanding if the user *can* go through with the bid.
+/**
+ * Bid caching helper class to ensure we only store one bid per assetId/chainId combination, clean up expired bids,
+ * and provide a quick calculation endpoint to determine outstanding liquidity.
+ */
+export class AuctionCache {
```
LayneHaber, line 14: Wondering the best way to handle the case where we win the bid, and the bid expires before we are able to prepare the receiver. Perhaps there should be a way to increase the expiry or set a `confirmed` flag?

LayneHaber (commented):
Looks good, pending tests.

Still need to figure out how to extend the lifetime of a bid when processing `SenderPrepared` transactions

packages/router/src/adapters/cache/auction.ts
jakekidd: oh right, like we want to keep the liquidity reserved, right?

packages/router/src/bindings/contractReader/index.ts
```diff
@@ -214,6 +214,20 @@ export const handleSingle = async (
     const preparePayload: PreparePayload = _transaction.payload;
     try {
       logger.info("Preparing receiver", requestContext, methodContext);
+      const didConfirm = cache.auctions.confirmBid(
```
rhlsthrm, line 217: This seems a bit unnecessary? Our signature on a bid is essentially our commitment, so I'm not sure what we gain by confirming that it's in the cache.

rhlsthrm (changes requested):
Great job on this, annoying that we replicated some work, but no worries. We can probably just go with this implementation at first. Pending some answers/clarifications. Also, I think we should add the config on the 150% thing.

packages/router/src/bindings/contractReader/index.ts
jakekidd: You're misreading the purpose of this. It's not for our sake here, it's for the sake of keeping that outstanding liquidity "held" as outstanding until we've confirmed that our prepare has deducted from our overall balance.

packages/router/src/adapters/cache/auction.ts
```diff
@@ -0,0 +1,93 @@
...
+   * @param bid - bid to add
+   */
+  public addBid(bid: Omit<CachedBid, "confirmed">) {
+    const key = this.getCachedBidCompositeKey(bid.chainId, bid.assetId);
+    const bids = this.auctions.get(key) ?? new Map<string, CachedBid>();
+    bids.set(bid.transactionId, {
+      ...bid,
+      confirmed: false,
+    });
+    this.auctions.set(key, bids);
```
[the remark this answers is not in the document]
jakekidd: They do expire? Each time we calculate outstanding liquidity we remove expired bids...?

packages/router/src/bindings/contractReader/index.ts
jakekidd: https://github.com/connext/nxtp/pull/700/files/7f1b5df049f918c48c9a8f326788ae93b955dc90#r778491800

jakekidd: So:
`bid sent ---> bid accepted ---> user prepares ---> router sees user prepared -***-> router prepares and liquidity is deducted from router's pool`
***at some point here, asynchronously, the bid could "expire" but we want the liquidity to remain outstanding until our prepare is complete

jakekidd: I'll add a clarifying comment

packages/router/src/adapters/cache/auction.ts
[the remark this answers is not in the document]
rhlsthrm: Oops, I guess I overlooked this.

rhlsthrm (approved):

rhlsthrm:
This is merged, but I just realized we should really have a way to turn this functionality off (i.e. accept bids always) just in case something goes wrong. We need to make sure things are properly feature-flagged through config.

PR #702, opened 2022-01-05
Polling loop

rhlsthrm:
## Checklist

- [ ] Run load tests.
- [ ] Update CHANGELOG.md.

### DIFF
--- a/packages/sdk/src/subgraph/subgraph.ts
+++ b/packages/sdk/src/subgraph/subgraph.ts
@@ -148,6 +148,10 @@ export class Subgraph {
     }
   }
 
+  get isPolling(): boolean {
+    return !!this.pollingLoop;
+  }
+
   public stopPolling(): void {
     if (this.pollingLoop != null) {
       clearInterval(this.pollingLoop);
@@ -739,6 +743,9 @@ export class Subgraph {
     filter: (data: SubgraphEventPayloads[T]) => boolean = (_data: SubgraphEventPayloads[T]) => true,
     timeout?: number,
   ): void {
+    if (!this.pollingLoop) {
+      throw new PollingNotActive();
+    }
     const args = [timeout, callback].filter((x) => !!x);
     this.evts[event].pipe(filter).attach(...(args as [number, any]));
   }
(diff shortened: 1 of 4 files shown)


LayneHaber (changes requested):
Needs changelog updates.

Also, how was this tested? Only using the updated integration test I added? Can we add some unit tests for this?

packages/sdk/test/unit/sdkBase.spec.ts
```diff
@@ -1118,6 +1113,39 @@ describe("NxtpSdkBase", () => {
       expect(res.transactionResponse.chainId).to.be.eq(sendingChainId);
     });
 
+    it.skip("finish transfer starts polling if not polling already, and stops after", async () => {
```
LayneHaber, line 1116: why is this skipped if it was just added?

packages/sdk/src/subgraph/subgraph.ts
```diff
@@ -161,6 +165,12 @@ export class Subgraph {
         const { methodContext, requestContext } = createLoggingContext("pollingLoop");
         try {
           await this.getActiveTransactions();
+
+          const activeTxsSize = this.activeTxs.size;
+          if (activeTxsSize == 0) {
+            clearInterval(this.pollingLoop!);
+            this.pollingLoop = undefined;
+          }
```
LayneHaber, line 173: Why do we only stop polling if there are no active transactions, wouldn't that result in some weird behavior?

LayneHaber (changes requested):
Can't skip a unit test we just added :(

packages/sdk/test/unit/sdkBase.spec.ts
LayneHaber: skipped because needs to be deleted. modified expectations to unskip, but may still need to be deleted

LayneHaber (changes requested):
After a conversation with @sanchaymittal found that this feature needs a bit more TLC before releasing:

- edgecase: user prepares, polling starts, subgraph loop race or not up to date so sees no active transaction, stops polling when it shouldnt
- improvement: start polling based on chainId and block number
- improvement: make polling functional instead of object oriented (similar to the router)

we will likely have immediate UX regressions if we merge this, which is not worth the reduction in subgraph queries. leaving this out of next release and taking the time to implement correctly and completely :)

thanks @sanchaymittal for the thoughtful consideration!

rhlsthrm:
OK but unfortunately this had another fix that we do need which is where the relayer send does not work without polling because the promise is set up in the wrong place. I will pull this into another branch.

sanchaymittal:
Got alternative implementation live on main

PR #706, opened 2022-01-05
[txservice] Rewind nonce if transaction fails and was never mined

jakekidd:
## The Problem

- If a tx at nonce X failed but never got mined (possibly indicating it was not sent correctly), we throw... but never backfill at that nonce! Thus causing a blockade.

## The Solution

- When a tx errors out, we check to see if it got mined properly. If not, then we reset the local nonce to the value of this tx's nonce. Then, on the next transaction, we will backfill.
  - Worst case scenario, if the tx *does* end up getting mined even after resetting local nonce, on the next sent transaction we will just end up favoring the signer's current transactionCount over the local nonce (since we prefer whichever number is higher!).
  - Only problem with this diff is that we're resetting the local nonce but not actually guaranteeing that we send a backfill... meaning that we have to *hope* that another transaction comes in soon. Regardless, it is definitely an improvement and better solution than we had previously (which was no solution for this edge case at all...)
  

## Checklist

- [ ] Run load tests.
- [x] Update CHANGELOG.md.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Next Release
 
+- [txservice] Rewind nonce if transaction fails and was never mined
 - [sdk] Return subgraph query in generic query method
 - [router] Track bids in memory to only bid 150% of available liquidity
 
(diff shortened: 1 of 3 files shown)


LayneHaber (changes requested):
Needs changelog updates. Also can we add some unit tests for this as well?

LayneHaber (commented):
As soon as you update the changelog you can merge this, i would even allow a changelog update post-merge since it lgtm... but changelog

PR #711, opened 2022-01-06
add: sanity check for router contract

sanchaymittal:
## Checklist

- [ ] Run load tests.
- [ ] Update CHANGELOG.md.

### DIFF
--- a/packages/router/src/lib/operations/status.ts
+++ b/packages/router/src/lib/operations/status.ts
@@ -6,27 +6,13 @@ import { handlingTracker, activeTransactionsTracker } from "../../bindings/contr
 import { version } from "../../../package.json";
 
 export const getStatus = (_requestContext: RequestContext<string>): StatusResponse => {
-  const { config, isRouterContract, signerAddress, routerAddress } = getContext();
+  const { isRouterContract, signerAddress, routerAddress, chainAssetSwapPoolMap } = getContext();
 
   const routerVersion = version;
   const trackerLength = handlingTracker.size;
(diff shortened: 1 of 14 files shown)


packages/router/src/router.ts
```diff
@@ -90,6 +91,25 @@ export const makeRouter = async () => {
     context.contractReader = subgraphContractReader();
     context.contractWriter = contractWriter();
 
+    context.chainAssetSwapPoolMap = new Map();
+    // sanity check if router contract
+    context.config.swapPools.forEach((pool) => {
```
rhlsthrm, line 96: This is an async function. You need to do:
```ts
await Promise.all(context.config.swapPools.map(async (pool) => {...}))
```

rhlsthrm (changes requested):

rhlsthrm (approved):

packages/router/src/lib/operations/status.ts
```diff
@@ -35,7 +21,7 @@ export const getStatus = (_requestContext: RequestContext<string>): StatusRespon
     signerAddress,
     trackerLength,
     activeTransactionsLength,
-    swapPools: chainAssetMap,
+    swapPools: chainAssetSwapPoolMap,
```
LayneHaber, line 24: if this response goes over the wire, this will not be properly formatted because it's a map. needs to be proper json

LayneHaber (changes requested):

LayneHaber (approved):

PR #715, opened 2022-01-06
[router] Run estimate gas check before calling relayers for contract methods

jakekidd:
## The Problem

- If a transaction would revert onchain, we were previously calling relayers/router network to execute an onchain tx and then waiting up to a *combined* 6(?) mins for both before finally executing with txservice only to find out that the tx will revert onchain.

## The Solution

- If we *are* using relayers, check immediately if the tx will revert with an estimateGas call. It will throw TransactionReverted if that's the case (and txservice error parser will pluck out the contract error code for us), giving us an immediate response. If it's successful, it'll pass through and we move on to relayers.
- This is an urgent fix that covers a possible(?) attack vector (relayers can lie and say they executed when it reality it would revert I think?)
- This isn't the most ideal in the event that we *do* end up using txservice, because that means we are doing the estimate gas call twice.

## Checklist

- [ ] Run load tests.
- [x] Update CHANGELOG.md.

### DIFF
--- a/packages/router/test/adapters/contract/contract.spec.ts
+++ b/packages/router/test/adapters/contract/contract.spec.ts
@@ -37,6 +37,15 @@ import { ERC20Interface } from "@connext/nxtp-contracts/typechain/ERC20";
 
 const requestContext = createRequestContext("TEST");
 const encodedDataMock = "0xabcde";
+const chainIdMock = txDataMock.sendingChainId;
+const onchainTxMock = {
+  to: routerContractAddressMock,
+  data: encodedDataMock,
+  value: constants.Zero,
+  chainId: chainIdMock,
+  from: mkAddress("0xa"),
+};
+const routerRelayerFeeAssetMock = mkAddress("0x1a2b3c");
 
 let interfaceMock: SinonStubbedInstance<Interface>;
 let routerInterfaceMock: SinonStubbedInstance<Interface>;
@@ -396,7 +405,7 @@ describe("Contract Adapter", () => {
     });
   });
 
-  describe("getRouterBalance", async () => {
+  describe("#getRouterBalance", async () => {
     it("should work", async () => {
       txServiceMock.readTx.resolves("10");
       const ret = await ContractFns.getRouterBalance(1337, routerAddrMock, mkAddress());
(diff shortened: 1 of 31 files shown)


LayneHaber (changes requested):
Tests, changelog

packages/router/test/adapters/contract/contract.spec.ts
```diff
@@ -31,12 +31,21 @@ import { createStubInstance, SinonStubbedInstance, stub, SinonStub, restore, res
...
 const requestContext = createRequestContext("TEST");
 const encodedDataMock = "0xabcde";
+const chainIdMock = txDataMock.sendingChainId;
+const onchainTxMock = {
+  to: routerContractAddressMock,
+  data: encodedDataMock,
+  value: constants.Zero,
+  chainId: chainIdMock,
+  from: routerAddress,
+};
```
LayneHaber, line 47: This should line up with the `txReceiptMock` as well

packages/router/test/adapters/contract/contract.spec.ts
```diff
@@ -396,29 +405,27 @@ describe("Contract Adapter", () => {
...
+    it("should work (without relayers)", async () => {
       const res = await ContractFns.prepareRouterContract(
-        chainId,
+        chainIdMock,
         prepareParamsMock,
         routerContractAddressMock,
         sigMock,
-        routerRelayerFeeAsset,
+        routerRelayerFeeAssetMock,
         "1",
```
LayneHaber, line 429: why leave out the router relayer fee from the constants tho

LayneHaber (approved):
Some nit-picky comments but lgtm!

packages/router/test/adapters/contract/contract.spec.ts
jakekidd: lol not sure, didn't really think about it

packages/router/test/adapters/contract/contract.spec.ts
jakekidd: Yeah a good point. We should def condense some of the mocking we've done across all unit tests in router at some point

PR #716, opened 2022-01-07
Metrics update

LayneHaber:
## The Problem

Metrics collection had a couple of problems:
- Gas was not being properly tracked on normal transactions
- Costs of the router relays was not being properly tracked on router-contract transactions

## The Solution

- Make sure to track gas consumed not gas units
- Move all `gasConsumed` updates to the contract writer
- Add a `relayerFeesPaid` counter to track fees paid to transaction relayers

## Checklist

- [ ] Run load tests.
- [x] Update CHANGELOG.md.

### DIFF
--- a/packages/router/src/adapters/contract/contract.ts
+++ b/packages/router/src/adapters/contract/contract.ts
@@ -29,26 +29,11 @@ import {
   incrementGasConsumed,
 } from "../../lib/helpers";
 import { TransactionReasons } from "../../lib/entities";
+import { incrementRelayerFeesPaid } from "../../lib/helpers/metrics";
 
-type RouterContractPrepareEvent = {
-  invariantData: InvariantTransactionData;
-  routerRelayerFeeAsset: string;
-  routerRelayerFee: BigNumber;
-  caller: string;
-  transactionHash: string;
-};
-type RouterContractFulfillEvent = {
-  txData: TransactionData;
-  routerRelayerFeeAsset: string;
-  routerRelayerFee: BigNumber;
-  caller: string;
-  transactionHash: string;
-};
-type RouterContractCancelEvent = RouterContractFulfillEvent;
-
-export const prepareEvt = new Evt<{ event: RouterContractPrepareEvent; args: PrepareParams; chainId: number }>(); // TODO: fix types
-export const fulfillEvt = new Evt<{ event: RouterContractFulfillEvent; args: FulfillParams; chainId: number }>();
-export const cancelEvt = new Evt<{ event: RouterContractCancelEvent; args: CancelParams; chainId: number }>();
+export const prepareEvt = new Evt<{ event: any; args: PrepareParams; chainId: number }>(); // TODO: fix types
+export const fulfillEvt = new Evt<{ event: any; args: FulfillParams; chainId: number }>();
+export const cancelEvt = new Evt<{ event: any; args: CancelParams; chainId: number }>();
 export const removeLiquidityEvt = new Evt<{
   event: any;
   args: RemoveLiquidityParams;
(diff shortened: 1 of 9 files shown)


packages/router/src/lib/helpers/metrics.ts
```diff
@@ -326,6 +351,43 @@ export const incrementGasConsumed = async (
...
+    assetId,
+    relayerFee,
+    reason,
+  });
+
+  const usd = await convertToUsd(assetId, chainId, relayerFee, requestContext);
+
+  relayerFeesPaid.inc({ reason, chainId, assetId }, usd);
+
+  throw new Error(`Implement me`);
```
jakekidd, line 388: ??

jakekidd: lol looks like u implemented

LayneHaber: omg lol

packages/router/src/adapters/contract/contract.ts
```diff
@@ -649,6 +762,15 @@ export const cancelRouterContract = async (
     },
     requestContext,
   );
+
+  incrementGasConsumed(
+    chainId,
+    receipt,
+    txData.sendingChainId === chainId ? TransactionReasons.CancelSender : TransactionReasons.CancelReceiver,
+    requestContext,
+  );
```
jakekidd, line 771: Okay soooo this kinda sucks but, even with estimateGas calls beforehand, it's still possible for the sent transaction to revert onchain - which means we spent the gas (likely edge case I can think of is on a redundant call after a router reboot).

However, if the transaction reverts onchain, txService.sendTx would throw, meaning this block would not be reached.

So we need to have a try/catch/finally here, where in the finally block we incrementGasConsumed if either we got the receipt back (was successful) or the error thrown was from the transaction reverting onchain

Which is now making me realize that txservice should *probably* distinguish between an onchain revert (meaning gas  money was spent) and an offchain/gas estimate revert...

jakekidd: For the last thing, we could just have TransactionReverted have an optional receipt member that gets set to the mined receipt in the event it reverts onchain

LayneHaber: Yes, but is there a way to check from the txservice the total gas consumed for a transaction (including bumping, retries, reverts etc)? I am aware of this edgecase, but not sure if there's currently a way to track it.

Polling the signer for native funds downed isn't really a great option, since there's no way to see which transactions are using the most gas

LayneHaber: Also, this problem exists in gelato as well. I'm not sure how the reverted transactions are handled there cc @rhlsthrm

packages/router/src/adapters/contract/contract.ts
```diff
@@ -628,8 +731,18 @@ export const cancelRouterContract = async (
       const { event } = await cancelEvt
         .pipe(({ args }) => args.txData.transactionId === txData.transactionId)
         .waitFor(300_000);
+
+      // increment router fees
+      incrementRelayerFeesPaid(
```
jakekidd, line 736: this needs to come *after* we call `getTransactionReceipt`! We have no idea if the relayer fees are paid at all until we get a receipt back indicating it executed.

packages/router/src/adapters/contract/contract.ts
```diff
@@ -594,6 +687,16 @@ export const cancelRouterContract = async (
...
```
jakekidd, line 692: see below comment. This also applies to prepare/fulfill etc

packages/router/src/adapters/contract/contract.ts
jakekidd: The problem exists for the gelato / relayer logic only with the above implementation. Just wait for the receipt, if we get it back, increment fees paid (bc then we *know* the fees were paid)

The only reason the edge case will persist here is bc we are explictly throwing if the tx was reverted on chain, but need to catch that and do an increment anyway in that event.

And yes, if the transaction is reverted onchain, the OnchainTransaction object internal to txservice grabs the (revert tx) receipt anyway, so we might as well use it.

LayneHaber: You're right -- totally thought that we should be paying them even in reverts (which we should) but that's not how its implemented. Let's just return it as part of the error then -- can work on that tomorrow (if @rhlsthrm doesn't get it done)

packages/router/src/adapters/contract/contract.ts
LayneHaber: What do you mean? I'm not sure of the edgecase you're trying to address here (I think gelato just doesn't return anything if it fails)

packages/router/src/adapters/contract/contract.ts
rhlsthrm: We don't pay them in reverts, I think that's a risk that relayers take on themselves. They need to basically stop from getting frontrun.

packages/router/src/adapters/contract/contract.ts
rhlsthrm: I think @jakekidd's logic is correct here. Unless we know the tx succeeded, we won't know that the relayer fee was taken from the contract (which only happens in a successful contract call). However, it should be at this point if this EVT didn't error.

rhlsthrm (approved):

rhlsthrm (approved):

packages/router/src/adapters/contract/contract.ts
jakekidd: I mean, the stalling here comes from the `getTransactionReceipt` call. Our gelato relayer call is just an API POST request and we have no idea whether it's successful until we either get the receipt back from `getTransactionReceipt` or we timeout doing so.

PR #722, opened 2022-01-07
[router] Contract adapters refactor

jakekidd:
## The Problem

Lots of redundant code in contract adapters for router.

## The Solution

Condense into using helper methods. Basically I moved this 1-2-3 step logic and all the relayer handling into one separate helper method `sendRouterContractTx`. Instead, every router contract method just produces the unique arguments and passes it down to the helper. This will help us improve the logic here much quicker in the future (if we need any additional preflight checks, any more metrics recorded, anything else logged, or any fixes done, it will all be in one place).

Still needs unit tests written.

## Checklist

- [ ] Add unit tests.
- [ ] Update CHANGELOG.md.

### DIFF
--- a/packages/router/src/adapters/contract/contract.ts
+++ b/packages/router/src/adapters/contract/contract.ts
@@ -12,6 +12,7 @@ import {
   RemoveLiquidityParams,
   InvariantTransactionData,
   TransactionData,
+  MethodContext,
 } from "@connext/nxtp-utils";
 import { BigNumber, constants, Contract, providers, utils } from "ethers";
 import { Evt } from "evt";
@@ -28,7 +29,7 @@ import {
   getRouterContractInterface,
   incrementGasConsumed,
 } from "../../lib/helpers";
-import { TransactionReasons } from "../../lib/entities";
+import { TransactionReason, TransactionReasons } from "../../lib/entities";
 import { incrementRelayerFeesPaid } from "../../lib/helpers/metrics";
 
 export const prepareEvt = new Evt<{ event: any; args: PrepareParams; chainId: number }>(); // TODO: fix types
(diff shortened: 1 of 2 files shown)


packages/router/src/adapters/contract/contract.ts
```diff
@@ -44,6 +45,197 @@ export const removeLiquidityEvt = new Evt<{
...
+    amount: string;
+  },
+  reason: TransactionReason,
+  requestContext: RequestContext,
+) => {
+  const { txService } = getContext();
+  // Listen for event on contract.
+  const { event } = transactionId
+    ? await evt.pipe(({ args }) => args.txData.transactionId === transactionId).waitFor(300_000)
+    : await evt.waitFor(300_000);
```
LayneHaber, line 77: Waiting for just any event wouldn't work here, you still need something to match as an identifier -- none of our evts are that specific

packages/router/src/adapters/contract/contract.ts
```diff
@@ -44,6 +45,197 @@ export const removeLiquidityEvt = new Evt<{
...
+  // Retrieve the receipt.
+  const receipt = await txService.getTransactionReceipt(chainId, event.transactionHash);
```
LayneHaber, line 79: Is there a case where this would return `undefined`? The [function](https://github.com/connext/nxtp/blob/153dbe146ea52ec5f583407630d9f9e79ab19ea0/packages/txservice/src/chainreader.ts#L158-L160) doesn't have that option in the types, but it is in the [ethers docs](https://docs.ethers.io/v5/api/providers/provider/#Provider-getTransactionReceipt).

We should probably update the types in the `chainreader` and handle the `undefined` case here.

packages/router/src/adapters/contract/contract.ts
```diff
@@ -44,6 +45,197 @@ export const removeLiquidityEvt = new Evt<{
...
+    reason: TransactionReason;
+  },
+  methodContext: MethodContext,
+  requestContext: RequestContext,
+): Promise<providers.TransactionReceipt> => {
+  const { txService, wallet, logger, messaging } = getContext();
+  const { chainId, routerContractAddress, encodedData, params, signature, relayerFee, reason } = args;
+
+  // Determine which event and payload type to use based on the transaction type.
+  let evt: Evt<any>;
```
LayneHaber, line 124: Why not just make this a generic function that passes in the types rather than typing it as `any`?

packages/router/src/adapters/contract/contract.ts
LayneHaber: Ahh I see, you are handling the `removeLiquidity` in the `:` block. Since that is the only reason where a call wouldn't have a `transactionId`, can we handle the `undefined` case more explicitly?

packages/router/src/adapters/contract/contract.ts
```diff
@@ -273,121 +458,24 @@ export const prepareRouterContract = async (
...
+      encodedData,
+      params: prepareParams,
+      signature,
+      relayerFee: useRelayer
+        ? {
+            asset: routerRelayerFeeAsset,
+            amount: routerRelayerFee,
+          }
+        : undefined,
+      reason: TransactionReasons.PrepareReceiver,
```
LayneHaber, line 474: Could still be a `metaTx` relay, as seen [here](https://github.com/connext/nxtp/blob/cdaccfdc3018803e2df6bba26e6b77f969de6582/packages/router/src/lib/operations/metaTx.ts#L156)

packages/router/src/adapters/contract/contract.ts
```diff
@@ -483,122 +571,24 @@ export const fulfillRouterContract = async (
...
-  logger.info("Router contract fulfill: sending using txservice", requestContext, methodContext, { fulfillParams });
-  const receipt = await txService.sendTx(
-    onchainTx,
+      relayerFee: useRelayer
+        ? {
+            asset: routerRelayerFeeAsset,
+            amount: routerRelayerFee,
+          }
+        : undefined,
+      reason: TransactionReasons.FulfillSender,
```
LayneHaber, line 587: Could still be a `metaTx` relay, as seen [here](https://github.com/connext/nxtp/blob/cdaccfdc3018803e2df6bba26e6b77f969de6582/packages/router/src/lib/operations/metaTx.ts#L205)

packages/router/src/adapters/contract/contract.ts
```diff
@@ -676,127 +666,24 @@ export const cancelRouterContract = async (
...
-    chainId,
-    receipt,
-    txData.sendingChainId === chainId ? TransactionReasons.CancelSender : TransactionReasons.CancelReceiver,
+      relayerFee: useRelayer
+        ? {
+            asset: routerRelayerFeeAsset,
+            amount: routerRelayerFee,
+          }
+        : undefined,
+      reason: chainId === txData.sendingChainId ? TransactionReasons.CancelSender : TransactionReasons.CancelReceiver,
```
LayneHaber, line 682: Could still be a `metaTx` relay, as seen [here](https://github.com/connext/nxtp/blob/cdaccfdc3018803e2df6bba26e6b77f969de6582/packages/router/src/lib/operations/metaTx.ts#L251)

LayneHaber (changes requested):
What else do you want to add to make this a non-draft PR? Just the tests and changelog stuff?

packages/router/src/adapters/contract/contract.ts
jakekidd: bc im dumb

sanchaymittal:

PR #731, opened 2022-01-11
[DNM] [router][sdk] Subgraph endpoint

jakekidd:
## The Problem

- Need to use our subgraph endpoint for health in the fallback subgraph
- No need to have hardcoded subgraph URLs in the codebase if we are serving the URLs from the endpoint
- No need to accept configured subgraph URLs when Connext will be keeping the subgraph endpoint updated with latest URLs.

## The Solution

- Implemented a call to subgraph health endpoint to replace previous health methods
- FallbackSubgraph now uses a Map of [url:string]: subgraph, enabling us to dynamically set what subgraphs are available once we've called sync().
- Removed ChainReader usage in sdk/Subgraph class.
- Remove configured subgraph URL(s) requirement. We no longer require URLs to be passed to FallbackSubgraph's constructor, since it can get them from the health endpoint.

## Checklist - Must Complete!

- [x] Make sure sync() has been called (or is in progress) when the first request is called in SDK. May need to implement a serialized queue (as a sort of thread lock) to make sure sync can be awaited in sep. threads.
- [x] Health endpoint needs some changes: add analytics endpoint, need to make sure the number options are correct.
- [x] Unit tests: need to add / flesh out unit tests for FallbackSubgraph.
- [x] Update CHANGELOG.md.


jakekidd:
Going to reimplement the config support in this PR; we need to be able to pass in configured subgraphs apparently for integration testing.

(Also, may need to reintroduce the hardcoded subgraphs for analytics, since not sure if the endpoint for analytics is going to be ready yet.)

packages/sdk/src/subgraph/subgraph.ts
```diff
@@ -168,25 +167,19 @@ export class Subgraph {
     }
   }
 
-  getSyncStatus(chainId: number): SubgraphSyncRecord {
+  getSyncStatus(chainId: number): SubgraphSyncRecord | undefined {
     const { requestContext, methodContext } = createLoggingContext(this.getSyncStatus.name);
     const records = this.sdks[chainId].records;
     this.logger.debug(
-      `Getting sync status for chain ${chainId} with ${records.length} records`,
+      `Subgraph sync status for chain ${chainId} (${records.length} records)`,
```
LayneHaber, line 174: These types of logs make it difficult to search for something in LogDNA

packages/utils/src/fallbackSubgraph.ts
```diff
@@ -36,85 +48,102 @@ type Subgraph<T extends SubgraphSdk> = {
...
   constructor(
     private readonly chainId: number,
-    // We use the URIs in sync records for reference in logging.
-    sdks: { client: T; url: string }[],
+    private readonly generateClient: (url: string) => T,
     private readonly maxLag: number,
+    private readonly domain: SubgraphDomain = SubgraphDomain.COMMON,
+    urls: string[] = [],
     private readonly stallTimeout = 10_000,
   ) {
```
LayneHaber, line 141: The interface of this utility class changes a lot, and that's not necessarily the main improvement outlined in the PR body. @rhlsthrm what would the semver procedure for this when none of the other interfaces changed?

packages/utils/src/fallbackSubgraph.ts
```diff
@@ -175,116 +225,158 @@ export class FallbackSubgraph<T extends SubgraphSdk> {
...
-          } catch (e) {
-            errors.push(e);
-          }
-        }
+      // Target this chain's endpoint.
+      const endpointUrl = endpoint.concat(`?chainId=${this.chainId}`);
+      let response: AxiosResponse<SubgraphHealth[]> | undefined = undefined;
+      try {
+        response = await axios.get(endpointUrl);
+      } catch (e) {}
```
LayneHaber, line 265: Swallowing the error is likely not a good idea (it should at least have the `message` logged at a debug level).

Also, would we explicitly not want to retry this request as well?

packages/utils/src/fallbackSubgraph.ts
```diff
@@ -313,10 +405,97 @@ export class FallbackSubgraph<T extends SubgraphSdk> {
...
+        cps: 0,
+        reliability: 0,
+        avgExecTime: 0,
+      },
+    };
+  }
 }
+
+// TODO: Remove, replace with API call / hosted data.
+export const getDeployedAnalyticsSubgraphUrls = (chainId: number, chainData?: Map<string, ChainData>): string[] => {
```
LayneHaber, line 450: Any particular reason for the function name change? Otherwise breaks interfaces unnecessarily

LayneHaber (changes requested):
Looks like this needs more changes for test coverage. Couple of questions:

1. Is the hosted URL ready?
2. Should we include analytics URL in the chain data and pull from there?
3. It seems like there's a lot of additional changes outside of the url switch, some of which break interfaces. Can we separate those changes out?

In the future, let's leave PRs as drafts until they're ready for review

packages/utils/src/fallbackSubgraph.ts
jakekidd: We could retry this.

Also, I'm not going to log it here, but I'll attach the error to the subgraph sync record. The real reason we ignore the error is because it basically just indicates the health endpoint is down / currently not supported - meaning we should resort to using the getBlockNumber method.

packages/utils/src/fallbackSubgraph.ts
jakekidd: why? who's calling this? They shouldn't be - and this should be removed. This hardcoded stuff is intended to be deprecated when there's a health endpoint for analytics as well.

jakekidd:
> Looks like this needs more changes for test coverage. Couple of questions:
> 
> 1. Is the hosted URL ready?
> 2. Should we include analytics URL in the chain data and pull from there?
> 3. It seems like there's a lot of additional changes outside of the url switch, some of which break interfaces. Can we separate those changes out?
> 
> In the future, let's leave PRs as drafts until they're ready for review

1. Yep. See for yourself: https://subgraph-ts-worker.connext.workers.dev/subgraph_health/?chainId=137
2. Yeah I would honestly rather implement this as a second step / separate PR. The analytics subgraphs can stay hardcoded for now, but personally I think we should have a separate health endpoint that basically 
3. I need more clarification on who's interfacing with FallbackSubgraph and the hardcoded analytics urls method. Internally, all interfacing has been adjusted. Otherwise, it wouldn't compile.

Also unfortunately, this PR has kinda always been ready for review. I implemented it one way, then received the feedback that it should be done another way. Would be great if we could go back from r4r -> draft, but can't afaik.

Otherwise, a lot of the CI failures have been integration test, which I haven't been able to validate beforehand locally (due to my computer not being set up for that yet), so I take responsibility for labeling this as r4r in order to use CI lol. Can we get CI to work with drafts?

packages/utils/src/fallbackSubgraph.ts
rhlsthrm: I think it's fine to leave the same names for the functions since these are external. Let's not make assumptions about who is calling what. This function should not necessarily be deprecated, it should use chainData instead if it can.

packages/utils/src/fallbackSubgraph.ts
LayneHaber: Consensus: not really guaranteed that we are the only consumers of these functions, will release at `0.2.0`

packages/utils/test/fallbackSubgraph.spec.ts
```diff
@@ -0,0 +1,90 @@
...
+    beforeEach(() => {
+      stub(fallbackSubgraph as any, "getOrderedSubgraphs").returns(mockOrderedSubgraphs);
+      mockFailingSubgraphSdkMethod = stub().rejects(failingSubgraphSdkError);
+    });
+    it("happy: should return first successful result, and update metrics", async () => {});
+    it("should retry if initial method call throws", async () => {});
+    it("should resort to next subgraph in orderedSubgraphs if first one fails", async () => {});
+    it("should throw if syncRequired and no subgraphs are synced", async () => {});
+    it("should throw if no subgraphs available", async () => {});
+    it("should throw if none of the subgraph clients succeeded", async () => {});
```
LayneHaber, line 76: if we want to merge without the unit tests, then they should be skipped so we dont get the false impression they are actually implemented (as they will appear "passing").

but i dont think we should merge without the unit tests, especially because there's a lot of error handling logic that wouldn't necessarily be hit in the load tests

LayneHaber (changes requested):
Needs to have the unit tests implemented before merge.

When releasing, we should release at 0.2.0 to reflect the interface changes. Since all of these classes/functions are exported, we can't make any assumptions about who is or isn't consuming them

packages/utils/test/fallbackSubgraph.spec.ts
jakekidd: yep left this as a placeholder then forgot about it lol, on it now

jakekidd:
> Needs to have the unit tests implemented before merge.
> 
> When releasing, we should release at 0.2.0 to reflect the interface changes. Since all of these classes/functions are exported, we can't make any assumptions about who is or isn't consuming them

Oh to clarify, my question about who is consuming them was really rhetorical... imo, the interfaces that changed here should *not* have been consumed by anyone outside the library.

In an ideal world they wouldn't be exported to outside the library at all... but that's a sep. convo

packages/utils/src/fallbackSubgraph.ts
```diff
@@ -36,85 +48,127 @@ type Subgraph<T extends SubgraphSdk> = {
...
+type SubgraphHealthError = {
+  message: string;
+  block: number;
+  handler: any;
+};
+
+type SubgraphHealth = {
+  data: {
+    // TODO: Deprecate this unnecessary wrapping object.
+    indexingStatusForCurrentVersion: {
```
LayneHaber, line 61: If we are releasing a breaking change, and this isn't live yet, why not just deprecate now?

LayneHaber: Oh is this nested in a `data.data.indexingStatusForCurrentVersion`? Let's just fix this now

LayneHaber (commented):
Seems like the tech debt is easy enough to just address now

packages/sdk/src/subgraph/subgraph.ts
```diff
@@ -168,25 +167,16 @@ export class Subgraph {
...
-        synced: false,
-        syncedBlock: 0,
-        latestBlock: 0,
-      }
-    );
+    this.logger.debug("Retrieved subgraph sync status.", requestContext, methodContext, {
+      chainId,
+      records,
+      // For easy searching of "no records found" in logs.
+      recordsFounds: records.length > 0,
```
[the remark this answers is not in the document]
jakekidd: I founds the precious recordsis

LayneHaber (approved):
🎉

PR #742, opened 2022-01-14
[sdk] Sdk sendTransaction helper

jakekidd:
## The Problem

SDK was operating on assumption in a lot of places that the injected signer has already been reconnected to the correct providers.

## The Solution

A `sendTransaction` helper in `NxtpSdkBase` connects the injected signer to the providers passed down in the SDK config as soon we intend to execute a transaction. Otherwise, the signer could be connected to the wrong chain ID.

## Checklist

- [ ] Update CHANGELOG.md.


sanchaymittal:

PR #744, opened 2022-01-14
Add v1PriceOracle

wanglonghong:
## Checklist

- [ ] Run load tests.
- [ ] Update CHANGELOG.md.


rhlsthrm (approved):

rhlsthrm:
Can you add the v1 price oracle addresses to the deploy script so that when we deploy we can automatically set the price oracle?

packages/contracts/contracts/ConnextPriceOracle.sol
```diff
@@ -136,6 +142,12 @@ contract ConnextPriceOracle is PriceOracle {
         assetPrices[_token] = _price;
     }
 
+    function setV1PriceOracle(address _v1PriceOracle) external {
```
LayneHaber, line 145: If we're changing things, can we make the `require` a modifier? Is there a reason not to?

packages/contracts/test/connextPriceOracle.spec.ts
```diff
@@ -243,9 +265,22 @@ describe("ConnextPriceOracle.sol", () => {
         .setDexPriceInfo(tokenA.address, stableToken.address, lpTokenAddress, true);
 
       const dexPrice = await connextPriceOracle.getPriceFromDex(tokenA.address);
-      const tokenPrice = await connextPriceOracle.getPriceFromDex(tokenA.address);
+      const tokenPrice = await connextPriceOracle.getTokenPrice(tokenA.address);
```
LayneHaber, line 268: One piece of feedback that we have gotten is:

> The `ChainlinkAdapter` calls out to a Chainlink oracle receiving the `latestRoundData()`. If there is a problem with Chainlink starting a new round and finding consensus on the new value for the oracle (e.g. Chainlink nodes abandon the oracle, chain congestion, vulnerability/attacks on the chainlink system) consumers of this contract may continue using outdated stale or incorrect data (if oracles are unable to submit no new round is started). `aggregator.latestRoundData();` is missing additional validation to ensure that the round is complete and has returned a valid/expected price

Would we want to add in these checks as a safety mechanism (especially since we have multiple price feeds)?

LayneHaber (changes requested):
Need to update the changelog.

Also, ran `yarn workspace @connext/nxtp-contracts coverage` and it seems like there are some missing tests in the `ConnextPriceOracle.sol` from the report:

```
--------------------------|----------|----------|----------|----------|----------------|
File                      |  % Stmts | % Branch |  % Funcs |  % Lines |Uncovered Lines |
--------------------------|----------|----------|----------|----------|----------------|
 contracts/               |    97.11 |    90.37 |    97.37 |    97.18 |                |
  ConnextPriceOracle.sol  |    96.67 |    92.31 |    90.91 |    96.67 |        115,168 |
  ```

File: packages/contracts/contracts/ConnextPriceOracle.sol (hunk 1 above)
Comment (wanglonghong) [reply packages/contracts/contracts/ConnextPriceOracle.sol:145]: In next line, `require(msg.sender == admin, "only admin can set v1PriceOracle address");` so what you want is to make this line as modifier???

File: packages/contracts/test/connextPriceOracle.spec.ts (hunk 2 above)
Comment (wanglonghong) [reply packages/contracts/test/connextPriceOracle.spec.ts:268]: As far as I know, the `latestRoundData` of chainlink is the one that is already completed. It seems the reporter is just thinking about `getRoundData`. When we use `getRoundData`, we have to definitely check if that round is completed or not. 
I couldn't find a more better way right now.

**Comment by wanglonghong** [comment]
> Can you add the v1 price oracle addresses to the deploy script so that when we deploy we can automatically set the price oracle?

yep!

**Comment by wanglonghong** [comment]
> Need to update the changelog.
> 
> Also, ran `yarn workspace @connext/nxtp-contracts coverage` and it seems like there are some missing tests in the `ConnextPriceOracle.sol` from the report:
> 
> ```
> --------------------------|----------|----------|----------|----------|----------------|
> File                      |  % Stmts | % Branch |  % Funcs |  % Lines |Uncovered Lines |
> --------------------------|----------|----------|----------|----------|----------------|
>  contracts/               |    97.11 |    90.37 |    97.37 |    97.18 |                |
>   ConnextPriceOracle.sol  |    96.67 |    92.31 |    90.91 |    96.67 |        115,168 |
> ```

great catch!

**Review by rhlsthrm — APPROVED** [review]

File: packages/contracts/contracts/ConnextPriceOracle.sol (hunk 1 above)
Comment (LayneHaber) [reply packages/contracts/contracts/ConnextPriceOracle.sol:145]: If it's used in other places, yeah i think it makes sense. right now it is used 5 times with the only major difference being the error messages.

File: packages/contracts/test/connextPriceOracle.spec.ts (hunk 2 above)
Comment (LayneHaber) [reply packages/contracts/test/connextPriceOracle.spec.ts:281]: Here is the full text of the feedback:

> The ChainlinkAdapter calls out to a Chainlink oracle receiving the latestRoundData(). If there is a problem with Chainlink starting a new round and finding consensus on the new value for the oracle (e.g. Chainlink nodes abandon the oracle, chain congestion, vulnerability/attacks on the chainlink system) consumers of this contract may continue using outdated stale or incorrect data (if oracles are unable to submit no new round is started).
```
    function getPriceFromChainlink(address _tokenAddress) public view returns (uint256) {
        AggregatorV3Interface aggregator = aggregators[_tokenAddress];
        if (address(aggregator) != address(0)) {
            ( , int answer, , , ) = aggregator.latestRoundData();

            // It's fine for price to be 0. We have two price feeds.
            if (answer == 0) {
                return 0;
            }

            // Extend the decimals to 1e18.
            uint retVal = uint(answer);
            uint price = retVal.mul(10**(18 - uint(aggregator.decimals())));

            return price;            
        }
        return 0;        
    }
```
> aggregator.latestRoundData(); is missing additional validation to ensure that the round is complete and has returned a valid/expected price.

> Consider validating the output of latestRoundData() to match the following code snippet:

    function getPriceFromChainlink(address _tokenAddress) public view returns (uint256) {
        AggregatorV3Interface aggregator = aggregators[_tokenAddress];
        if (address(aggregator) != address(0)) {
            ( roundId, answer, , updateTime, answeredInRound ) = aggregator.latestRoundData();

            require(updateTime != 0, "Incomplete round");
            require(answeredInRound >= roundId, "Stale price");

            // It's fine for price to be 0. We have two price feeds.
            if (answer == 0) {
                return 0;
            }

            // Extend the decimals to 1e18.
            uint retVal = uint(answer);
            uint price = retVal.mul(10**(18 - uint(aggregator.decimals())));

            return price;            
        }
        return 0;        
    }

**Review by LayneHaber — COMMENTED** [review]
Coverage looks much better!

```
--------------------------|----------|----------|----------|----------|----------------|
File                      |  % Stmts | % Branch |  % Funcs |  % Lines |Uncovered Lines |
--------------------------|----------|----------|----------|----------|----------------|
 contracts/               |    97.74 |    90.83 |    98.67 |     97.8 |                |
  ConnextPriceOracle.sol  |      100 |    96.15 |      100 |      100 |                |
```

still not explicitly approving pending my responses to previous review comments

packages/contracts/test/connextPriceOracle.spec.ts
wanglonghong: I am still not sure about this strict check. Here is an important comment which was written in chainlink aggregator sources deployed to live networks. 
  /**
   * @notice get data about the latest round. Consumers are encouraged to check
   * that they're receiving fresh data by inspecting the updatedAt and
   * answeredInRound return values.
   * Note that different underlying implementations of AggregatorV3Interface
   * have slightly different semantics for some of the return values. Consumers
   * should determine what implementations they expect to receive
   * data from and validate that they can properly handle return data from all
   * of them.
   * @return roundId is the round ID from the aggregator for which the data was
   * retrieved combined with a phase to ensure that round IDs get larger as
   * time moves forward.
   * @return answer is the answer for the given round
   * @return startedAt is the timestamp when the round was started.
   _*** (Only some AggregatorV3Interface implementations return meaningful values)**_
   * @return updatedAt is the timestamp when the round last was updated (i.e.
   * answer was last computed)
   * @return answeredInRound is the round ID of the round in which the answer
   * was computed.
   **_* (Only some AggregatorV3Interface implementations return meaningful values)_**
   * @dev Note that answer and updatedAt may change between queries.
   */

LayneHaber (approved):
Looks good!

PR #756, opened 2022-01-17
Fix logs

LayneHaber:
## The Problem

- The `handling tracker` log being at info blows up logs when searching by `transactionId`
- Not waiting safe confirmations when handling `ReceiverCancelled`
- Add more logs about the receipt you are responding to when you have waited safe confirmations

## The Solution

- Move noisy logs to `debug`
- Wait safe confirmations for `ReceiverCancelled`
- Add logs to each action about the confirmed receipt for easier reorg debugging

## Checklist

- [x] Unit tests
- [x] Update CHANGELOG.md.

### DIFF
--- a/packages/router/src/bindings/contractReader/index.ts
+++ b/packages/router/src/bindings/contractReader/index.ts
@@ -213,7 +213,13 @@ export const handleSingle = async (
     }
     const preparePayload: PreparePayload = _transaction.payload;
     try {
-      logger.info("Preparing receiver", requestContext, methodContext);
+      logger.info("Preparing receiver", requestContext, methodContext, {
+        senderPrepareTransaction: {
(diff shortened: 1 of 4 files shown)


packages/router/src/bindings/contractReader/index.ts
```diff
@@ -62,13 +62,14 @@ export const bindContractReader = async () => {
           };
         });
 
-        logger.info("active transactions tracker", requestContext, methodContext, {
+        logger.info("Trackers overview", requestContext, methodContext, {
```
rhlsthrm, line 65: I really like having the tracker in info tbh.

LayneHaber: its there in debug, but when you're trying to dig down on something it blows up the logs

wanglonghong (approved):
lgtm

PR #777, opened 2022-01-27
Amarok loadtest

rhlsthrm:
## Checklist

- [ ] Run load tests.
- [ ] Update CHANGELOG.md.


packages/integration/test/load/loadTest.ts
```diff
@@ -0,0 +1,81 @@
...
+
+class LoadTestEnvironment{
+  private config:Config;
+  private testConfig:LoadTestConfig;
+  private containersUp!:boolean;
+
+   constructor(testParams:LoadTestConfig){
+    this.config = getConfig();
+    this.testConfig = testParams;
+    if(testParams.spawnContainters){
```
rhlsthrm, line 17: Need to get your prettier running format on save...

rhlsthrm (commented):
Looks great so far, from what I see no reason why we can't merge this now, and keep building in pieces.

The next goal should be to have an integration.spec running so that we can replace it with this.

packages/integration/test/load/loadTest.ts
base0010: woops fixed in my IDE

PR #781, opened 2022-01-28
add watch tower messaging to utils and router

rhlsthrm:
## Checklist

- [ ] Run load tests.
- [ ] Update CHANGELOG.md.


packages/router/src/lib/helpers/fulfill.ts
```diff
@@ -1,3 +1,5 @@
 import { signRouterFulfillTransactionPayload as _signRouterFulfillTransactionPayload } from "@connext/nxtp-utils";
 
 export const signRouterFulfillTransactionPayload = _signRouterFulfillTransactionPayload;
+
+export const WATCH_TOWER_BUFFER = 2 * 3600; // time buffer if watchtower can handle fulfill
```
jakekidd, line 5: Let's call this `WATCHTOWER_CALL_TIMEOUT` ?

jakekidd: Also, I *really* dislike this way of doing things. Whatever timeout we set here determines how long the window is for batching on the watchtower is... that's the same constant in two codebases.

A better idea: let's have the watchtower have an API to check whether a fulfill got collected into the latest batch.

jakekidd: Opened an issue for this improvement https://github.com/connext/watchtower/issues/15

jakekidd (changes requested):
See my comment regarding usage of a "batch status API" as well as this issue here:

https://github.com/connext/watchtower/issues/15

jakekidd (approved):
batch status api added, lgtm!

PR #792, opened 2022-02-06
[sdk] remove promise before interval

sanchaymittal:
## Checklist

- [ ] Run load tests.
- [ ] Update CHANGELOG.md.

### DIFF
--- a/packages/sdk/src/subgraph/subgraph.ts
+++ b/packages/sdk/src/subgraph/subgraph.ts
@@ -156,12 +156,11 @@ export class Subgraph {
   }
 
   async startPolling(): Promise<void> {
-    await Promise.all(
-      Object.keys(this.sdks).map(async (_chainId) => {
-        const chainId = parseInt(_chainId);
-        this.pollingStopperBlock[chainId] = await this.chainReader.getBlockNumber(chainId);
-      }),
-    );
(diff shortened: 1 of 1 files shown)


packages/sdk/src/subgraph/subgraph.ts
```diff
@@ -156,12 +156,11 @@ export class Subgraph {
   }
 
   async startPolling(): Promise<void> {
-    await Promise.all(
-      Object.keys(this.sdks).map(async (_chainId) => {
-        const chainId = parseInt(_chainId);
-        this.pollingStopperBlock[chainId] = await this.chainReader.getBlockNumber(chainId);
-      }),
-    );
+    Object.keys(this.sdks).map(async (_chainId) => {
```
rhlsthrm, line 159: Isn't this a problem? This is going to be racing with the below code now.

rhlsthrm (changes requested):

packages/sdk/src/subgraph/subgraph.ts
sanchaymittal: problem is coming cause lifi and xpollinate both initiate waitFor after initiating the sdk. And pollingLoop was undefined cause of promise
So this should work now.

rhlsthrm (approved):

PR #814, opened 2022-03-03
Amarok Redis Changes

base0010:
## The Problem
The redis "schema" 

## The Solution

Need to store some elements in ordered sets, and get the latest scores of records to assist in that; as the SCAN command doesn't guarantee the order of elements returned. 

## Checklist

- [x ] Tested locally
- [ ] Write better test cases/get it to play nice with nyc/mocah


base0010:
not supposed to go in main

PR #867, opened 2022-03-17
840 test upgradability

liu-zhipeng:
## Checklist

- [ ] Run load tests.
- [ ] Update CHANGELOG.md.


packages/deployments/contracts/deploy/deploy.ts
```diff
@@ -1,48 +1,90 @@
...
+  hre: HardhatRuntimeEnvironment,
+) => {
+  const artifact = await hre.deployments.getExtendedArtifact(artifactName);
+  const deploymentSubmission: DeploymentSubmission = {
+    address: address,
+    args: args,
+    ...artifact,
+  };
+
+  hre.deployments.save(`${name}`, deploymentSubmission);
```
LayneHaber, line 21: can just put in `name` directly since it's already a string

packages/deployments/contracts/deploy/deploy.ts
```diff
@@ -1,48 +1,90 @@
...
+  address: string,
+  name: string,
+  args: any[],
+  artifactName: string,
+  hre: HardhatRuntimeEnvironment,
+) => {
+  const artifact = await hre.deployments.getExtendedArtifact(artifactName);
+  const deploymentSubmission: DeploymentSubmission = {
+    address: address,
+    args: args,
```
LayneHaber, line 17: should change to 

```js
{
    ...
    address,
    args
}
```

packages/deployments/contracts/deploy/deploy.ts
```diff
@@ -1,48 +1,90 @@
...
-  console.log(`${name} proxyAddress:`, proxyAddress);
-  return hre.ethers.getContractAt(proxyDeployment.abi, proxyDeployment.address);
+  if (proxyDeployment) {
+    console.log(`${name} proxy deployed. upgrading........`);
+    beaconAddress = await hre.upgrades.erc1967.getBeaconAddress(proxyDeployment.address);
+    await hre.upgrades.upgradeBeacon(beaconAddress, factory);
+
+    console.log(`${name} proxy upgraded.`);
+
+    implementation = await hre.upgrades.beacon.getImplementationAddress(beaconAddress);
```
LayneHaber, line 48: Wouldn't the implementation always have to be redeployed when upgrading? Then set on the beacon?

packages/deployments/contracts/deploy/deploy.ts
```diff
@@ -1,48 +1,90 @@
...
+    proxy = await hre.ethers.getContractAt(proxyDeployment.abi, proxyDeployment.address);
+  } else {
+    // Deploy new implementation
+    const beacon = await hre.upgrades.deployBeacon(factory);
```
LayneHaber, line 52: Really cool you can do this with the plugin!

packages/deployments/contracts/deploy/deploy.ts
```diff
@@ -177,6 +207,7 @@ const func: DeployFunction = async (hre: HardhatRuntimeEnvironment): Promise<voi
       from: deployer.address,
       args: [WRAPPED_ETH_MAP.get(+chainId)],
       log: true,
+      skipIfAlreadyDeployed: true,
```
LayneHaber, line 210: Do we know if this matches the compiled bytecode against what's onchain? Seems like it should, but just curious.

If it does, and it works for proxies, we should do this with things like the `BridgeRouter`, `TokenRegistry`, etc.

packages/deployments/contracts/src/tasks/mintTestToken.ts
```diff
@@ -3,32 +3,31 @@ import { task } from "hardhat/config";
 
 export default task("mint", "Mint test tokens")
   .addParam("amount", "Amount (real units)")
-  .addOptionalParam("to", "Override address to mint to")
+  .addParam("mintto", "Override address to mint to")
```
LayneHaber, line 6: we should keep this as `to` vs `mintto`

liu-zhipeng: updated

packages/deployments/contracts/deploy/deploy.ts
liu-zhipeng: implementation won't be redeployed if not updated.

packages/deployments/contracts/deploy/deploy.ts
liu-zhipeng: Sure.

packages/deployments/contracts/deploy/deploy.ts
liu-zhipeng: `const deployedImpl = await hre.upgrades.prepareUpgrade(beaconAddress, factory);`

prepareUpgrade will be the solution. finally fixed.
So if the implementation wasn't changed, deploying script won't redeploy implementation and call `upgradeTo` as well.

Great plugin if updated a bit!

PR #886, opened 2022-03-24
865 origin properties

### LINKED ISSUE #865: [contracts] Set origin properties on fulfill
Should set the `sender` and `originDomain` for the slow-liquidity path of `fulfill` to allow permissioned calls to pass by checking these properties. They can be accessed via reentrancy a-la `ReentrancyGuard` or the `L2CrossDomainMessenger` from optimism. These should be stored in a single storage slot (could make use of `TypedMemView` library).

Names may be off depending on #859

LayneHaber:
## The Problem

Fixes #865 by adding reentrant properties. Uses `TypedMemView` to take up one storage slot for both properties, but changes the `process` interface.

Before merging needs:
- [x] Forge tests working
- [x] Forge test coverage of `LibCrossDomainProperty` and `Executor`

## Checklist

- [ ] Run load tests.
- [ ] Update CHANGELOG.md.


packages/deployments/contracts/contracts/interpreters/Executor.sol
```diff
@@ -11,92 +13,138 @@ import "@openzeppelin/contracts-upgradeable/token/ERC20/IERC20Upgradeable.sol";
...
-   *        calldata execution
-   * @param to The address to execute the calldata on
-   * @param assetId The assetId of the funds to approve to the contract or
-   *                send along with the call
-   * @param amount The amount to approve or send with the call
-   * @param callData The data to execute
+   * call data executes can be payable, and will have `amount` sent
+   * along with the function (or approved to the contract). If the
+   * call fails, rather than reverting, funds are sent directly to 
+   * some provided fallbaack address
```
jakekidd, line 86: fallbaaack

jakekidd: to the choppah

packages/deployments/contracts/contracts/interpreters/Executor.sol
```diff
@@ -11,92 +13,138 @@ import "@openzeppelin/contracts-upgradeable/token/ERC20/IERC20Upgradeable.sol";
...
+    // it is the wrong type on conversion, and revert occurs with empty type as
+    // well
+    properties = _properties;
+  
+    // Try to execute the callData
+    // the low level call will return `false` if its execution reverts
+    (success, returnData) = _to.call{value: isNative ? _amount : 0}(_callData);
+    
+    // Unset properties
+    properties = LibCrossDomainProperty.EMPTY_BYTES;
```
jakekidd, line 129: this usage of temporary storage `properties` is cool... I can't help but wonder if it would be more gas efficient to optionally allow the user to specify a type of an interpreter layer that accepts an additional argument, like msg.sender, but I think there's additional complexity that comes with that.

LayneHaber: This is so you can allow the `to` contract to do this:

```ts
IExecutor(msg.sender).originSender()
IExecutor(msg.sender).origin()
```

within whatever calldata you're passing through. adding in params makes it difficult because you wouldn't be able to use the low-level `call` (can't insert params there) so you would be forcing everyone to write a middleware contract

LayneHaber: that being said, it likely would be more gas efficient in cases where you don't need to check those properties

jakekidd: yeah, makes sense! will properties *always* include `msg.sender` then? are there any other properties that will for sure be specified? lol

jakekidd (approved):

liu-zhipeng (approved):

PR #892, opened 2022-03-28
[cache] Index transfers by transfer ID and collate

jakekidd:
## The Problem

In current model, we basically publish the `NewXCall` event  whenever there's a new `xcall` in general. This will result in the router immediately getting hundreds/thousands/etc `NewXCall` events fired off every time it boots and it's behind the current state on-chain. As a result, the router would be spamming the seq. with bids.

While we'll probably have some cold-start logic in the future (e.g. put a nonce to start with in the config for each chain, etc), I think it would be best to just go ahead and collate the data at some point in the process here in order to derive which `NewXCall` events are "valid".

## The Solution

TL;DR: "Don't publish NewXCall if it's not a New XCall"

I propose we do the collation in the cache itself. The update for this is pretty straightforward. We would move from indexing transfer data by domain+transferId -> indexing it *just* by `transferId` alone.

The upsert process (which results effectively in collation) would then look like this:
- look up `transferId` in current cache
- sanity check: confirm that the core transfer data / metadata stuff is the same (otherwise, just assume it's invalid... this shouldn't be possible/happen, but just for sanity)
- combine the data for the new given transfer with the existing cached transfer (if there is an existing cached transfer). This will result in combining the origin chain Transfer subgraph entity (with `xcall` property filled out) with the destination chain subgraph entity (with `execute` and `reconcile`, if those txs were completed)

NOTE: This won't achieve *perfect* knowledge of whether a `NewXCall` is truly a new xcall, since it's possible for the destination subgraph (or endpoint) to lag or fail for a bit, for instance. But it's much better than blindly forward every XCall the second we see it's an XCall.

This will also allow us to condense our storage/caching significantly! Otherwise, we'd be storing a lot of the same metadata for on two domains.

### DIFF
--- a/packages/adapters/cache/src/lib/caches/transactions.ts
+++ b/packages/adapters/cache/src/lib/caches/transactions.ts
@@ -139,9 +139,9 @@ export class TransactionsCache extends Cache {
       if (xcall.transactionHash && !execute?.transactionHash && !reconcile?.transactionHash) {
         // If xcall but no execute or reconcile, then it's possibly a new transfer.
         newXCalls[transferId] = stringified;
-      } else if (execute?.transactionHash) {
-        // If execute, then it's a completed transfer. In case we've previously marked the xcall as new in
-        // this batch, we need to remove it from the newXCalls list.
+      } else if (execute?.transactionHash || reconcile?.transactionHash) {
+        // If execute (or reconcile), then it's a completed transfer. In case we've previously recorded
+        // the xcall as new in this batch, we need to remove it from the newXCalls list.
         delete newXCalls[transferId];
       }
       // Retrieve latest nonce for this domain.
--- a/packages/adapters/cache/src/lib/caches/transactions.ts
+++ b/packages/adapters/cache/src/lib/caches/transactions.ts
@@ -39,7 +39,7 @@ export class TransactionsCache extends Cache {
     const status = this.data.scanStream({
       match: `${txid}`,
     });
-    return new Promise((res, _) => {
+    return new Promise((res) => {
       status.on("data", (txidMatch: string) => {
         this.logger.debug("found txid");
         const val = this.data.get(txidMatch);
@@ -76,7 +76,7 @@ export class TransactionsCache extends Cache {
       count: 1,
     });
     let txData: XTransfer;
-    return new Promise((res, _) => {
+    return new Promise((res) => {
       txDataStream.on("data", async (data: string) => {
         const crossChainData = await this.data.hget(`${this.prefix}:${domain}`, data);
         txData = JSON.parse(crossChainData!) as XTransfer;
@@ -104,7 +104,7 @@ export class TransactionsCache extends Cache {
     });
 
     let txData: XTransfer;
-    return new Promise((res, _) => {
+    return new Promise((res) => {
       txDataStream.on("data", async (data: string) => {
         const crossChainData = await this.data.hget(`${this.prefix}:${domain}`, data);
         txData = JSON.parse(crossChainData!) as XTransfer;
@@ -116,7 +116,7 @@ export class TransactionsCache extends Cache {
   }
 
   public async storeTransfers(transfers: XTransfer[]): Promise<void> {
-    const nonceIncreasesByDomain: { [domain: string]: boolean } = {};
+    const nonceDidIncreaseForDomain: { [domain: string]: boolean } = {};
     const highestNonceByDomain: { [domain: string]: number } = {};
     const newXCalls: { [transferId: string]: string } = {};
     for (let transfer of transfers) {
@@ -140,7 +140,7 @@ export class TransactionsCache extends Cache {
         // If xcall but no execute or reconcile, then it's possibly a new transfer.
         newXCalls[transferId] = stringified;
       } else if (execute?.transactionHash || reconcile?.transactionHash) {
-        // If execute (or reconcile), then it's a completed transfer. In case we've previously recorded
+        // If execute (or reconcile), then it's a stale xcall. In case we've previously recorded
         // the xcall as new in this batch, we need to remove it from the newXCalls list.
         delete newXCalls[transferId];
       }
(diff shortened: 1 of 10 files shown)


packages/adapters/cache/src/lib/caches/transactions.ts
```diff
@@ -39,7 +39,7 @@ export class TransactionsCache extends Cache {
     const status = this.data.scanStream({
       match: `${txid}`,
     });
-    return new Promise((res, rej) => {
+    return new Promise((res, _) => {
```
LayneHaber, line 42: You don't need to include it if you're not going to use it. That being said if it exits in anyway that is not handled by your `res` then it will just hang

packages/adapters/cache/src/lib/caches/transactions.ts
```diff
@@ -104,21 +115,57 @@ export class TransactionsCache extends Cache {
     });
   }
 
-  public async storeTxData(txs: XTransfer[]): Promise<void> {
-    for (const tx of txs) {
+  public async storeTransfers(transfers: XTransfer[]): Promise<void> {
+    const nonceIncreasesByDomain: { [domain: string]: boolean } = {};
```
LayneHaber, line 119: can make the types a bit cleaner by doing: `Record<string, boolean>` but then its not labeled

packages/adapters/cache/src/lib/caches/transactions.ts
```diff
@@ -104,21 +115,57 @@ export class TransactionsCache extends Cache {
...
+    const newXCalls: { [transferId: string]: string } = {};
+    for (let transfer of transfers) {
+      const existing = await this.getTransferByTransferId(transfer.transferId);
+      // TODO: Sanity check: no update needed if this transfer is same as the one already stored!
+
+      // TODO: Sanity check: make sure that the existing transfer's metadata is the same as the old one!
+
+      // Update the existing transfer with the data from the new one; this will collate the transfer across
+      // domains, since our cache is indexed by transferId.
+      transfer = existing ? { ...existing, ...transfer } : transfer;
```
LayneHaber, line 130: What fields are in the `transfer`? would you want to always override existing fields with the newly pulled ones?

packages/adapters/cache/src/lib/caches/transactions.ts
```diff
@@ -104,21 +115,57 @@ export class TransactionsCache extends Cache {
...
       // gte(1) => added, 0 => updated,
       // reference: https://redis.io/commands/hset
-      await this.data.hset(`${this.prefix}:${tx.originDomain}`, `${tx.nonce}:${tx.transferId}`, JSON.stringify(tx));
-      //move pointer to latest Nonce
-      const latestNonce = (await this.data.hget(`${this.prefix}:${tx.originDomain}`, "latestNonce")) ?? "0";
-      if (Number(tx.nonce) > parseInt(latestNonce)) {
-        //if this txns nonce is > the current pointer to latest nonce point to this one now
-        await this.data.hset(`${this.prefix}:${tx.originDomain}`, "latestNonce", tx.nonce);
-        await this.data.publish(StoreChannel.NewHighestNonce, tx.nonce.toString());
+      await this.data.hset(`${this.prefix}:transfers`, `${transferId}`, stringified);
```
LayneHaber, line 138: What do the race conditions look like here? Wondering if there's anything that could cause a problem with that

packages/adapters/cache/src/lib/caches/transactions.ts
```diff
@@ -104,21 +115,57 @@ export class TransactionsCache extends Cache {
...
+      if (xcall.transactionHash && !execute?.transactionHash && !reconcile?.transactionHash) {
+        // If xcall but no execute or reconcile, then it's possibly a new transfer.
+        newXCalls[transferId] = stringified;
+      } else if (execute?.transactionHash || reconcile?.transactionHash) {
+        // If execute (or reconcile), then it's a completed transfer. In case we've previously recorded
+        // the xcall as new in this batch, we need to remove it from the newXCalls list.
+        delete newXCalls[transferId];
```
LayneHaber, line 145: in slow liquidity, you will still need to `execute` the transfer if it is `reconcile`d. i'm assuming this is just for sending bids, but if its for maintaining the cache for any future actions then this could be problematic

packages/adapters/cache/src/lib/caches/transactions.ts
LayneHaber: also maybe we can rename to `didNonceIncreaseByDomain` or something because `nonceIncreases...` makes it sound like you're storing the delta between nonces there

packages/adapters/cache/src/lib/caches/transactions.ts
jakekidd: oops right

packages/adapters/cache/src/lib/caches/transactions.ts
jakekidd: ty

packages/adapters/cache/src/lib/caches/transactions.ts
jakekidd: see type `XTransfer` from utils

yes, pretty much always override... given that I add those two sanity checks above lol

packages/adapters/cache/src/lib/caches/transactions.ts
jakekidd: we execute every `hset` one at a time. if we executed them in parallel with a Promise.all, i'm sure we'd run into some issues updating the `newXCalls` object lol...

otherwise, no race afaik... assuming we don't call this method twice....

we could use a thread-safe flag to prevent multiple calls running (i.e. p-queue to serialize). Should I do that? hmmm

packages/adapters/cache/src/lib/caches/transactions.ts
jakekidd: oh you right, that comment is misleading. I should probs say "then it's a stale xcall"

rhlsthrm:
Nice!

PR #940, opened 2022-04-06
925 poll cache

### LINKED ISSUE #925: [router] Poll cache for txs to bid on
Instead of pub-sub on the prepared txs in routers, we should poll the xcalled txs. This will give us redundancy and crash-tolerance. The router should poll every 30 (?) seconds and submit a bid for every tx that is xcalled. 

In order for this to be effective, we need to make sure the cache status is updated with transfers that get `Executed` on the destination side.

jakekidd:
Fixes #925

### DIFF
--- a/packages/adapters/cache/src/lib/caches/transfers.ts
+++ b/packages/adapters/cache/src/lib/caches/transfers.ts
@@ -1,9 +1,9 @@
 import { XTransfer } from "@connext/nxtp-utils";
 
 import { StoreChannel } from "../entities";
+import { getHelpers } from "../helpers";
 
 import { Cache } from ".";
-import { getHelpers } from "../helpers";
 
 /**
  * Redis Store Details:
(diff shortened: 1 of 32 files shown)


packages/adapters/cache/src/lib/caches/transfers.ts
```diff
@@ -106,46 +111,46 @@ export class TransfersCache extends Cache {
...
-    //   });
-    // });
-    // TODO: Reimplement status API for transfers?
-    throw new Error("Not implemented");
+  private async removePending(domain: string, transferId: string): Promise<boolean> {
+    const currentPending = await this.getPending(domain);
+    const index = currentPending.findIndex((id) => id === transferId);
+    if (index >= 0) {
+      currentPending.splice(index, 1);
+      await this.data.set(`${this.prefix}:pending:${domain}`, JSON.stringify(currentPending));
```
wanglonghong, line 151: I don't think it is proper solution to store pending transfers. If there are a lot of pending transfers, it can take much longer to read. 
I recommend to use `hset` insteadof set

jakekidd: good catch, ty

PR #942, opened 2022-04-06
add relayer whitelist

liu-zhipeng:
## Checklist

- [ ] Run load tests.
- [ ] Update CHANGELOG.md.


packages/deployments/contracts/contracts/Connext.sol
```diff
@@ -273,6 +282,19 @@ contract Connext is
...
+   * @param approved - approved or not
+   */
+  function setupRelayer(address relayer, bool approved) external override onlyOwner {
+    if (relayer == address(0)) revert Connext__setupRelayer_badAddress();
+    if (approvedRelayers[relayer] == approved) revert Connext__setupRelayer_alreadyDid();
+    approvedRelayers[relayer] = approved;
+
+    emit RelayerSetup(relayer, approved);
+  }
+
```
LayneHaber, line 297: I'm inclined to have separate functions for adding and removing relayers (same as we do with assets/routers)

packages/deployments/contracts/contracts/Connext.sol
```diff
@@ -515,7 +537,7 @@ contract Connext is
       // Send funds to the user
       AssetLogic.transferAssetFromContract(adopted, _args.params.to, amount, wrapper);
     } else {
-      // Send funds to interprepter
+      // Send funds to executor
```
LayneHaber, line 540: 👍

packages/deployments/contracts/contracts/Connext.sol
```diff
@@ -679,6 +701,11 @@ contract Connext is
       return;
     }
 
+    // If the sender is not approved relayer, revert()
+    if (!approvedRelayers[msg.sender]) {
+      revert Connext__handleRelayerFees_notApprovedRelayer();
+    }
```
LayneHaber, line 707: @rhlsthrm we need the address of the gelato exec contract to make sure we don't break testnet

packages/deployments/contracts/contracts/interfaces/IConnext.sol
```diff
@@ -131,6 +131,13 @@ interface IConnext {
    */
   event AssetRemoved(bytes32 canonicalId, address caller);
 
+  /**
+   * @notice Emitted when a rlayer is added or removed from whitelists
+   * @param relayer - The relayer address to be added or removed
+   * @param approved - Added or removed?
+   */
+  event RelayerSetup(address relayer, bool approved);
```
LayneHaber, line 139: See above comments re: separate codepaths

packages/deployments/contracts/src/tasks/preflight.ts
```diff
@@ -174,5 +178,15 @@ export default task("preflight", "Ensure correct setup for e2e demo with a speci
       } else {
         console.log("*** Sufficient relayer fees present!");
       }
+
+      // Add relayer
+      const relayer = _relayer ?? process.env.RELAYER_ADDRESS;
+      if (relayer && isAddress(relayer)) {
```
LayneHaber, line 184: should error if it is not an address

LayneHaber (changes requested):

packages/deployments/contracts/contracts/Connext.sol
liu-zhipeng: Trying to reduce contract bytecode size as much as I can. lol

Fixed.
[https://github.com/connext/nxtp/commit/ba57917bf911ee3be020d48634f3518ffe8a8750](url)

packages/deployments/contracts/src/tasks/preflight.ts
liu-zhipeng: https://github.com/connext/nxtp/commit/6caf990bba55cce35f08d56ef5ecf378bcb872cf

PR #948, opened 2022-04-07
Redis fix

rhlsthrm:
## The Problem

Redis was broken in prod.

## The Solution

Fix race conditions.

## Checklist

- [ ] Run load tests.
- [ ] Update CHANGELOG.md.


rhlsthrm:
> Could this be somehow tested by injecting latency on the redis calls? Would love some safety here.
Hmm I don't think so because we would need to inject latency into the redis callback implementation itself.

> Also, not on my laptop right now so can't check but would the sequencer also suffer from this issue?
This is the sequencer, but yes I looked and fixed all the functions that have this issue.

carlomazzaferro:
Makes sense. Alright then, I'm happy with this as long as we add a task to at least plan some integration testing that verifies the redis interaction works

PR #950, opened 2022-04-07
Adds Relayer fee router

lmcorbalan:
## The Problem

Subtask of #900 
Relayers need a RelayerFeesRouter to send a claim message from the remote domain to the origin domain where the fees are deposited

## The Solution

Create a RelayerFeeRouter contract
 
## Checklist

- [ ] Run load tests.
- [ ] Update CHANGELOG.md.


packages/deployments/contracts/contracts/nomad-xapps/contracts/relayer-fee-router/RelayerFeeRouter.sol
```diff
@@ -0,0 +1,190 @@
...
+
+// ============ External Imports ============
+import {Home} from "../../../nomad-core/contracts/Home.sol";
+import {Version0} from "../../../nomad-core/contracts/Version0.sol";
+import {TypedMemView} from "../../../nomad-core/libs/TypedMemView.sol";
+
+// import "../../../../lib/forge-std/src/console.sol";
+
+/**
+ * @title BridgeRouter
```
jakekidd, line 19: nit: `BridgeRouter` -> `RelayerFeeRouter`

packages/deployments/contracts/contracts/nomad-xapps/contracts/relayer-fee-router/RelayerFeeRouter.sol
```diff
@@ -0,0 +1,190 @@
...
+
+  function initialize(address _xAppConnectionManager, address _connext) public initializer {
+    connext = IConnext(_connext);
+    __XAppConnectionClient_initialize(_xAppConnectionManager);
+
+    emit SetConnext(_connext);
+  }
+
+  /**
+   * @notice Sets the transaction manager.
```
jakekidd, line 104: nit, should say `Sets connext address` ?

lol was this copy/pasted from somewhere? cause we might have to fix the source haha

packages/deployments/contracts/contracts/nomad-xapps/contracts/relayer-fee-router/RelayerFeeRouter.sol
```diff
@@ -0,0 +1,190 @@
...
+   * @param _amount The amount of fees to claim
+   * @param _transactionIds A group of transaction ids to claim for fee bumps
+   */
+  function send(
+    uint32 _domain,
+    address _recipient,
+    uint256 _amount,
+    bytes32[] calldata _transactionIds
+  ) external onlyConnext {
+    if (_amount == 0 && _transactionIds.length == 0) revert RelayerFeeRouter__send_claimEmpty();
```
jakekidd, line 128: Why are we sending `amount` ? @LayneHaber 

I feel like we just need to send the transactionIds, they should be mapped to the corresponding amounts that the user deposited on the origin chain

packages/deployments/contracts/contracts/nomad-xapps/contracts/relayer-fee-router/RelayerFeeRouter.sol
```diff
@@ -0,0 +1,190 @@
...
+    uint256 amount = _msg.amount();
+    bytes32[] memory transactionIds = _msg.transactionIds();
+
+    // TODO - claim fee in connext
+
+    // emit Receive event
+    emit Receive(_originAndNonce(_origin, _nonce), _origin, recipient, amount, transactionIds);
+  }
+
+  /**
```
jakekidd, line 170: can we use a utils library `_localDomain` and `originAndNonce`, and share it with `BridgeRouter` ? Makes unit testing easier later

packages/deployments/contracts/contracts/nomad-xapps/contracts/relayer-fee-router/RelayerFeeRouter.sol
```diff
@@ -0,0 +1,190 @@
...
+    return XAppConnectionClient._localDomain();
+  }
+
+  /**
+   * @notice Internal utility function that combines
+   * `_origin` and `_nonce`.
+   * @dev Both origin and nonce should be less than 2^32 - 1
+   * @param _origin Domain of chain where the transfer originated
+   * @param _nonce The unique identifier for the message from origin to destination
+   * @return Returns (`_origin` << 32) & `_nonce`
```
jakekidd, line 185: nit: `@return` uses `AND &` where we use `OR |` below lol

packages/deployments/contracts/contracts/nomad-xapps/contracts/relayer-fee-router/RelayerFeeRouter.sol
```diff
@@ -0,0 +1,190 @@
...
+
+  // ============ Modifiers ============
+
+  /**
+   * @notice Restricts the caller to the local bridge router
+   */
+  modifier onlyConnext() {
+    if (msg.sender != address(connext)) revert RelayerFeeRouter__onlyConnext_notConnext();
+    _;
+  }
```
jakekidd, line 92: Might be cool to inherit this `connext` property, modifier, initialize, etc from a shared `abstract` contract between `BridgeRouter` and `RelayerFeeRouter`

jakekidd: `RouterCore` ?

packages/deployments/contracts/contracts_forge/RelayerFeeRouter.t.sol
```diff
@@ -0,0 +1,205 @@
...
+  using RelayerFeeMessage for bytes29;
+
+  // ============ Libraries ============
+  using stdStorage for StdStorage;
+
+  event Send(
+    uint32 domain,
+    address recipient,
+    uint256 amount,
+    bytes32[] _transactionIds,
```
jakekidd, line 41: nit: doesn't need underscore here lol

but actually, we should put these `Send` and `Receive` events in an `interface IRelayerFeeBridge`

packages/deployments/contracts/contracts/nomad-xapps/contracts/relayer-fee-router/RelayerFeeRouter.sol
```diff
@@ -0,0 +1,190 @@
...
+    bytes memory _message
+  ) external override onlyReplica onlyRemoteRouter(_origin, _sender) {
+    // parse tokenId and action from message
+    bytes29 _msg = _message.ref(0).mustBeClaimFees();
+
+    address recipient = _msg.recipient();
+    uint256 amount = _msg.amount();
+    bytes32[] memory transactionIds = _msg.transactionIds();
+
+    // TODO - claim fee in connext
```
jakekidd, line 164: Would be cool to stub this for now

packages/deployments/contracts/contracts_forge/RelayerFeeRouter.t.sol
LayneHaber: If the contract size is small enough i'm actually okay with not having it an interface, but i recognize that's inconsistent throughout the repo. 

Also this is a test contract so really it should still be able to pull these events without defining them

packages/deployments/contracts/contracts_forge/RelayerFeeRouter.t.sol
```diff
@@ -0,0 +1,205 @@
...
+    relayerFeeRouterImplementation = new RelayerFeeRouter();
+
+    proxy = new ERC1967Proxy(
+      address(relayerFeeRouterImplementation),
+      abi.encodeWithSelector(RelayerFeeRouter.initialize.selector, xAppConnectionManager, connext)
+    );
+
+    relayerFeeRouter = RelayerFeeRouter(payable(address(proxy)));
+
+    // Shouldn't enrollRemoteRouter be protected by onlyOwner?
```
LayneHaber, line 94: Yes, see [here](https://github.com/nomad-xyz/monorepo/blob/73c873233e40d5a0f8c8a5fcd2785e8f9736ca55/packages/contracts-router/contracts/Router.sol#L34-L39). Removed it for testing purposes and seems I forgot to add it back

packages/deployments/contracts/contracts/nomad-xapps/contracts/relayer-fee-router/RelayerFeeRouter.sol
LayneHaber, line 128: should fail for either condition (`transactionId` only following amount fixes)

packages/deployments/contracts/contracts/nomad-xapps/contracts/relayer-fee-router/RelayerFeeRouter.sol
LayneHaber: I think the return comment is designed to represent the semantic happenings not the actual operators used (i.e. returns left-shifted origin + nonce, not either or)

packages/deployments/contracts/contracts_forge/RelayerFeeRouter.t.sol
```diff
@@ -0,0 +1,205 @@
...
+    assertEq(address(relayerFeeRouter.xAppConnectionManager()), address(xAppConnectionManager));
+    assertEq(address(relayerFeeRouter.connext()), address(connext));
+  }
+
+  // ============ setConnext ============
+
+  // Should work
+  function testSetConnext() public {
+    vm.expectEmit(true, true, true, true);
+    emit SetConnext(connext2);
```
LayneHaber, line 109: Should add a similar event to the `BridgeRouter` (can be done via `ConnextRouter` contract)

packages/deployments/contracts/contracts_forge/RelayerFeeRouter.t.sol
```diff
@@ -0,0 +1,205 @@
...
+
+    bytes32[] memory transactionIds = new bytes32[](2);
+    transactionIds[0] = "AAA";
+    transactionIds[1] = "BBB";
+
+    relayerFeeRouter.send(remoteDomain, address(0), 1, transactionIds);
+  }
+
+  // Should work
+  function testSend(bytes32[] calldata _transactionIds, uint256 _amount) public {
```
LayneHaber, line 156: very nice! been neglecting property-based tests but would be good to start leaning into them

packages/deployments/contracts/contracts_forge/RelayerFeeRouter.t.sol
```diff
@@ -0,0 +1,205 @@
...
+    bytes memory message = RelayerFeeMessage.formatClaimFees(recipient, _amount, _transactionIds);
+
+    vm.expectCall(home, abi.encodeWithSelector(MockHome.dispatch.selector, remoteDomain, remote, message));
+    vm.expectEmit(true, true, true, true);
+    emit Send(remoteDomain, recipient, _amount, _transactionIds, remote, message);
+
+    relayerFeeRouter.send(remoteDomain, recipient, _amount, _transactionIds);
+  }
+
+  function testSend0Amount(bytes32[] calldata _transactionIds) public {
```
LayneHaber, line 167: Should make sure to add comments before every test thats more descriptive than the name.

I've been wondering if we should move to a test name convention of something like: `test_${ContractName}__${function}${test number for contract}` (i.e. `test_RelayerFeeRouter__send1`) so you can ensure when you run `test:forge -m ${name}` the test name is unique.

Thoughts?

packages/deployments/contracts/contracts_forge/RelayerFeeRouter.t.sol
```diff
@@ -0,0 +1,205 @@
...
+  // ============ handle ============
+  // Should work
+  function testHandle(
+    bytes32[] calldata _transactionIds,
+    uint32 _nonce,
+    uint256 _amount
+  ) public {
+    vm.assume(_amount != 0 || _transactionIds.length != 0);
+    uint64 originAndNonce = (uint64(remoteDomain) << 32) | _nonce;
+    bytes memory message = RelayerFeeMessage.formatClaimFees(recipient, _amount, _transactionIds);
```
LayneHaber, line 187: Because there is no assertion of handle or its effects, it's hard to say if the message is formatting/parsing properly. 

I would add in a separate message test contract with some simple tests that assert you can pull the proper values out of the formatted message

packages/deployments/contracts/contracts_forge/RelayerFeeRouter.t.sol
liu-zhipeng: yeah, i agree, we should have any special naming rule for forge tests. it was very annoying to find some meaningful name for test function.

packages/deployments/contracts/contracts/nomad-xapps/contracts/relayer-fee-router/RelayerFeeMessage.sol
```diff
@@ -0,0 +1,125 @@
...
+
+library RelayerFeeMessage {
+  // ============ Libraries ============
+
+  using TypedMemView for bytes;
+  using TypedMemView for bytes29;
+
+  // ============ Enums ============
+
+  // WARNING: do NOT re-write the numbers / order
```
liu-zhipeng, line 17: Good warning!

packages/deployments/contracts/contracts/nomad-xapps/contracts/relayer-fee-router/RelayerFeeMessage.sol
```diff
@@ -0,0 +1,125 @@
...
+  // of message types in an upgrade;
+  // will cause in-flight messages to be mis-interpreted
+  enum Types {
```
liu-zhipeng, line 20: good warning!

packages/deployments/contracts/contracts/nomad-xapps/contracts/relayer-fee-router/RelayerFeeMessage.sol
```diff
@@ -0,0 +1,125 @@
...
+  /**
+   * @notice Formats an claim fees message
+   * @param _recipient The address of the relayer
+   * @param _amount The amount of fees to claim
+   * @param _transactionIds A group of transaction ids to claim for fee bumps
+   * @return The formatted message
+   */
+  function formatClaimFees(
+    address _recipient,
+    uint256 _amount,
```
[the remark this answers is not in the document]
liu-zhipeng: I agree. we don't need to pass `amount`.

packages/deployments/contracts/contracts_forge/RelayerFeeRouter.t.sol
lmcorbalan: > Also this is a test contract so really it should still be able to pull these events without defining them
I think it isn't possible to emit an event from an external contract, nor to import events from other contracts.

Regarding having an `IRelayerFeeBridge, what would you say is the best place to put it?

packages/deployments/contracts/contracts/nomad-xapps/contracts/relayer-fee-router/RelayerFeeRouter.sol
lmcorbalan: I make sense to have shared things on a `ConnextRouter`

packages/deployments/contracts/contracts/nomad-xapps/contracts/relayer-fee-router/RelayerFeeMessage.sol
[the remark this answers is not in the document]
lmcorbalan: Some thoughts:

- If we don't send the amount and only the group of transaction ids, then it also make sense not to send the relayerFee within the transfer message, we can just keep track of the fees+bumps per transactionID on a mapping in the origin.
With this approach we'll need to allow partial claims, to cover a scenario where there are too many unclaimed transactions for a relayer, iterating through a big number of transactions to calculate the total could end in a DOS scenario for the relayer 

- If we send the amount and don't want to allow partial claims, we can keep track of a unclaimed total + list of transaction per relayer on the remote domain, then when a claim is initialized in Connext we send the unclaimed total to the RelayerFeeRouter plus a list of transaction id for claiming the bumps.

The first approach is simpler but a relayer may end needing to do multiple claims if the number of unclaimed transactions is big, the second adds complexity but would allow to claim the unclaimed total in one transaction + total/partial bumps

Any way we choose we need let the relayer set the list transactions to claim because the unclaimed list of transactions could grow up to the point to make it impossible to claim all at once

packages/deployments/contracts/contracts/nomad-xapps/contracts/relayer-fee-router/RelayerFeeMessage.sol
```diff
@@ -0,0 +1,125 @@
...
+
+  /**
+   * @notice Checks that view is a valid message length
+   * @param _view The bytes string
+   * @return TRUE if message is valid
+   */
+  function isValidClaimFeesLength(bytes29 _view) internal pure returns (bool) {
+    uint256 _len = _view.len();
+    return _len >= 53;
+  }
```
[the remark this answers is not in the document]
lmcorbalan: nice, but it should be both, like 
```solidity
return _len >= 53 && (_len - 53) % 32 = 0;
```

packages/deployments/contracts/contracts_forge/RelayerFeeRouter.t.sol
lmcorbalan: Should I fix this?

packages/deployments/contracts/contracts/nomad-xapps/contracts/relayer-fee-router/RelayerFeeRouter.sol
jakekidd: oh lol

packages/deployments/contracts/contracts/nomad-xapps/contracts/relayer-fee-router/RelayerFeeMessage.sol
```diff
@@ -0,0 +1,122 @@
...
+  // will cause in-flight messages to be mis-interpreted
+  enum Types {
+    Invalid, // 0
+    ClaimFees // 1
+  }
+
+  // ============ Constants ============
+
+  // before: 1 byte identifier + 20 bytes recipient + 32 bytes length + 32 bytes 1 transfer id = 85 bytes
+  uint256 private constant MIN_CLAIM_LEN = 85;
```
LayneHaber, line 26: awesome this is super helpful!

packages/deployments/contracts/contracts_forge/RelayerFeeRouter.t.sol
LayneHaber: yes please!

packages/deployments/contracts/contracts/nomad-xapps/contracts/relayer-fee-router/RelayerFeeRouter.sol
jakekidd: Maybe overkill, but we could also have `ConnextRouter` create/initialize both `RelayerRouter` and `PromiseRouter` as children... like a `RouterManager` lol

LayneHaber: No we shouldn't have them all spawn like that, they should be independent instances and only really need to know about the core `Connext` contract

packages/deployments/contracts/contracts_forge/RelayerFeeRouter.t.sol
LayneHaber: we could do the descriptive name, that works for me!

packages/deployments/contracts/test/connext.spec.ts
```diff
@@ -167,18 +167,22 @@ describe("Connext", () => {
...
+    // TODO - Use real RelayerFeeRouter
+    await originTm.initialize(originDomain, originBridge.address, originTokenRegistry.address, weth.address, ZERO_ADDRESS);
 
     destinationTm = await deployContractWithLibs<Connext>("Connext", {
       AssetLogic: assetLogic.address,
       ConnextUtils: connextUtils.address,
       RouterPermissionsManagerLogic: routerPermissionsManagerLogic.address,
     });
+
+    // TODO - Use real RelayerFeeRouter
```
LayneHaber, line 179: will this happen in another subtask?

LayneHaber (commented):
Couple of things:

- we are likely going to move to a `ConnextRouter` replacing the `BridgeRouter`, so some of the base-class suggestions should probably be included as separate issues/subtasks of that one rather than blocking this branch
- we should have more tests around the message specifically

we can include the message tests/base router contracts in separate PRs (I'll add issues for them now)

packages/deployments/contracts/contracts_forge/RelayerFeeRouter.t.sol
LayneHaber: i'm not sure if this is possible unfortunately -- i believe @patitonar looked into this a bit as well

packages/deployments/contracts/test/connext.spec.ts
patitonar: Yes, this was updated in the integration tests PR https://github.com/connext/nxtp/pull/981

PR #952, opened 2022-04-08
Feature/secrets management

carlomazzaferro:
## The Problem

Secrets management right now lacks a standardised solution. Currently, secrets are shared between team members in chats, sometimes stored in GH Actions, and sometimes even committed to repositories. 

We want a solution that meets the following security best-practices:

1. Centralised storage & retrieval of secrets
2. Access Control Lists to restrict entities to have access to the minimal set of secrets required to perform their duty
3. Secrets encryption

## The Solution

- Implement secrets encryption with [sops](https://github.com/mozilla/sops)
  - Allows for storage of secrets in the nxtp github repository, meeting 1.
  - Secrets are encrypted, meeting 2.
- Ability to decrypt is controlled by a fine-grained IAM roles set up, meeting 3.

## Specifics

`ops/config/<env>.secrets.json` will contain files the contents of which are the encrypted secrets for that environment. Current secrets are mostly related to the routers and sequencer deployment (mnemonics, infura keys, etc.), but any arbitrary secret can be added there. 

Whatever secret that needs to be used by a developer for local development, can be added there, but following the principle of least access, developers should have access _only_ to secrets that are for development, not production ones.

**SOPS**: sops allows with a single command to decrypt secrets and make them available to users. Any yaml, json, text file can be encrypted. It also allows you to securely make the variables available to a temporary shell, so that they can be use for single commands. E.g.

```
sops exec-env ops/env/testnet.secrets.json "echo $mnemonic"
```

Alternatively you can just view the contents of the `secrets.testnet.json` file with:

-  `sops ops/env/testnet/secrets.env` (will open the decrypted file in the default $EDITOR)
-  `sops -d ops/env/secrets.testnet.json` will show the contents of the decrypted file

**IAM & Access Control**: to be able to decrypt the secrets files, your AWS user must be able to access the AWS KMS key that was used to encrypt those files. The current IAM users able to do so for testnet secrets are defined in `ops/modules/kms/locals.tf` 

**LOCAL CREDS SETUP**: simply add the entry in your `~/.aws/credentials` :

```
[some-profile-name]
region=us-east-1
aws_access_key_id = XXXX
aws_secret_access_key = XXX
```
And your `~/.aws/config`
```
[some-profile-name]
region=us-east-1
```

Then, `export AWS_PROFILE=some-profile-name`, and as long as your AWS account is listed in the `locals.tf` file, you should be able to decrypt the secrets file.

## Checklist

- [ ] Run load tests.
- [ ] Update CHANGELOG.md.


rhlsthrm (changes requested):
Looks good, but the secrets names should match the environments we came up with so:
- testnet staging
- testnet prod
- mainnet staging
- mainnet prod

rhlsthrm (approved):

PR #955, opened 2022-04-11
Add relayer fee param in xcall

patitonar:
## The Problem

Subtask of https://github.com/connext/nxtp/issues/900
The user needs to provide the relayer fees in the xcall transaction

## The Solution

Update xcall to receive the relayerFee param and store it for future claim

## Checklist

- [ ] Run load tests.
- [ ] Update CHANGELOG.md.


lmcorbalan (approved):
LGTM

packages/deployments/contracts/contracts/lib/Connext/AssetLogic.sol
```diff
@@ -24,20 +26,22 @@ library AssetLogic {
   function transferAssetToContract(
     address _assetId,
     uint256 _specifiedAmount,
+    uint256 _additionalNativeAmount,
```
LayneHaber, line 29: what this function is doing feels a bit unintuitive -- im wondering if we should rename it or break it up. by the name it presumably just transfers an asset to the contract. instead it seems a catch-all method for transferring to the contract, wrapping native asset, and transferring some additional native asset that doesn't get wrapped.

perhaps it would be better to have a vanilla `transferAssetToContract`, `transferAndWrapNativeAsset`, `transferNativeAsset`

not 100% sure about it though so open to other feedback here

packages/deployments/contracts/contracts_forge/Connext.t.sol
```diff
@@ -12,26 +16,83 @@ import "../contracts/ProposedOwnableUpgradeable.sol";
...
+    address indexed to,
+    IConnext.CallParams params,
+    address transactingAsset,
+    address localAsset,
+    uint256 transactingAmount,
+    uint256 localAmount,
+    uint256 relayerFee,
+    uint256 nonce,
+    address caller
+  );
```
LayneHaber, line 46: will we always have to define our own events in test contracts like this or is there a better way to import them from the source?

packages/deployments/contracts/contracts_forge/Connext.t.sol
```diff
@@ -90,4 +151,156 @@ contract ConnextTest is ForgeHelper {
...
+      0,
+      0,
+      relayerFee,
+      0,
+      address(this)
+    );
+    connext.xcall{value: relayerFee}(args);
+  }
+
+  // Work if relayerFee is set to 0
```
LayneHaber, line 202: we should probably enforce that relayer fee is always greater than 0 while the `amount` can be 0. you will always need to pay someone to relay the tx, but the tx may not include any value (i.e. updating a twap on a destination chain)

packages/deployments/contracts/contracts/Connext.sol
```diff
@@ -422,6 +422,7 @@ contract Connext is
     (address _transactingAssetId, uint256 _amount) = AssetLogic.transferAssetToContract(
       _args.transactingAssetId,
       _args.amount,
+      _args.relayerFee,
```
LayneHaber, line 425: should assert that the relayer fee can't be 0

LayneHaber (changes requested):

packages/deployments/contracts/contracts/lib/Connext/AssetLogic.sol
lmcorbalan: I do agree with the idea of this function name is unintuitive, also I think we could have those 3 functions but I would  leave the logic to decide which on to use in this function rather to move it to the `execute`

packages/deployments/contracts/contracts_forge/Connext.t.sol
patitonar: Unfortunately, it seems there is no way of importing them

packages/deployments/contracts/contracts/lib/Connext/AssetLogic.sol
patitonar: Refactored the method, changed the name and split the logic into other methods https://github.com/connext/nxtp/pull/955/commits/3852d8958fb169c945cf90246c94e0fa58de2051

packages/deployments/contracts/contracts/Connext.sol
patitonar: Updated here https://github.com/connext/nxtp/pull/955/commits/01b71b34666cce4f65299932f3343fa83f9aac18

packages/deployments/contracts/contracts/lib/Connext/AssetLogic.sol
LayneHaber: awesome looks much cleaner!

LayneHaber (approved):

PR #956, opened 2022-04-11
Add InitiateClaim and Claim functionalities to Connext

lmcorbalan:
## The Problem

Subtask of #900 
The relayers need a way to initiate the claim for it fees in the destination domain which should be fulfilled in the origin domain by the RelayerFeeRouter

## The Solution

Add `initiateClaim` and `claim` functionalities to Connext

## Checklist

- [ ] Run load tests.
- [ ] Update CHANGELOG.md.

### DIFF
--- a/packages/deployments/contracts/contracts/Connext.sol
+++ b/packages/deployments/contracts/contracts/Connext.sol
@@ -65,7 +65,6 @@ contract Connext is
   error Connext__handleRelayerFees_notRtrSig();
   error Connext__setMaxRoutersPerTransfer_invalidMaxRoutersPerTransfer();
   error Connext__onlyRelayerFeeRouter_notRelayerFeeRouter();
-  error Connext__initiateClaim_notRelayer(bytes32 transferId);
 
   // ============ Constants =============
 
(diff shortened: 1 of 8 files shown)


packages/deployments/contracts/contracts/Connext.sol
```diff
@@ -193,6 +222,7 @@ contract Connext is
     nonce = 0;
     domain = _domain;
     bridgeRouter = BridgeRouter(_bridgeRouter);
+    relayerFeeRouter = RelayerFeeRouter(_relayerFeeRouter);
```
LayneHaber, line 225: do you guys think we should have interfaces for this instead?

LayneHaber:
LGTM! just curious about your thoughts on using interfaces. could just be adding more boiler plate though

packages/deployments/contracts/contracts/lib/Connext/ConnextUtils.sol
```diff
@@ -158,4 +164,61 @@ library ConnextUtils {
...
+   * @param _relayerFeeRouter - The local nomad relayer fee router
+   * @param _transferRelayer - Mapping of transactionIds to relayer
+   */
+  function initiateClaim(
+    uint32 _domain,
+    address _recipient,
+    bytes32[] calldata _transferIds,
+    RelayerFeeRouter _relayerFeeRouter,
+    mapping(bytes32 => address) storage _transferRelayer
+  ) external {
```
LayneHaber, line 183: we should require the length of the transferIds is greater than 1 since thats included in the valid message length [here](https://github.com/connext/nxtp/blob/b71c852657fe62f24fa6e91f1dc050c0a177f9cc/packages/deployments/contracts/contracts/nomad-xapps/contracts/relayer-fee-router/RelayerFeeMessage.sol#L96-L100) and [here](https://github.com/connext/nxtp/blob/b71c852657fe62f24fa6e91f1dc050c0a177f9cc/packages/deployments/contracts/contracts/nomad-xapps/contracts/relayer-fee-router/RelayerFeeMessage.sol#L26)

LayneHaber (changes requested):

packages/deployments/contracts/contracts/lib/Connext/ConnextUtils.sol
LayneHaber: disregard included in the relayer fee router!

LayneHaber (approved):

patitonar (approved):

PR #959, opened 2022-04-11
Save relayer who execute a transferId

lmcorbalan:
## The Problem

Subtask of #900 
A way to store who executed a given transactionId on the destination domain

## The Solution
Update the mapping `transferRelayer` with the address executing a each `transactionId`

## Checklist

- [ ] Run load tests.
- [ ] Update CHANGELOG.md.


packages/deployments/contracts/test/connext.spec.ts
```diff
@@ -561,36 +561,36 @@ describe("Connext", () => {
...
-
-      const beforeBalance = await user.getBalance();
-      const removeTx = await originTm.connect(router).removeRelayerFees(parseEther("0.5"), user.address);
-      await removeTx.wait();
-
-      expect(await originTm.routerRelayerFees(router.address)).to.be.eq(beforeRouterFee.add(parseEther("0.5")));
-      expect(await user.getBalance()).to.be.eq(beforeBalance.add(parseEther("0.5")));
-    });
-  });
+  // describe("addRelayerFees", () => {
```
LayneHaber, line 564: we can just delete all of these tests instead of commenting them out

packages/deployments/contracts/test/connext.spec.ts
```diff
@@ -896,12 +896,19 @@ describe("Connext", () => {
       nonce,
       local: local.address,
       amount: routerAmount,
-      feePercentage: constants.Zero,
-      relayerSignature: "0x",
+      // feePercentage: constants.Zero,
```
LayneHaber, line 899: delete

packages/deployments/contracts/test/connext.spec.ts
```diff
@@ -993,12 +1000,19 @@ describe("Connext", () => {
       nonce,
       local: local.address,
       amount: routerAmount,
-      relayerSignature: "0x",
+      // relayerSignature: "0x",
       routers: [router.address],
       originSender: user.address,
-      feePercentage: constants.Zero,
+      // feePercentage: constants.Zero,
```
LayneHaber, line 1006: dlt

packages/deployments/contracts/test/connext.spec.ts
```diff
@@ -1164,17 +1178,26 @@ describe("Connext", () => {
...
 
-        await destinationTm.connect(router).execute({
+        const execute = await destinationTm.connect(router).execute({
           params,
           nonce,
           local: local.address,
           amount: routersAmount,
-          feePercentage: constants.Zero,
-          relayerSignature: "0x",
+          // feePercentage: constants.Zero,
```
LayneHaber, line 1186: deeeeleet

packages/deployments/contracts/test/connext.spec.ts
```diff
@@ -1228,8 +1251,8 @@ describe("Connext", () => {
...
```
LayneHaber, line 1254: duhleat

packages/deployments/contracts/test/connext.spec.ts
```diff
@@ -1244,8 +1267,8 @@ describe("Connext", () => {
         nonce,
         local: local.address,
         amount: routersAmount,
-        feePercentage: constants.Zero,
-        relayerSignature: "0x",
+        // feePercentage: constants.Zero,
```
LayneHaber, line 1270: d-leet

packages/deployments/contracts/test/connext.spec.ts
```diff
@@ -1265,8 +1288,8 @@ describe("Connext", () => {
           nonce,
           local: local.address,
           amount: routersAmount,
-          feePercentage: constants.Zero,
-          relayerSignature: "0x",
+          // feePercentage: constants.Zero,
```
LayneHaber, line 1291: duhlete

LayneHaber (changes requested):
just need to add a bit more red to the PR :)

packages/deployments/contracts/contracts/Connext.sol
```diff
@@ -581,10 +551,8 @@ contract Connext is
       });
     }
 
-    // Pay metatx relayer
-    // NOTE: if this is done *without* fast liquidity, router will be address(0) and the relayer
-    // will always be paid
-    _handleRelayerFees(_transferId, _args.routers, _args.feePercentage, _args.relayerSignature);
+    // Set the relayer for this transaction to allow for future claim
+    transferRelayer[_transferId] = msg.sender;
```
jakekidd, line 555: `_handleRelayerFees` previously checked the router's `relayerSignature`... removing this means that the relayer can now call this execute fn using any router. There's no longer any security behind execute except the whitelist with the relayers.

While we aren't in prod yet so this isn't a real concern, I'm wondering if we should go ahead and knock this essential piece of tech-debt out of the way among the relayer fee changes here.

The essential piece I'm referring to: having a *different* signature from the router that makes this tx securely metatx-able. The signature digest should be something like `[transferId, router fee %, auction round number]`

We can write the check in a sep. branch, but personally I think it's worth considering it a blocker for getting this in. I'd *like* to proceed with that security mechanism in place lol.

jakekidd, line 555: Where is `transferRelayer` defined?

lmcorbalan: It is defined in the same Connext contract [here](https://github.com/connext/nxtp/blob/relayer-fee-execute/packages/deployments/contracts/contracts/Connext.sol#L173
) it was introduced with 29bc25ed154e7827352c82479d8877253e830c5d

packages/deployments/contracts/contracts/Connext.sol
```diff
@@ -574,6 +544,11 @@ contract Connext is
    * @return bytes32 The transfer id of the crosschain transfer
    */
   function execute(ExecuteArgs calldata _args) external override returns (bytes32) {
+    // If the sender is not approved relayer, revert()
+    if (!approvedRelayers[msg.sender]) {
+      revert Connext__execute_notApprovedRelayer();
+    }
+
```
liu-zhipeng, line 551: Ok

LayneHaber (approved):

PR #960, opened 2022-04-11
Poc connext router

LayneHaber:
## The Problem

- Transferring over the bridge is very expensive
- The `BridgeRouter` adds complexity, when what we really need is the `TokenRegistry`

## The Solution

This PR is a POC that explores what it would look like to take full ownership of the `BridgeRouter` / `BridgeMessage` piece and put the current `Connext.sol` logic into a `ConnextHandler`. This approach reduces gas by ~100K for the full flow, and lowers the complexity by:

- Not using `BridgeRouter.sol` / `BridgeMessage.sol`
- Moving core functionality from `Connext.sol` into a `ConnextHandler.sol` that combines the bridge and connext pieces

## Checklist

- [ ] Run load tests.
- [ ] Update CHANGELOG.md.


patitonar:
We like how this approach looks, reduces contract interactions and gas usage. The only concern we see is that now the current ConnextHandler size is ~28MB and there are some features that are not yet included (multipath, relayer fees, + the future ones). So we will need to look for new ways to reduce the size of the contract by applying similar strategies as we did before

LayneHaber:
> We like how this approach looks, reduces contract interactions and gas usage. The only concern we see is that now the current ConnextHandler size is ~28MB and there are some features that are not yet included (multipath, relayer fees, + the future ones). So we will need to look for new ways to reduce the size of the contract by applying similar strategies as we did before

Yes, but none of the logic is moved into any libraries. Hopefully it will not be incredibly painful

packages/deployments/contracts/contracts/Connext.sol
```diff
@@ -1,787 +0,0 @@
-// SPDX-License-Identifier: UNLICENSED
```
jakekidd, line 1: Why does it look like this entire file `Connext.sol` is being deleted? lol

I can't seem to find where it's getting replaced (if it's getting replaced)

jakekidd: oh nvm, I see, it's `ConnextHandler` now

packages/deployments/contracts/contracts/nomad-xapps/contracts/connext/ConnextHandler.sol
```diff
@@ -0,0 +1,524 @@
...
+    address payable _to
+  ) external override nonReentrant {
+    // transfer to specicfied recipient IF recipient not set
+    address recipient = routerRecipients(msg.sender);
+    recipient = recipient == address(0) ? _to : recipient;
+
+    ConnextUtils.removeLiquidity(_amount, _local, recipient, routerBalances, wrapper);
+  }
+
+  function xcall(XCallArgs calldata _args) external payable override returns (bytes32) {
```
jakekidd, line 381: where is the function descriptor / docstring? :(

packages/deployments/contracts/contracts/nomad-xapps/contracts/connext/ConnextHandler.sol
```diff
@@ -0,0 +1,524 @@
...
+  /**
+   * @notice Handles an incoming message
+   * @dev This function relies on nomad relayers and should not consume arbitrary amounts of
+   * gas
+   * @param _origin The origin domain
+   * @param _nonce The unique identifier for the message from origin to destination
+   * @param _sender The sender address
+   * @param _message The message
+   */
+  // ROUTER
```
jakekidd, line 411: nit: Can we get this to be formatted like the other place-markers?

Like:
```
// ============ Nomad Router ============
```
or
```
// ============ Nomad Handler ============
```
whatever yall think

packages/deployments/contracts/contracts/lib/Connext/AssetLogic.sol
```diff
@@ -14,38 +14,64 @@ library AssetLogic {
...
    * actual amount transferred (i.e. fee on transfer tokens)
+   * @param _fee - The fee amount in native asset included as part of the transaction that
+   * should not be considered for the transfer amount.
    * @param _wrapper - The address of the wrapper for the native asset on this domain
    * @return The assetId of the transferred asset
    * @return The amount of the asset that was seen by the contract (may not be the specifiedAmount
    * if the token is a fee-on-transfer token)
    */
-  function transferAssetToContract(
+  function handleIncomingAsset(
```
jakekidd, line 26: where is this being used?

packages/deployments/contracts/contracts/lib/Connext/AssetLogic.sol
```diff
@@ -14,38 +14,64 @@ library AssetLogic {
...
-    uint256 _specifiedAmount,
+    uint256 _assetAmount,
+    uint256 _fee,
     IWrapped _wrapper
   ) external returns (address, uint256) {
-    uint256 trueAmount = _specifiedAmount;
+    uint256 trueAmount = _assetAmount;
 
     if (_assetId == address(0)) {
+      if (msg.value != _assetAmount + _fee) revert AssetLogic__transferAssetToContract_notAmount();
```
jakekidd, line 35: nit: should we change the names of this custom error ..?

packages/deployments/contracts/contracts/lib/Connext/ConnextUtils.sol
```diff
@@ -158,4 +735,366 @@ library ConnextUtils {
...
+   * @param _assetAmount - The specified amount to transfer. May not be the
+   * actual amount transferred (i.e. fee on transfer tokens)
+   * @param _fee - The fee amount in native asset included as part of the transaction that
+   * should not be considered for the transfer amount.
+   * @param _wrapper - The address of the wrapper for the native asset on this domain
+   * @return The assetId of the transferred asset
+   * @return The amount of the asset that was seen by the contract (may not be the specifiedAmount
+   * if the token is a fee-on-transfer token)
+   */
+  function _handleIncomingAsset(
```
jakekidd, line 1023: 2 wrapper functions called `_handleIncomingAsset` ? And neither of them use `AssetLogic.handleIncomingAsset` ?

packages/deployments/contracts/contracts/interfaces/IConnext.sol
```diff
@@ -1,7 +1,7 @@
-import "../nomad-xapps/contracts/bridge/BridgeMessage.sol";
-
 // SPDX-License-Identifier: UNLICENSED
-pragma solidity 0.8.11;
+pragma solidity ^0.8.11;
+
+import "../nomad-xapps/contracts/connext/ConnextMessage.sol";
 
 interface IConnext {
```
jakekidd, line 6: nit: rename to `IConnextHandler` ?

packages/deployments/contracts/contracts/lib/Connext/ConnextUtils.sol
```diff
@@ -3,52 +3,326 @@ pragma solidity 0.8.11;
...
-    address _local,
-    address _to,
-    uint256 _amount
-  ) external pure returns (bytes32) {
-    IConnext.ReconciledTransfer memory transfer = IConnext.ReconciledTransfer({
-      local: _local,
-      amount: _amount,
-      to: _to
-    });
+  function addAssetId(
```
jakekidd, line 247: These functions here (admin functions) have no permissioning / permissions modifiers. Is that okay for now?

packages/deployments/contracts/contracts/lib/Connext/ConnextUtils.sol
```diff
@@ -158,4 +735,366 @@ library ConnextUtils {
...
+    if (_args.xCallArgs.params.to == address(0)) {
+      revert ConnextUtils__xcall_emptyTo();
+    }
+  }
+
+  /**
+   * @notice Processes an `xcall`
+   * @dev Need this to prevent stack too deep
+   */
+  function _processXcall(
```
jakekidd, line 759: nit: can we call this `_xcallProcess` ? Would read better with `_xcallSanityChecks`... and would read better if we have to do this for other methods in the future (e.g. `_reconcileProcess`)

sorry

packages/deployments/contracts/contracts/lib/Connext/ConnextUtils.sol
```diff
@@ -158,4 +735,366 @@ library ConnextUtils {
...
+        transactingAssetId: transactingAssetId,
+        amount: amount,
+        bridgedAmt: bridgedAmt,
+        bridged: bridged
+      })
+    );
+  }
+
+  /**
+   * @notice Calculates a transferId base son `execute` arguments
```
jakekidd, line 812: Hello, have you met my `base son`

jakekidd: never talk to me or my `base son` ever again

packages/deployments/contracts/contracts/lib/Connext/ConnextUtils.sol
```diff
@@ -158,4 +735,366 @@ library ConnextUtils {
...
+          _args.executeArgs.originSender,
+          tokenId,
+          tokenDomain,
+          _args.executeArgs.amount
+        )
+      );
+  }
+
+  /**
+   * @notice Calculates a transferId base son `xcall` arguments
```
jakekidd, line 832: never talk to me or my base son ever again

jakekidd: you've heard of JSON, now try base son

packages/deployments/contracts/contracts/lib/Connext/ConnextUtils.sol
```diff
@@ -158,4 +735,366 @@ library ConnextUtils {
...
+    (uint32 domain, bytes32 id) = _args.tokenRegistry.getTokenId(_asset);
+
+    // format token id
+    bytes29 tokenId = ConnextMessage.formatTokenId(domain, id);
+
+    // send message
+    return ConnextMessage.formatMessage(tokenId, action);
+  }
+
+  function _handleExecuteTransaction(
```
jakekidd, line 902: add a little function docstring pls, just with the `@notice base son` and `@dev Need this to prevent stack too deep` - that's actually an important clarifier imo

packages/deployments/contracts/contracts/lib/Connext/ConnextUtils.sol
```diff
@@ -61,27 +335,325 @@ library ConnextUtils {
...
+    mapping(bytes32 => bool) storage _reconciledTransfers,
+    mapping(address => mapping(address => uint256)) storage _routerBalances,
+    mapping(bytes32 => IStableSwap) storage _adoptedToLocalPools,
+    mapping(bytes32 => address) storage canonicalToAdopted,
+    RouterPermissionsManagerInfo storage _routerInfo,
+    mapping(bytes32 => address) storage _transferRelayer
+  ) external returns (bytes32 transferId) {
+    (bytes32 transferId, bool reconciled) = _executeSanityChecks(_args, _transferRelayer, _reconciledTransfers);
+
+    // check to see if the transfer was reconciled
```
jakekidd, line 495: nit: inaccurate comment lol, should go above sanity checks if anything

packages/deployments/contracts/contracts/lib/Connext/ConnextUtils.sol
```diff
@@ -61,27 +335,325 @@ library ConnextUtils {
...
+    (uint256 amount, address adopted) = _handleExecuteLiquidity(
```
jakekidd, line 496: Note: the only thing this method does if the transfer is `reconciled` (meaning if the transfer is not a fast-liq transfer) is literally just call `_swapFromLocalAssetIfNeeded`

We may want to optimize this by removing the `isFast` boolean input to this function and literally deciding whether or not to call it here, as the consumer

And then calling `_swapFromLocalAssetIfNeeded` separately, using the result.

As in:
```solidity
if (reconciled) {
  (uint256 amount, address adopted) = _swapFromLocalAssetIfNeeded(...);
} else {
  (uint256 amount, address adopted) = _handleExecuteLiquidity(
        transferId,
        // no `!reconciled` argument!
      _args,
      _routedTransfers,
      _routerBalances,
      _adoptedToLocalPools,
      canonicalToAdopted,
      _routerInfo
    );
```

Would save us an argument in general and some gas in the case that this is a slow liq transfer

packages/deployments/contracts/contracts/lib/Connext/ConnextUtils.sol
```diff
@@ -61,27 +335,325 @@ library ConnextUtils {
...
+      _args,
+      _routedTransfers,
+      _routerBalances,
+      _adoptedToLocalPools,
+      canonicalToAdopted,
+      _routerInfo
+    );
+
+    // execute the transaction
+    _handleExecuteTransaction(_args, amount, adopted, transferId, reconciled);
```
jakekidd, line 508: in keeping up with the naming convention for extending functions (functions we're using to avoid stack too deep errors), maybe we could change `_handleExecuteTransaction` to `_executeProcess` ?

packages/deployments/contracts/contracts/lib/Connext/ConnextUtils.sol
```diff
@@ -158,4 +735,366 @@ library ConnextUtils {
...
+      for (uint256 i; i < pathLen; ) {
+        // while theres no way for a router to have sufficient liquidity
+        // if they have never been approved, this check ensures they weren't
+        // removed from the whitelist
+        if (!_routerInfo.approvedRouters[_args.executeArgs.routers[i]]) {
+          revert ConnextUtils__execute_notSupportedRouter();
+        }
+
+        // decrement routers liquidity
+        _routerBalances[_args.executeArgs.routers[i]][_args.executeArgs.local] -= routerAmount;
```
jakekidd, line 975: What happens if the router doesn't have enough liquidity? Are we checking that?

jakekidd: I guess it would just revert, it is a uint after all

packages/deployments/contracts/contracts/nomad-xapps/contracts/relayer-fee-router/RelayerFeeMessage.sol
```diff
@@ -0,0 +1,118 @@
...
+  /**
+   * @notice Parse The group of transfers ids to claim for fee bumps
+   * @param _view The message
+   * @return The group of transfers ids to claim for fee bumps
+   */
+  function transferIds(bytes29 _view) internal pure typeAssert(_view, Types.ClaimFees) returns (bytes32[] memory) {
+    uint256 length = _view.indexUint(LENGTH_ID_START, LENGTH_ID_LEN);
+
+    bytes32[] memory ids = new bytes32[](length);
+    for (uint256 i = 0; i < length; i++) {
```
jakekidd, line 81: Can do `unchecked i`

packages/deployments/contracts/contracts/nomad-xapps/contracts/relayer-fee-router/RelayerFeeMessage.sol
```diff
@@ -0,0 +1,118 @@
...
+      ids[i] = _view.index(TRANSFER_IDS_START + i * TRANSFER_ID_LEN, TRANSFER_ID_LEN);
```
jakekidd, line 82: I *believe* we can do this message division math `unchecked`... yes?

packages/deployments/contracts/contracts/nomad-xapps/contracts/connext/ConnextHandler.sol
```diff
@@ -0,0 +1,530 @@
...
+  function _addLiquidityForRouter(
+    uint256 _amount,
+    address _local,
+    address _router
+  ) internal {
+    // Sanity check: router is sensible
+    if (_router == address(0)) revert ConnextHandler__addLiquidityForRouter_routerEmpty();
+
+    // Sanity check: nonzero amounts
+    if (_amount == 0) revert ConnextHandler__addLiquidityForRouter_amountIsZero();
```
jakekidd, line 516: Note that amount will be 0 if the router is adding liquidity for native assets... no?

jakekidd: ik we dont support that yet, so i figured maybe that's why this is here. but just double checking

packages/deployments/contracts/contracts/lib/Connext/ConnextUtils.sol
```diff
@@ -61,27 +335,325 @@ library ConnextUtils {
...
+    mapping(bytes32 => address) storage _transferRelayer,
+    mapping(bytes32 => bool) storage _reconciledTransfers
+  ) private returns (bytes32, bool) {
+    if (_args.executeArgs.routers.length > _args.maxRoutersPerTransfer)
+      revert ConnextUtils__execute_maxRoutersExceeded();
+
+    bytes32 transferId = _getTransferId(_args);
+
+    // require this transfer has not already been executed
+    if (_transferRelayer[transferId] != address(0)) {
```
jakekidd, line 635: shouldn't this be `_transferRelayer[transferId] == address(0)` ?

packages/deployments/contracts/contracts/lib/Connext/ConnextUtils.sol
```diff
@@ -3,52 +3,326 @@ pragma solidity 0.8.11;
...
+  }
+
+  struct ExecuteLibArgs {
+    IConnext.ExecuteArgs executeArgs;
+    uint256 maxRoutersPerTransfer;
+    ITokenRegistry tokenRegistry;
+    IWrapped wrapper;
+    IExecutor executor;
+    uint256 LIQUIDITY_FEE_NUMERATOR;
+    uint256 LIQUIDITY_FEE_DENOMINATOR;
```
jakekidd, line 82: nit: instead of `CONSTANT_CASE`, use `camelCase`  for args?

packages/deployments/contracts/contracts/lib/Connext/ConnextUtils.sol
```diff
@@ -61,27 +335,325 @@ library ConnextUtils {
...
+  }
+
+  /**
+   * @notice Called on the destination domain to disburse correct assets to end recipient
+   * and execute any included calldata
+   * @dev Can be called prior to or after `handle`, depending if fast liquidity is being
+   * used.
+   */
+  function execute(
+    ExecuteLibArgs calldata _args,
```
jakekidd, line 484: we require that the user pass in the `LIQUIDITY_FEE_NUMERATOR` and `LIQUIDITY_FEE_DENOMINATOR`, but it looks like we ignore these args in favor of a local constant... maybe we should separate those arguments out for now? Just pass them in directly using the local constants to `_handleExecuteLiquidity => _getFastTransferAmount`

packages/deployments/contracts/contracts/lib/Connext/ConnextUtils.sol
```diff
@@ -3,52 +3,326 @@ pragma solidity 0.8.11;
...
+   */
+  event LiquidityRemoved(address indexed router, address to, address local, uint256 amount, address caller);
+
+  /**
+   * @notice Emitted when `xcall` is called on the origin domain
+   */
+  event XCalled(
+    bytes32 indexed transferId,
+    IConnext.XCallArgs xcallArgs,
+    xCalledEventArgs args,
```
jakekidd: aren't these event args a bit redundant here?

`XCallArgs` includes `transactingAssetId` and `amount` already.

`XCalledArgs` seems to have only 2 unique vars defined `bridgedAmt` and `bridged`... what are those vars referring to exactly?

jakekidd: To illustrate, here's everything that's being emitted in this event:
```
transferId, xcallArgs(params(to, callData, originDomain, destinationDomain), transactingAssetId, amount, relayerFee), args(transactingAssetId, amount, bridgedAmt, bridged), nonce, message, caller
```

(notice `transactingAssetId` and `amount` appear twice)

packages/deployments/contracts/contracts/lib/Connext/AssetLogic.sol
```diff
@@ -31,7 +31,7 @@ library AssetLogic {
     uint256 _assetAmount,
     uint256 _fee,
     IWrapped _wrapper
-  ) external returns (address, uint256) {
+  ) internal returns (address, uint256) {
```
lmcorbalan, line 34: I had eliminated AssetLogic (https://github.com/connext/nxtp/pull/1015/commits/0d7dd679017fbbf275f561cc006946bcb9a001e3) and moved is content to ConnextUtils to avoid doing double delegated calls, to be honest I like much more having this separated AssetLogic but we meed all its functions to be `internal` make this an embed library instead of an linked one

packages/deployments/contracts/contracts/lib/Connext/ConnextUtils.sol
lmcorbalan: The `onlyOwner` modifier is in `ConnextHandler.addAssetId` which calls this function

packages/deployments/contracts/contracts/lib/Connext/ConnextUtils.sol
lmcorbalan: :joy:

packages/deployments/contracts/contracts/lib/Connext/ConnextUtils.sol
lmcorbalan: no, `_transferRelayer[transferId] == address(0)` means it hasn't been executed thus it doesn't rever

packages/deployments/contracts/contracts/lib/Connext/ConnextUtils.sol
lmcorbalan: Not sure if I understand this one.
The `execute` on ConnextLogic (former ConextUtils) is only used by ConnextHandler, not by the user

packages/deployments/contracts/contracts/lib/Connext/ConnextUtils.sol
lmcorbalan: Maybe the names are redundant? but the values could be different.
e.g.
- xcallArgs. transactingAssetId could be `address(0)` so args.transactingAssetId would be `address(wrapper)` (https://github.com/connext/nxtp/blob/poc-connext-router/packages/deployments/contracts/contracts/lib/Connext/ConnextLogic.sol#L720-L722)
- amount could be different if the token being transfer is a deflationary ERC20 (https://github.com/connext/nxtp/blob/poc-connext-router/packages/deployments/contracts/contracts/lib/Connext/AssetLogic.sol#L75)

packages/deployments/contracts/contracts/nomad-xapps/contracts/relayer-fee-router/RelayerFeeMessage.sol
lmcorbalan: 🤔 not sure if I understand this one

packages/deployments/contracts/contracts/nomad-xapps/contracts/connext/ConnextHandler.sol
LayneHaber: Amount will never be 0, even is using native assets. It should always be the amount you add, and is asserted in the `AssetLogic` library

packages/deployments/contracts/contracts/lib/Connext/ConnextUtils.sol
LayneHaber: theres some renaming that will happen in a separate PR. this will really only be relevant for `xcall`, `execute`, and `reconcile` -- basically im thinking:

- the `ConnextHandler` has the top level functions
- the `ConnextLogic` has the same top level function names
- if there are smaller functions within the `ConnextLogic` function, it should be named `_process${TOP_LEVEL_NAME}${DESCRIPTOR_OF_PROCESS}`. could also be named `_handle{...}` with the same structure

we can do this in other PRs

packages/deployments/contracts/contracts/nomad-xapps/contracts/connext/ConnextHandler.sol
LayneHaber: can add in later in cleanup PRs

lmcorbalan: added

packages/deployments/contracts/contracts/nomad-xapps/contracts/connext/ConnextHandler.sol
lmcorbalan: deleted

packages/deployments/contracts/contracts/lib/Connext/ConnextUtils.sol
jakekidd: looks good! yeah layne explained these are indeed not redundant. we'll have to fix up names later

packages/deployments/contracts/contracts/lib/Connext/ConnextUtils.sol
jakekidd: Yeah this was a misinterpretation on my part, which is why i resolved it haha. Ty for answer tho

PR #969, opened 2022-04-12
[contracts/router] Router permits

### LINKED ISSUE #968: [contracts] Router metatx signature
Since we'll be eliminating the relayer fee signature in #900, there will no longer be any security measures preventing the sequencer/relayer from sending whatever execute tx params to chain that they like. Obviously this isn't of any major concern operating on testnets, but would be good to get this critical piece of tech debt out of the way (or at the very least, leave some room for the final implementation in the future).

In order to be certain that the router has consented to executing a transfer with the amount specified, the router who is LPing must sign a hash of the ExecuteArgs. We'll need to pass in this signature, and recover it on chain to be sure it came from the router.

The signature must include the router array length (indicating what level of multipath transfer the router consents to) and include the transfer ID. Here is the draft of the signed content:
```solidity
  /**
   * @notice The permit represents the router's consent to use its funds to `execute` a transfer. This
   * information should be hashed and signed by each router in the routers array individually.
   * @param params - The CallParams. These are consistent across sending and receiving chains.
   * @param local - The local asset for the transfer, will be swapped to the adopted asset if
   * appropriate
   * @param round - The length of the routers array. Indicates how many routers this router assents to
   * conducting a multipath transfer with.
   * @param originSender - The sender of the transfer on the origin chain (i.e. the user, but not
   * necessarily the xcaller).
   */
  struct RouterPermit {
    bytes transferId;
    CallParams params;
    address local;
    uint16 round;
    address originSender;
  }
  ```

### PR DESCRIPTION
**Pull request by jakekidd**
Fixes #968 

- Added a `RouterPermit` struct.
- Router signs the content of that struct.
- On-chain comparison of signatures for each router in the `routers` array.

### CODE REVIEW

File: packages/deployments/contracts/contracts/interfaces/IConnext.sol [hunk 1]
```diff
@@ -91,17 +91,36 @@ interface IConnext {
...
+   * @param local - The local asset for the transfer, will be swapped to the adopted asset if
+   * appropriate
+   * @param quorum - The length of the routers array. Indicates how many routers this router assents to
+   * conducting a multipath transfer with.
+   */
+  struct RouterPermit {
+    bytes transferId;
+    CallParams params;
+    address local;
+    uint16 quorum;
```
LayneHaber, line 122: we shouldn't call this a quorum -- its the number of routers in the path (they all have to support it, not just a subset). maybe just call it `pathLength`

packages/deployments/contracts/contracts/lib/Connext/ConnextUtils.sol
```diff
@@ -49,13 +49,38 @@ library ConnextUtils {
...
+  function verifyRouterPermit(
+    bytes32 _permit,
+    address[] calldata _routers,
+    bytes[] calldata _signatures
+  ) external view {
+    for (uint i = 0; i < _signatures.length; i++) {
+      // TODO: If the sender *is* the router, skip
+      if (msg.sender == _routers[i]) {
+        continue;
+      }
```
LayneHaber, line 61: need to check this against the relayer checks that happen in execute as well (though really the router shouldn't be sending this in the majority of cases

packages/deployments/contracts/contracts/lib/Connext/ConnextUtils.sol
```diff
@@ -49,13 +49,38 @@ library ConnextUtils {
     return keccak256(abi.encode(transfer));
   }
 
+  function verifyRouterPermit(
+    bytes32 _permit,
+    address[] calldata _routers,
+    bytes[] calldata _signatures
+  ) external view {
+    for (uint i = 0; i < _signatures.length; i++) {
+      // TODO: If the sender *is* the router, skip
```
LayneHaber, line 58: remove `TODO`, just leave the comment

packages/deployments/contracts/contracts/lib/Connext/ConnextUtils.sol
```diff
@@ -49,13 +49,38 @@ library ConnextUtils {
...
+      require(recoverHashedSignature(_permit, _signatures[i]) == _routers[i], "Invalid signature");
+    }
+  }
+
+  /**
+   * @notice Holds the logic to recover the signer from a hashed encoded payload.
+   * @param _hashed - The bytes32 payload that was signed
+   * @param _sig - The signature you are recovering the signer from
+   */
+  function recoverHashedSignature(bytes32 _hashed, bytes calldata _sig) public pure returns (address) {
```
LayneHaber, line 72: why not pass in the encoded information instead of the hash of it and always use the same recovery method? we should only have one of these low-level recovery functions

packages/deployments/contracts/contracts/interfaces/IConnext.sol
LayneHaber: also since were not using this anywhere we should really delete it

packages/deployments/contracts/contracts/lib/Connext/ConnextUtils.sol
jakekidd: Yeah I agree. However, I was trying to do the gas-efficient thing of recycling calldata, and reduce the number of `abi.encode` and `keccak256` calls here, since we might be looping over multiple router sigs. I'll see what I can do to reduce this to one function

packages/deployments/contracts/contracts/interfaces/IConnext.sol
jakekidd: Yep, will do. Was initially using it lol

packages/deployments/contracts/contracts/lib/Connext/ConnextUtils.sol
jakekidd: wym? This is contingent on there only being one signature/verification method, i.e. the relayer signature isn't checked.

jakekidd:
Close to completion here. A few things needed:
- [ ] Router needs to actually produce the permit signature in the off-chain code, should use a utility for this.
- [ ] Need the new relayer fee design to be implemented/merged before this
- [ ] Need to resolve gas optimization issue around using a sep. `recoverHashedSignatures`

jakekidd:
Closing - this got done in the poc-connext-router branch

PR #991, opened 2022-04-15
Fix sanity checks

### LINKED ISSUE #990: [router][sequencer] Fix sanity checks
The current sanity checks are not implemented correctly. If the sanity check is a simple estimateGas call, it can fail in the multipath case.

This is how it should work:

For router, before execute:
- Check liquidity
- Send bid

For sequencer:
- Save bid without checking anything
- estimateGas only before sending to relayer

rhlsthrm:
## The Problem

Fixes #990 

## The Solution

Cleans up sanity checks.

## Checklist

- [ ] Run load tests.
- [ ] Update CHANGELOG.md.


packages/agents/router/src/lib/operations/execute.ts
```diff
@@ -92,15 +84,16 @@ export const execute = async (params: XTransfer): Promise<void> => {
...
-      bid,
-    });
+  // sanity check
+  const bal = await subgraph.getAssetBalance(destinationDomain, routerAddress, executeLocalAsset);
+  logger.info("Checking balance", requestContext, methodContext, { bal: bal.toString() });
+  if (bal.lt(receivingAmount)) {
+    // throw new NotEnoughAmount({
+    //   bal: bal.toString(),
+    //   receivingAmount: receivingAmount.toString(),
+    // });
```
jakekidd, line 94: Why is this commented out? Not functional yet?

packages/agents/sequencer/src/lib/operations/auctions.ts
```diff
@@ -131,81 +100,112 @@ export const selectBids = async (_requestContext: RequestContext) => {
...
 
-  for (const transferId of Object.keys(auctions)) {
-    const { bids, origin, destination } = auctions[transferId];
-
-    // TODO: deprecate... necessary for now
-    const bidData = await cache.auctions.getBidData(transferId);
-    if (!bidData) {
-      logger.error("Bid data not found for transfer!", requestContext, methodContext, undefined, {
-        transferId,
+  await Promise.all(
```
jakekidd, line 111: concerned about doing a promise.all here... eventually, we need to account for the fact that one transfer's auction can affect another.

A router might have $100, and bid on 2 $100 transfers. If we execute both of them simultaneously, it might seem like they pass at first... but only one can go through.

So we have to do a for loop, at the very least by domain.

Should update our local tracking of routers' funds in the cache or something when we spend a router's liq on a transfer.

PR #994, opened 2022-04-16
fix local asset issues: wip

jakekidd:


rhlsthrm:
Fixed by this: https://github.com/connext/nxtp/pull/995

PR #1012, opened 2022-04-20
Dedup the bids

wanglonghong:
## Checklist

- [ ] Run load tests.
- [ ] Update CHANGELOG.md.

### DIFF
--- a/packages/adapters/cache/src/lib/caches/auctions.ts
+++ b/packages/adapters/cache/src/lib/caches/auctions.ts
@@ -131,7 +131,11 @@ export class AuctionsCache extends Cache {
     let keys: string[] = [];
     await new Promise((res) => {
       stream.on("data", (resultKeys: string[] = []) => {
-        keys = keys.concat(resultKeys);
+        // Note that resultKeys will sometimes contain duplicates due to SCAN's implementation in Redis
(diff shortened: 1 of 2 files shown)


packages/adapters/cache/src/lib/caches/auctions.ts
```diff
@@ -131,7 +131,11 @@ export class AuctionsCache extends Cache {
     let keys: string[] = [];
     await new Promise((res) => {
       stream.on("data", (resultKeys: string[] = []) => {
-        keys = keys.concat(resultKeys);
+        // Note that resultKeys will sometimes contain duplicates due to SCAN's implementation in Redis
+        // link : https://redis.io/commands/scan/#scan-guarantees
+        for (const resultKey of resultKeys) {
```
rhlsthrm, line 136: We need to dedup the bids inserting INTO the cache, not reading from the cache.

rhlsthrm (changes requested):

packages/adapters/cache/src/lib/caches/auctions.ts
wanglonghong: we don't add the same bid more than once.  the problem exists when reading

rhlsthrm (approved):

PR #1020, opened 2022-04-20
Sanity

jakekidd:
Addresses #990


packages/agents/sequencer/src/lib/operations/auctions.ts
```diff
@@ -53,11 +54,30 @@ export const storeBid = async (
...
-  // - Relayer is approved for this chain (?).
-  // - Router has sufficient funds (?). May want to actually check this later.
+  // Sanity: Check router is approved.
+  const { router } = bid;
+  const { destinationDomain } = bidData.params;
+  let routerApproved = await cache.routers.getApproval(router, destinationDomain);
+  if (!routerApproved) {
+    // Regardless of whether the cache returned false or undefined, we want to check via subgraph
+    // now to see if router is approved.
+    routerApproved = await subgraph.isRouterApproved(destinationDomain, router);
```
rhlsthrm, line 64: A router has to be approved to add liquidity so technically checking liquidity should cover all the cases of router approved, asset approved, and liquidity added.

packages/agents/sequencer/src/lib/operations/auctions.ts
```diff
@@ -170,10 +190,55 @@ export const executeAuctions = async (_requestContext: RequestContext) => {
...
+          const { router } = randomBid;
+          const { amount: _amount, local: asset } = bidData;
+          const amount = BigNumber.from(_amount);
+          let routerLiquidity: BigNumber | undefined = await cache.routers.getLiquidity(router, destination, asset);
+
+          if (!routerLiquidity) {
+            // Either we haven't cached the liquidity yet, or the value cached has become expired.
+            routerLiquidity = await subgraph.getAssetBalance(destination, router, asset);
+            if (routerLiquidity) {
+              await cache.routers.setLiquidity(router, destination, asset, routerLiquidity);
```
rhlsthrm, line 203: Is it dangerous to have this value cached vs querying the subgraph every time? Subgraphs do a lot of caching themselves. I'm a little worried this value is not going to get properly decremented when the liquidity is removed after an execution and the cache will become out of sync.

packages/agents/sequencer/src/lib/operations/auctions.ts
```diff
@@ -194,6 +259,12 @@ export const executeAuctions = async (_requestContext: RequestContext) => {
               origin,
               destination,
             });
+
+            // Update router liquidity record to reflect spending.
+            routerLiquidity = routerLiquidity.sub(amount);
+            await cache.routers.setLiquidity(router, destination, asset, routerLiquidity);
```
rhlsthrm, line 265: I see liquidity is being removed here, but it seems like it won't get re-added on its own, it will expire itself then we rely on the value expiring and then getting re-queried? I guess that works I just don't fully see why we need to cache the liquidity at all.

packages/agents/sequencer/src/lib/operations/auctions.ts
jakekidd: I mean... router could have been unapproved loll, but okay, that's cool with me

packages/agents/sequencer/src/lib/operations/auctions.ts
jakekidd: As soon as we send off a single execute, it will become out of sync

packages/agents/sequencer/src/lib/operations/auctions.ts
jakekidd: you're right that it could probs just be set in memory in some local variable/mapping. I just thought that using the cache was a better model for storage, even if it's truly just temporary.

packages/agents/sequencer/src/lib/operations/auctions.ts
jakekidd: We are rapid-fire looping through these txs. When an `execute` is sent off (to the relayer), it is never sent+accepted right away. Even when it is accepted, the amount of liquidity difference won't be confirmed for another few blocks.

I'm trying to avoid double-spend of the router's funds. Although technically speaking, it likely wouldn't end in reverts here. It would end in the relayer doing nothing, and then we need track what the relayer is doing, and then we need to re-queue the transfer, etc etc etc

This way things are "pessimistic" in that we assume the tx will go thru and the router will be debited.

jakekidd: I think of the subgraph's record as a lagging indicator of router liquidity

jakekidd: Regardless, I think you might be right in another sense... Ima have router liquidity expire a lot faster, and remove the approval stuff for now (I kri... my unit tests, I worked so hard....)

packages/agents/sequencer/src/lib/operations/auctions.ts
jakekidd: Actually, I am sticking by this *for now*. It seems like it might be a useful/informative sanity check, esp. for new routers, if nothing else.

Left a TODO to later re-evaluate and remove if not useful. But for right now, it's clear we're struggling to have visibility on what's going wrong on sequencer.

Our hh tasks all check for it. For instance:
https://discord.com/channels/454734546869551114/964579852458012742/966795745988591627

Also: updated the expiry time for cached approval to be 24 hours

rhlsthrm: I don't get it. The only place the approved routers are checked is when you're adding liquidity: https://github.com/connext/nxtp/blob/amarok/packages/deployments/contracts/contracts/Connext.sol#L644

So having liquidity necessarily means your router is approved. Why do we need to specifically check for approval?

jakekidd: alright, i'll remove it

PR #1031, opened 2022-04-22
962 origin domain callback

liu-zhipeng:
## Checklist

- [ ] Run load tests.
- [ ] Update CHANGELOG.md.


packages/deployments/contracts/contracts/Connext.sol
```diff
@@ -574,16 +602,7 @@ contract Connext is
...
-        _transferId,
-        amount,
-        payable(_args.params.to),
-        adopted,
-        _isFast
-          ? LibCrossDomainProperty.EMPTY_BYTES
-          : LibCrossDomainProperty.formatDomainAndSenderBytes(_args.params.originDomain, _args.originSender),
-        _args.params.callData
-      );
+      _handleExcute(_transferId, amount, adopted, _isFast, _args);
```
LayneHaber, line 605: small typo -- `_handleExecute`

packages/deployments/contracts/contracts/Connext.sol
```diff
@@ -648,6 +675,9 @@ contract Connext is
     if (!isAssetOwnershipRenounced() && !approvedAssets[id]) revert Connext__addLiquidityForRouter_badAsset();
 
     // Transfer funds to coethWithErcTransferact
+    if (_local == address(0) && msg.value != _amount) {
+      revert Connext__addLiquidityForRouter_badAmount();
+    }
```
LayneHaber, line 680: this is not necessary as it is asserted/checked in the `AssetLogic.transferAssetToContract` function

packages/deployments/contracts/contracts/Connext.sol
```diff
@@ -776,6 +806,44 @@ contract Connext is
...
+   * some provided fallback address
+   * @param _transferId Unique identifier of transaction id that necessitated
+   * calldata execution
+   * @param _amount The amount to approve or send with the call
+   * @param _assetId The assetId of the funds to approve to the contract or
+   * send along with the call
+   * @param _isFast Is the Fast liquidity?
+   * @param _args - The `ExecuteArgs` for the transfer
+   */
+  function _handleExcute(
```
LayneHaber, line 823: (see above)

packages/deployments/contracts/contracts/interfaces/IConnext.sol
```diff
@@ -251,7 +255,8 @@ interface IConnext {
     uint256 _domain,
     address payable _bridgeRouter,
     address _tokenRegistry, // Nomad token registry
-    address _wrappedNative
+    address _wrappedNative,
+    address payable _promiseRouter
```
LayneHaber, line 259: will this be a payable address, or just an address? (same with `_bridgeRouter`)

packages/deployments/contracts/contracts/lib/Connext/AssetLogic.sol
```diff
@@ -31,7 +31,7 @@ library AssetLogic {
     if (_assetId == address(0)) {
       // When transferring native asset to the contract, always make sure that the
       // asset is properly wrapped
-      if (msg.value != _specifiedAmount) revert AssetLogic__transferAssetToContract_notAmount();
+      if (msg.value < _specifiedAmount) revert AssetLogic__transferAssetToContract_notAmount();
```
LayneHaber, line 34: is there a reason the equality shouldn't be asserted?

packages/deployments/contracts/contracts/nomad-xapps/contracts/XAppConnectionClient.sol
```diff
@@ -25,7 +25,7 @@ abstract contract XAppConnectionClient is OwnableUpgradeable {
 
   // ======== Initializer =========
 
-  function __XAppConnectionClient_initialize(address _xAppConnectionManager) internal initializer {
+  function __XAppConnectionClient_initialize(address _xAppConnectionManager) internal onlyInitializing {
```
LayneHaber, line 28: we should avoid changing these functions as eventually we will want to import the nomad contracts from the nomad packages

LayneHaber: (tried recently and was still having some issues)

packages/deployments/contracts/contracts/nomad-xapps/contracts/promise-router/PromiseMessage.sol
```diff
@@ -83,32 +92,38 @@ library PromiseMessage {
...
+    // before: 1 byte identifier + 32 bytes transferId + 20 bytes callback = 53 bytes
+    return _view.indexUint(53, 1) == 1;
   }
 
+  /**
+   * @notice Parse the returnData length from the message
+   * @param _view The message
+   * @return The returnData length
+   */
+  function lengthOfReturndata(bytes29 _view) internal pure returns (uint256) {
```
LayneHaber, line 109: should all be camel cased (i.e. `lengthOfReturnData`)

packages/deployments/contracts/contracts/nomad-xapps/contracts/promise-router/PromiseMessage.sol
```diff
@@ -118,10 +133,10 @@ library PromiseMessage {
    */
   function isValidPromiseCallbackLength(bytes29 _view) internal pure returns (bool) {
     uint256 _len = _view.len();
-    uint256 _length = lengthOfCalldata(_view);
-    // before = 1 byte identifier + 32 bytes transferId + 20 bytes callback address + 32 bytes length + x bytes data
-    // nonzero callback data
-    return _len > CALLDATA_START && _length > 0 && (CALLDATA_START + _length) == _len;
+    uint256 _length = lengthOfReturndata(_view);
```
LayneHaber, line 136: if you are dealing with an invalid `_view`, then this length could return an incorrect/arbitrary value. this would invalidate the `(RETURNDATA_START + _length) == _len` equality in some edge cases. 

with that being said, it is a bit difficult to create a strong assertion here -- the only thing that would hold is that it is the first statement (`_len > RETURNDATA_START`). maybe we could just add some comments around it about that? curious your thoughts @patitonar @lmcorbalan

packages/deployments/contracts/contracts/nomad-xapps/contracts/promise-router/PromiseRouter.sol
```diff
@@ -26,13 +30,22 @@ contract PromiseRouter is Version0, Router {
...
+  error PromiseRouter__process_notContractCallback();
+  error PromiseRouter__bumpCallbackFee_valueIsZero();
 
   // ============ Public Storage ============
 
   IConnext public connext;
 
+  // callback messages transferId => message
+  mapping(bytes32 => bytes) public promiseMessages;
+  mapping(bytes32 => uint256) public callbackFees;
```
LayneHaber, line 47: all of these stored items need natspec documentation on them

packages/deployments/contracts/contracts/nomad-xapps/contracts/promise-router/PromiseRouter.sol
```diff
@@ -154,26 +187,58 @@ contract PromiseRouter is Version0, Router {
...
-    // Should enforce relayer is whitelisted by calling local connext contract
+    // enforce relayer is whitelisted by calling local connext contract
+    if (!connext.isApprovedRelayer(msg.sender)) revert PromiseRouter__process_notApprovedRelayer();
 
     // Should transfer the stored relayer fee to the msg.sender
+    // TODO check if callbackFee which already paid is sufficient or not
+    //      If not, should revert and should add more fee.
+    // if (gasFee > callbackFees[transferId]) {
+    //   revert PromiseRouter__process_insufficientCallbackFee();
+    // }
```
LayneHaber, line 215: I think we can remove this `TODO` -- we should have users in charge of setting their own fee rather than relying on the domain to establish it

packages/deployments/contracts/contracts_forge/PromiseMessage.t.sol
```diff
@@ -0,0 +1,59 @@
...
+  using TypedMemView for bytes29;
+  using PromiseMessage for bytes29;
+
+  // ============ Test set up ============
+
+  function setUp() public {}
+
+  // ============ format/parse ============
+  // Should work
+  function test_PromiseCallbackMessage_should_format_and_parse_a_massage_properly(
```
LayneHaber, line 32: just reformatted the names to be follow the structure: `test_${ContractName}__${functionName}_${testName}`

which means this test name should be:

`test_PromiseMessage__formatMessage_shouldWork`

or something (test name is a little tricky here, but the main thing is camel over snake casing)

packages/deployments/contracts/contracts_forge/PromiseRouter.t.sol
```diff
@@ -0,0 +1,349 @@
...
+
+// running tests (with logging on failure):
+// yarn workspace @connext/nxtp-contracts test:forge -vvv
+// run a single test:
+// yarn workspace @connext/nxtp-contracts test:forge -m testAddRouterAlreadyApproved -vvv
+
+// other forge commands: yarn workspace @connext/nxtp-contracts forge <CMD>
+// see docs here: https://onbjerg.github.io/foundry-book/index.html
+
+contract MackPromiseRouter is PromiseRouter {
```
LayneHaber, line 19: typo -- `MockPromiseRouter`

packages/deployments/contracts/contracts_forge/PromiseRouter.t.sol
```diff
@@ -0,0 +1,349 @@
...
+contract MockCallback is ICallback {
+  function callback(
+    bytes32 transferId,
+    bool success,
+    bytes memory data
+  ) external {
+    require(data.length != 0);
+  }
+}
+
```
LayneHaber, line 57: It seems generally we should have a `Mock.sol` that includes mocks for all of the constituent contracts since we seem to define the same ones in-file.

would be good to start creating that file now, but shouldn't block the PR

LayneHaber (changes requested):
One thing to note is that we are heavily refactoring the architecture of the contracts to follow the one laid out in #960 . it would be good to base this branch off of that architecture so we can avoid introducing more merge conflicts.

packages/deployments/contracts/contracts/nomad-xapps/contracts/promise-router/PromiseRouter.sol
```diff
@@ -154,26 +187,58 @@ contract PromiseRouter is Version0, Router {
...
+    if (callbackFees[transferId] == 0) revert PromiseRouter__process_insufficientCallbackFee();
+
+    address callbackAddress = _msg.callbackAddress();
+
+    if (!AddressUpgradeable.isContract(callbackAddress)) revert PromiseRouter__process_notContractCallback();
+
+    ICallback(callbackAddress).callback(transferId, _msg.returnSuccess(), _msg.returnData());
```
patitonar, line 222: What about the case when the callback call reverts? Is it ok to revert the whole transaction? or should we handle this via `.call` (and probably emit the result in events or something?)?

patitonar, line 222: What about the case when the callback call reverts? Is it ok to revert the whole transaction? or should we handle this via `.call` (and probably emit the result in events or something?)?

packages/deployments/contracts/contracts/lib/Connext/AssetLogic.sol
liu-zhipeng: That is for the callback fee. When calling xcall with ETH, callback fee will be included in `msg.value`. So `msg.value` will be greater than `amount`.

packages/deployments/contracts/contracts/interfaces/IConnext.sol
liu-zhipeng: `_promiseRouter` address should be payable address. because this address should be able to receive ETH as callback fee

packages/deployments/contracts/contracts/Connext.sol
liu-zhipeng: Because I updated `transferAssetToContract` function like below due to callback fee.
```
    if (_assetId == address(0)) {
      // When transferring native asset to the contract, always make sure that the
      // asset is properly wrapped
      if (msg.value < _specifiedAmount) revert AssetLogic__transferAssetToContract_notAmount();
      _wrapper.deposit{value: _specifiedAmount}();
      _assetId = address(_wrapper);
    } 
```
So i had to check `msg.value == amount` before calling `transferAssetToContract` function.

packages/deployments/contracts/contracts/nomad-xapps/contracts/XAppConnectionClient.sol
liu-zhipeng: yeah, but that is old version of openzeppelin.  they might have to update that later. 
anyway i will revert that for now.

packages/deployments/contracts/contracts/nomad-xapps/contracts/promise-router/PromiseRouter.sol
liu-zhipeng: Added

packages/deployments/contracts/contracts/nomad-xapps/contracts/promise-router/PromiseRouter.sol
liu-zhipeng: Okay

packages/deployments/contracts/contracts_forge/PromiseMessage.t.sol
liu-zhipeng: all renamed!
Good structure!

packages/deployments/contracts/contracts/nomad-xapps/contracts/promise-router/PromiseRouter.sol
liu-zhipeng: yeah, that is very good point. maybe we have to discuss that more.

packages/deployments/contracts/contracts/lib/Connext/AssetLogic.sol
liu-zhipeng: Fixed in `handleIncomingAsset` function

packages/deployments/contracts/contracts_forge/PromiseRouter.t.sol
liu-zhipeng: Created!

packages/deployments/contracts/contracts/nomad-xapps/contracts/promise-router/PromiseMessage.sol
```diff
@@ -22,13 +22,13 @@ library PromiseMessage {
...
   // ============ Constants ============
   uint256 private constant IDENTIFIER_LEN = 1;
-  // 1 byte identifier + 32 bytes transferId + 20 bytes callback + 32 bytes length + x bytes data
-  // before: 1 byte identifier + 32 bytes transferId + 20 bytes callback = 53 bytes
-  uint256 private constant LENGTH_CALLDATA_START = 53;
-  uint8 private constant LENGTH_CALLDATA_LEN = 32;
+  // 1 byte identifier + 32 bytes transferId + 20 bytes callback + 1 byte success + 32 bytes length + x bytes data
+  // before: 1 byte identifier + 32 bytes transferId + 20 bytes callback + 1 byte success= 54 bytes
+  uint256 private constant LENGTH_RETURNDATA_START = 54;
+  uint8 private constant LENGTH_RETURNDATA_LEN = 32;
```
jakekidd, line 28: Why does `PromiseMessage` need this extra 1 byte success? Just curious

packages/deployments/contracts/contracts/lib/Connext/ConnextLogic.sol
```diff
@@ -714,6 +718,11 @@ library ConnextLogic {
     if (_args.xCallArgs.params.to == address(0)) {
       revert ConnextLogic__xcall_emptyTo();
     }
+
+    // ensure callback fee is zero when callback address is not contract
+    if (!AddressUpgradeable.isContract(_args.xCallArgs.params.callback) && _args.xCallArgs.params.callbackFee > 0) {
+      revert ConnextLogic__xcall_nonZeroCallbackFeeForNonContractCallback();
+    }
```
LayneHaber, line 725: we should allow 0 fees as this will happen in an hour and if a relayer chooses to forward without a fee, that is their decision to make.

packages/deployments/contracts/contracts/nomad-xapps/contracts/promise-router/PromiseRouter.sol
```diff
@@ -154,23 +211,58 @@ contract PromiseRouter is Version0, Router {
...
   }
 
-  function process(bytes32 transactionId, bytes memory _message) internal {
-    // Should parse out the return data and callback address from message
-    // Should execute the callback() function on the provided Callback address
-    // Should enforce relayer is whitelisted by calling local connext contract
+  function process(bytes32 transferId) public payable nonReentrant {
+    // parse out the return data and callback address from message
+    bytes memory _message = promiseMessages[transferId];
+    if (_message.length == 0) revert PromiseRouter__process_InvalidMessage();
```
LayneHaber, line 227: casing of error is wrong here

packages/deployments/contracts/contracts/nomad-xapps/contracts/promise-router/PromiseRouter.sol
```diff
@@ -41,33 +63,63 @@ contract PromiseRouter is Version0, Router {
...
    */
   event Receive(
     uint64 indexed originAndNonce,
     uint32 indexed origin,
     bytes32 transferId,
     address callbackAddress,
-    bytes data
+    bool success,
+    bytes data,
+    bytes message
```
jakekidd, line 103: is this message (informally) bounded in length somewhere? if not, is emitting the message advisable?

packages/deployments/contracts/contracts/nomad-xapps/contracts/promise-router/PromiseRouter.sol
```diff
@@ -154,23 +211,58 @@ contract PromiseRouter is Version0, Router {
...
+    bytes memory _message = promiseMessages[transferId];
+    if (_message.length == 0) revert PromiseRouter__process_InvalidMessage();
+
+    bytes29 _msg = _message.ref(0).mustBePromiseCallback();
+
+    // enforce relayer is whitelisted by calling local connext contract
+    if (!connext.isApprovedRelayer(msg.sender)) revert PromiseRouter__process_notApprovedRelayer();
+
+    // enfore callback fee is not zero
+    if (callbackFees[transferId] == 0) revert PromiseRouter__process_insufficientCallbackFee();
```
LayneHaber, line 235: should be permitted

packages/deployments/contracts/contracts/nomad-xapps/contracts/promise-router/PromiseRouter.sol
```diff
@@ -154,23 +211,58 @@ contract PromiseRouter is Version0, Router {
...
+
+    address callbackAddress = _msg.callbackAddress();
+
+    if (!AddressUpgradeable.isContract(callbackAddress)) revert PromiseRouter__process_notContractCallback();
+
+    (bool success, ) = callbackAddress.call(
+      abi.encodeWithSelector(ICallback.callback.selector, transferId, _msg.returnSuccess(), _msg.returnData())
+    );
```
LayneHaber, line 243: why use lower level call instead of using the interface?

LayneHaber: to pay the relayer for failing calldata?

packages/deployments/contracts/contracts/nomad-xapps/contracts/promise-router/PromiseRouter.sol
jakekidd, line 235: What happens if the callback executor doesn't want a fee (i.e. what if it's the user)?

packages/deployments/contracts/contracts/nomad-xapps/contracts/promise-router/PromiseRouter.sol
```diff
@@ -154,23 +211,58 @@ contract PromiseRouter is Version0, Router {
...
     // emit Receive event
-    emit Receive(_originAndNonce(_origin, _nonce), _origin, transferId, callbackAddress, data);
+    emit Receive(_originAndNonce(_origin, _nonce), _origin, transferId, callbackAddress, success, data, _message);
   }
 
-  function process(bytes32 transactionId, bytes memory _message) internal {
-    // Should parse out the return data and callback address from message
-    // Should execute the callback() function on the provided Callback address
-    // Should enforce relayer is whitelisted by calling local connext contract
+  function process(bytes32 transferId) public payable nonReentrant {
```
jakekidd, line 224: Where function docstring/descriptor? :'((

packages/deployments/contracts/contracts/nomad-xapps/contracts/promise-router/PromiseRouter.sol
LayneHaber: there are upper bounds within the nomad system when you send a message, so we kind of offload that to them

packages/deployments/contracts/contracts/nomad-xapps/contracts/promise-router/PromiseRouter.sol
jakekidd: So this method is external now? Basically, we set the `promiseMessages[transferId]` in the `handler`, and anyone can call `process` ?

packages/deployments/contracts/contracts/nomad-xapps/contracts/promise-router/PromiseRouter.sol
jakekidd: Idk if the optimizer is handling this, but we can probs afford to assign to memory a value here:
```
uint256 fee = callbackFees[transferId];
```
then we don't have to do a second SLOAD/read operation below

packages/deployments/contracts/contracts/nomad-xapps/contracts/promise-router/PromiseMessage.sol
LayneHaber: that is the `success` of the calldata executed on the destination domain

packages/deployments/contracts/contracts/nomad-xapps/contracts/promise-router/PromiseRouter.sol
jakekidd, line 227: Don't we need to encode and hash to check message length?

packages/deployments/contracts/contracts/nomad-xapps/contracts/promise-router/PromiseRouter.sol
jakekidd: Seems like it should be limited to relayers only, if this is the case. Otherwise we are subject to frontrunning problems

packages/deployments/contracts/contracts/nomad-xapps/contracts/promise-router/PromiseMessage.sol
liu-zhipeng: `success` field is the boolean field that shows if an external call success or not.

packages/deployments/contracts/contracts/lib/Connext/ConnextLogic.sol
liu-zhipeng: Okay

packages/deployments/contracts/contracts/nomad-xapps/contracts/promise-router/PromiseRouter.sol
```diff
@@ -154,23 +211,58 @@ contract PromiseRouter is Version0, Router {
...
     // Should transfer the stored relayer fee to the msg.sender
+    AddressUpgradeable.sendValue(payable(msg.sender), callbackFees[transferId]);
+    callbackFees[transferId] = 0;
+  }
+
+  /**
+   * @notice This function will be called on the origin domain to increase the callback fee
+   * @param _transferId - The unique identifier of the crosschain transaction
+   */
+  function bumpCallbackFee(bytes32 _transferId) external payable {
```
jakekidd, line 257: Technically speaking, anyone can bump for transferId that doesn't exist.

While we can typically just say 'ok, but don't bump callback fee for random non-existent transferIds'

there's still a bit of a problem here - there's a race condition where the user may have submitted a bump, but the `process` function has already been called (within the same block, or perhaps 1 block ago, etc.)

So, we should probably check to make sure that this isn't the case.
- We could check to make sure `promiseMessages[transferId]` exists, but that would mean we can only bump fee *after* `handle` has been called, and not prior to handle (e.g. in anticipation that a higher fee may be needed bc market conditions have changed, etc)
- Seeing as how we enforce that the callback fee **must** be > 0 when we call `process`, maybe we can enforce that the callback fee is > 0 when we call send? If so, we could technically check `callbackFee[_transferId] > 0` as a requirement to bump.
- Otherwise, just add a separate mapping for `processedTransferIds` and mark it as true at the end of the `process` call above, and then check it here

packages/deployments/contracts/contracts/nomad-xapps/contracts/promise-router/PromiseRouter.sol
jakekidd: also, why payable? lol

packages/deployments/contracts/contracts/nomad-xapps/contracts/promise-router/PromiseRouter.sol
liu-zhipeng: zero Fee or any account?

packages/deployments/contracts/contracts/nomad-xapps/contracts/promise-router/PromiseRouter.sol
LayneHaber: it checks the relayer in there

packages/deployments/contracts/contracts/nomad-xapps/contracts/promise-router/PromiseRouter.sol
LayneHaber: this is more of an empty / not null check

packages/deployments/contracts/contracts/nomad-xapps/contracts/promise-router/PromiseRouter.sol
LayneHaber: you should be able to bump the fee before handle is called -- why should you not be able to?

packages/deployments/contracts/contracts/nomad-xapps/contracts/promise-router/PromiseRouter.sol
LayneHaber: does not need to be payable though

packages/deployments/contracts/contracts/nomad-xapps/contracts/promise-router/PromiseRouter.sol
LayneHaber: ahh okay i see what you're saying -- you can either bump a transfer that doesn't exist or one that was already processed unless there's a flag. 

i dont think we should require the process fees are greater than 0 -- ill go with the flag

LayneHaber (approved):

packages/deployments/contracts/contracts/nomad-xapps/contracts/promise-router/PromiseRouter.sol
LayneHaber: ahh okay i see what you're saying -- you can either bump a transfer that doesn't exist or one that was already processed unless there's a flag.

i dont think we should require the process fees are greater than 0 -- ill go with the first option as updating 3 mappings seems insane

Issue #929, opened 2022-04-04
[e2e] Extend smoke tests to form base of integration tests


rhlsthrm:
The current smoke tests can form the base of an integration test suite. The MVP version of integration tests doesn't need to run Nomad ops.

We need to:
- Spin up chains
- Deploy contracts (find out what's the minimum subset, most likely Connext contracts + token registry)
- Spin up subgraphs
- Run a simple test suite to send a transfer and verify that it gets completed (need to figure out relayer issue here)

jakekidd:
https://github.com/connext/nxtp/pull/1075 merged

PR #1033, opened 2022-04-23
[integration] Testnet e2e integration test

jakekidd:
## What is?

Made this test to improve visibility, accountability, and ease-of-use for conducting e2e. Can be run with a local router/sequencer pair or using the live hosted router/sequencer instances.

Additionally, tidied up logs a bit across the board, and added some configuration options to make running locally a lot more feasible.

Partly (?) addresses #929.

## How do?
Check out the README, it's like, actually informative


packages/agents/router/src/router.ts
```diff
@@ -97,10 +98,23 @@ export const makeRouter = async () => {
...
     await bindCache();
 
-    logger.info("Router ready!");
+    logger.info("Bindings initialized.");
+
+    console.log(
+      "Router boot complete!",
+      `
+
+        _|_|_|     _|_|     _|      _|   _|      _|   _|_|_|_|   _|      _|   _|_|_|_|_|
```
rhlsthrm, line 110: <3

rhlsthrm (approved):
Definitely should try to pull from deployments but other than that, looks great!

packages/agents/router/src/router.ts
```diff
@@ -97,10 +98,23 @@ export const makeRouter = async () => {
...
     await bindServer();
     await bindMetrics();
-    await bindSubgraph(context.config.subgraphPollInterval!);
+    await bindSubgraph();
     await bindCache();
 
-    logger.info("Router ready!");
+    logger.info("Bindings initialized.");
+
+    console.log(
```
rhlsthrm, line 106: Why is this not logger.info?

rhlsthrm (changes requested):

packages/agents/router/src/router.ts
jakekidd: I can make the message logger.info, but the connext branding thing has to be console.log or it literally will just show up incoherently in the logs haha

PR #1064, opened 2022-04-28
Abstract Relayer

rhlsthrm:
## The Problem

Relayer needs to use special params to allow testnet to work.

## The Solution

Fix the params, but also separate relayer implementation into adapter.

## Checklist

- [ ] Run load tests.
- [ ] Update CHANGELOG.md.


wanglonghong (approved):
It looks like a good change in architecture to have an abstract relayer.

Issue #1060, opened 2022-04-27
[contracts] Third Party Sponsoring Fees


LayneHaber:
### Discussed in https://github.com/connext/nxtp/discussions/935

<div type='discussions-op-text'>

<sup>Originally posted by **LayneHaber** April  4, 2022</sup>
# Summary

Third parties (i.e. chain teams) can fund a `Vault` that will pay the liquidity and relayer fee (up to a cap) in for each transfer. This will allow them to create free onboarding onto their chain via subsidies, and give our users a better UX by (1) covering fees and (2) providing them with gas on the chain immediately.

# Motivation

Allowing third parties to sponsor transfers would create an opportunity for domain teams to support free bridging. Coupled with providing liquidity incentives for routers who support the domain, this creates an easy onboarding incentive program and experience.

With nomad, we are able to be the default bridge on many new domains as they launch, and allowing domain teams to sponsor transfers and liquidity would be the base for a great incentive program.

# Design Requirements

- Must allow third parties to sponsor `relayerFee` for each transfer
- Must allow third parties to sponsor `liquidityFee` for each transfer
- Should be easy to fund on the destination domain only
- Should not require origin domain changes (contracts shouldn’t need to upgrade to have their transfers sponsored)

# Proposed Solution

There is a whitelist that contains a `Vault` address on the domain. The `Vault` is a contract that can be funded by the sponsor, and has the following interface:

```tsx
interface IVault {
  // Should be callable by the Connext contract only. Should:
  // - call `addLiquidityFor` to send the calculated fee to the router
  // - return the amount of liquidity router was reimbursed
  function reimburseLiquidityFees(uint256 fee, address asset, address router) public returns (uint256);

  // Should be callable by the Connext contract only. Should:
  // - take in an amount of relayer fee specified on origin chain
  // - convert that amount to destination domain gas
  // - send the user the destination domain gas
  // - 
  function reimburseRelayerFees(uint256 relayerFee) public;
}
```

Then, a crosschain transfer would look like:

1. User makes an `xcall` with a specified destination domain (domain is offering rewards)
    1. `relayerFee` is specified
2. Router sees transaction, bid is selected, `execute` is called on destination domain
3. Within `execute`, a whitelist is checked to see if the current domain is offering rebates. The whitelist is a `mapping(uint32 => Vault)`
    1. This whitelist is set by an `onlyOwner` function
4. If it is not whitelisted for rewards, deduct liquidity fees and send relayer fees to relayer
[...]

PR #1066, opened 2022-04-28
Sponsoring fee execute logic

lmcorbalan:
## The Problem

subtask of #1060 
ConnextHandler execute logic needs to be integrated with ISponsorVault to be able to sponsor liquidity and relayer fees

## The Solution

Add new logic to handle sponsoring fees

## Checklist

- [ ] Run load tests.
- [ ] Update CHANGELOG.md.

### DIFF
--- a/packages/deployments/contracts/contracts/lib/Connext/ConnextLogic.sol
+++ b/packages/deployments/contracts/contracts/lib/Connext/ConnextLogic.sol
@@ -5,6 +5,7 @@ import {IConnextHandler} from "../../interfaces/IConnextHandler.sol";
 import {IStableSwap} from "../../interfaces/IStableSwap.sol";
 import {IWrapped} from "../../interfaces/IWrapped.sol";
 import {IExecutor} from "../../interfaces/IExecutor.sol";
+import {ISponsorVault} from "../../interfaces/ISponsorVault.sol";
 import {LibCrossDomainProperty} from "../LibCrossDomainProperty.sol";
 import {RouterPermissionsManagerInfo} from "./RouterPermissionsManagerLogic.sol";
 import {AssetLogic} from "./AssetLogic.sol";
@@ -17,7 +18,7 @@ import {TypeCasts} from "../../nomad-core/contracts/XAppConnectionManager.sol";
 import {Home} from "../../nomad-core/contracts/Home.sol";
 
 import {ECDSAUpgradeable} from "@openzeppelin/contracts-upgradeable/utils/cryptography/ECDSAUpgradeable.sol";
-import {SafeERC20Upgradeable, AddressUpgradeable} from "@openzeppelin/contracts-upgradeable/token/ERC20/utils/SafeERC20Upgradeable.sol";
+import {SafeERC20Upgradeable, IERC20Upgradeable, AddressUpgradeable} from "@openzeppelin/contracts-upgradeable/token/ERC20/utils/SafeERC20Upgradeable.sol";
 
 library ConnextLogic {
   // ============ Libraries ============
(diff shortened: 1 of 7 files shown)


packages/deployments/contracts/contracts/lib/Connext/ConnextLogic.sol
```diff
@@ -151,6 +155,13 @@ library ConnextLogic {
    */
   event MaxRoutersPerTransferUpdated(uint256 maxRoutersPerTransfer, address caller);
 
+  /**
+   * @notice Emitted when the sponsorVault variable is updated
+   * @param sponsorVault - The sponsorVault new value
+   * @param caller - The account that called the function
+   */
+  event SponsorVaultUpdated(address sponsorVault, address caller);
+
```
liu-zhipeng, line 164: It would be better if includes old `sponsorVault` address on event.

packages/deployments/contracts/contracts/lib/Connext/ConnextLogic.sol
```diff
@@ -339,6 +350,18 @@ library ConnextLogic {
...
+   * @notice Used to set the sponsor vault
+   * @param _sponsorVault The new sponsor vault
+   * @param _currentSponsorVault The current sponsor vault
+   */
+  function setSponsorVault(address _sponsorVault, address _currentSponsorVault) external {
+    if (_sponsorVault == _currentSponsorVault)
+      revert ConnextLogic__setSponsorVault_invalidSponsorVault();
+
+    emit SponsorVaultUpdated(_sponsorVault, msg.sender);
+  }
```
liu-zhipeng, line 363: There is no parts to set new sponsorVault. or this is for just emitting event?

packages/deployments/contracts/contracts/lib/Connext/ConnextLogic.sol
```diff
@@ -339,6 +350,18 @@ library ConnextLogic {
     emit MaxRoutersPerTransferUpdated(_newMax, msg.sender);
   }
 
+  /**
+   * @notice Used to set the sponsor vault
+   * @param _sponsorVault The new sponsor vault
+   * @param _currentSponsorVault The current sponsor vault
+   */
+  function setSponsorVault(address _sponsorVault, address _currentSponsorVault) external {
```
jakekidd, line 358: we need to require that this request comes from the `_currentSponsorVault` or something... otherwise, it could be a griefing vector to overwrite sponsor's vaults freely

packages/deployments/contracts/contracts/lib/Connext/ConnextLogic.sol
```diff
@@ -875,6 +898,33 @@ library ConnextLogic {
...
+    if (address(_args.sponsorVault) != address(0)) {
+      // fast liquidity path
+      if (!_reconciled) {
+        // Vault will return the amount of the fee they sponsored in the native fee
+        // NOTE: some considerations here around fee on transfer tokens and ensuring
+        // there are no malicious `Vaults` that do not transfer the correct amount. Should likely do a
+        // balance read about it
+
+        uint256 starting = IERC20Upgradeable(_adopted).balanceOf(address(this));
+        uint256 sponsored = _args.sponsorVault.reimburseLiquidityFees(_adopted, _args.executeArgs.amount);
```
jakekidd, line 912: Have we addressed/discussed what happens if devs do *not* want to sponsor liq fees but *do* want to sponsor relayer fees?

jakekidd: Obviously, this would return 0 right?

But that would be really gas inefficient, what with doing this contract call to `reimburseLiquidityFees` and then a bunch of useless addition.

We could have a mapping, however, that the sponsor vault owner could update in `setSponsorVault` that indicates whether to reimburse liquidity fees or not. Would technically add the cost of accessing storage in the use case where we *are* reimbursing both fees, but access lists in the contract call could alleviate that....

packages/deployments/contracts/contracts/lib/Connext/ConnextLogic.sol
jakekidd: ah wait, misinterpreted the use case entirely. The `setSponsorVault` call comes from the `ConnextHandler`, which requires the owner. My b.

LayneHaber (approved):
lgtm! (pending liu's event fixes)

PR #1068, opened 2022-04-28
[relayer] Add a basic relayer agent we can run locally; debug e2e

jakekidd:
## What is?

E2E testing was briefly blocked due to gelato relayer not working. While we can definitely rely on gelato in the future, we should be prepared to have brief moments like this where a new featureset conflicts with the current gelato implementation. Being able to adapt to new featuresets quickly so we can move forward with quickly iterating/debugging any issues with E2E is highly desirable at this stage.

## How do?

Pass in a relayerUrl to sequencer, and it will use that relayer first before it tries gelato. 

You can run the normal integration tests locally (include a router mnemonic), and put `LOCAL_RELAYER_ENABLED=true`. Specify a specific `RELAYER_MNEMONIC` if you like, or it will just reuse your router's wallet as a backup.

If you really want to, you can run the relayer locally with `yarn workspace @connext/nxtp-relayer run dev`.

Also, this includes numerous fixes to subgraph queries, FallbackSubgraph, issues with xcall/execute arguments, etc.


rhlsthrm (commented):
Nice work, this was a significant effort and will be very useful for e2e/local testing as well as expanding to domains if we need support faster than Gelato is available. Still unclear how to configure the relayer URL but I will look at the code.

PR #1069, opened 2022-04-28
Add SponsorVault contract

lmcorbalan:
## The Problem

subtask of #1060 
There is no SponsorVault contract

## The Solution

Create SponsorVault contract

## Checklist

- [ ] Run load tests.
- [ ] Update CHANGELOG.md.


packages/deployments/contracts/contracts/SponsorVault.sol
```diff
@@ -0,0 +1,237 @@
...
+  address public connext;
+
+  /**
+   * @notice The origin domain to this domain gas token rates
+   * @dev Used when no oracle is available
+   */
+  mapping(uint32 => Rate) public rates;
+
+  /**
+   * @notice The maximum amount of relayer fee to sponsor
```
LayneHaber, line 50: need a note about what asset this is in

packages/deployments/contracts/contracts/SponsorVault.sol
```diff
@@ -0,0 +1,237 @@
...
+   */
+  uint256 public relayerFeeCap;
+
+  /**
+   * @notice The origin domain to this domain gas token oracle
+   * @dev Used to calculate sponsored relayer fee
+   */
+  address public gasTokenOracle;
```
LayneHaber, line 58: shouldn't this conform to an interface?

packages/deployments/contracts/contracts/SponsorVault.sol
```diff
@@ -0,0 +1,237 @@
...
+
+  /**
+   * @notice The this domain gas token to token exchange
+   * @dev Used to calculate sponsored relayer fee
```
LayneHaber, line 62: need some more clarification in this comment as well -- is this a dex you can use to get a price or is it a dex used to swap out some tokens? im not sure why you would need to swap anything, you should just need to fund this vault with lots of native asset, and then some way to get the `origin domain native asset : current domain native asset` price

packages/deployments/contracts/contracts/SponsorVault.sol
```diff
@@ -0,0 +1,237 @@
...
+
+  /**
+   * @notice Emitted when a new gas token oracle is set
+   */
+  event GasTokenOracleUpdated(address oracle, address caller);
+
+  /**
+   * @notice Emitted when a new token exchange is set
+   */
+  event TokenExchangeUpdated(address token, address tokenExchange, address caller);
```
LayneHaber, line 99: events should probably have old / new values here

packages/deployments/contracts/contracts/SponsorVault.sol
LayneHaber: ahhhh okay, so it swaps into the right asset? should have some global slippage set by the owner in that case as well then

packages/deployments/contracts/contracts/SponsorVault.sol
```diff
@@ -0,0 +1,237 @@
...
+  /**
+   * @notice Performs liquidity fee reimbursement.
+   * @dev Uses the token exchange or liquidity deposited in this contract.
+   * @param _token The address of the token
+   * @param _liquidityFee The liquidity fee amount
+   * @return Sponsored liquidity fee amount
+   */
+  function reimburseLiquidityFees(address _token, uint256 _liquidityFee) external override onlyConnext returns (uint256) {
+    // do not sponsor liquidity fee when there is no token exchange nor liquidity for the diven token
+    if (tokenExchanges[_token] == address(0) && IERC20(_token).balanceOf(address(this)) < _liquidityFee) return 0;
```
LayneHaber, line 186: if you end up in the case where the balance of the contract is `bal > 0 && bal < liquidityFee` then the balance should be used to sponsor the transfer

packages/deployments/contracts/contracts/SponsorVault.sol
```diff
@@ -0,0 +1,237 @@
...
+    } else if (rates[_originDomain].den != 0) {
+      relayerFee = _originRelayerFee * rates[_originDomain].num / rates[_originDomain].den;
+    }
+
+    relayerFee = relayerFee > relayerFeeCap ? relayerFeeCap : relayerFee;
+
+    // TODO - what if 0 < balance < relayerFee ?
+    if (relayerFee > 0 && address(this).balance >= relayerFee) {
+        Address.sendValue(_to, relayerFee);
+    }
```
LayneHaber, line 224: same as above -- just give what is available even if it doesn't fully reimburse

LayneHaber (changes requested):

packages/deployments/contracts/contracts/SponsorVault.sol
```diff
@@ -0,0 +1,237 @@
...
+  /**
+   * @notice Emitted when a new connext is set
+   */
+  event ConnextUpdated(address connext, address caller);
+
+  /**
+   * @notice Emitted when a new rate is set
+   */
+  event RateUpdated(uint32 originDomain, Rate rate, address caller);
+
```
liu-zhipeng, line 85: I am wondering why we have `caller` address on all events. Actually, we can get `caller` from `transaction.from`  on subgraph handlers. If there is no special reason, it will be inefficient.

packages/deployments/contracts/contracts/SponsorVault.sol
```diff
@@ -0,0 +1,237 @@
...
+    if (tokenExchanges[_token] == address(0) && IERC20(_token).balanceOf(address(this)) < _liquidityFee) return 0;
+
+    if (tokenExchanges[_token] != address(0)) {
+      ITokenExchange tokenExchange = ITokenExchange(tokenExchanges[_token]);
+      uint256 amountIn = tokenExchange.getInGivenExactOut(_token, _liquidityFee);
+
+      if (address(this).balance < amountIn) return 0;
+
+      tokenExchange.swapExactOut{value: amountIn}(_token, _liquidityFee, msg.sender);
+    } else {
```
liu-zhipeng, line 195: `ITokenExchange` would be the external dexs like Uniswap? If so, we have to use their interface. And i think  `swapExactOut` function has to have `amountOutMin` due to slippage.

packages/deployments/contracts/contracts/SponsorVault.sol
lmcorbalan: I followed the pattern from the events defined in Connext.
Regarding getting que `caller` from the `transaction.from` on the subgraph, I think it would have the address of the transaction origin so if the this contract is called from another contract (some governance contract maybe) `caller` is the address of the caller contract  but `transaction.from` is the address of the EOA which initiates the transaction.

packages/deployments/contracts/contracts/SponsorVault.sol
lmcorbalan: Fixed the comment, it was wrong.
The tokenExchange would be used to get the token used to pay the liquidity fee swapping in the current domain native asset wouldn't the `maxAmountIn` be implicitly defined in the `value` sent when calling `swapExactOut` ?

packages/deployments/contracts/contracts/SponsorVault.sol
lmcorbalan: What if the native asset balance is not enough to swap for the whole liquidity fee? Should it pay as liquidity fee the result of swapping in the whole native asset balance?

packages/deployments/contracts/contracts/SponsorVault.sol
lmcorbalan: see https://github.com/connext/nxtp/pull/1069#discussion_r865047730

packages/deployments/contracts/contracts/SponsorVault.sol
liu-zhipeng: okay. then that is fine

packages/deployments/contracts/contracts/SponsorVault.sol
LayneHaber: it should pay as much as it can, yes

packages/deployments/contracts/contracts/SponsorVault.sol
LayneHaber: > The tokenExchange would be used to get the token used to pay the liquidity fee using the current domain native asset to swap in, so wouldn't the maxAmountIn be implicitly defined in the value sent when calling swapExactOut ?

im not sure i follow what you're saying here 😅 . Let me restate to test:

The `tokenExchange` is used to get the token that the liquidity fee is charged in, using the native asset stored on this contract. And you're saying that the `maxAmountIn` is implicitly defined as the `value` when you're sending to the `swapExactOut` call?

I guess that should work, if I'm understanding correctly. How would you know that the `value` is enough for the `swapExactOut` (I'm using the assumption that the `swapExactOut` means you get the exact value out of the AMM, so the value you put *in* could vary)?

LayneHaber: The main critique is just that the contract should have "sensible bounds". Though I could definitely see the argument where you're giving this money away anyways so maybe its okay to have really generous slippage tolerances

packages/deployments/contracts/contracts/SponsorVault.sol
liu-zhipeng: and we have to build a contract for token exchange like a kind of aggregator? or we will use existing dexs like uniswap?

packages/deployments/contracts/contracts/SponsorVault.sol
lmcorbalan: >The tokenExchange is used to get the token that the liquidity fee is charged in, using the native asset stored on this contract. And you're saying that the maxAmountIn is implicitly defined as the value when you're sending to the swapExactOut call?

yes

> I guess that should work, if I'm understanding correctly. How would you know that the value is enough for the swapExactOut (I'm using the assumption that the swapExactOut means you get the exact value out of the AMM, so the value you put in could vary)?

The `ITokenExchange` has a `getInGivenExactOut` function, to get the amount you need to send. 

Now that I think it better it may not work, bc if the amount being sent to the AMM is not enough due to slippage the swap would probably revert due to the exact out, so now I think is would be better to use a `swapExactIn` and use the resulting amount as the sponsor.

packages/deployments/contracts/contracts/SponsorVault.sol
lmcorbalan: > and we have to build a contract for token exchange like a kind of aggregator? or we will use existing dexs like uniswap?

I would say the `tokenExchange` would be an `adaptor` for existing DEXs, if not we would need to choose some specific DEX and use its interface, right?

lmcorbalan:
@LayneHaber Shouldn't the vault have a function for adding and removing liquidity? (native and ERC20 tokens) I think at least `removing` is a must have

packages/deployments/contracts/contracts/SponsorVault.sol
LayneHaber: Agree!

packages/deployments/contracts/contracts/SponsorVault.sol
LayneHaber: I think using the adapter pattern makes more sense -- there are different dexes with different liquidity depths depending on the swap you're doing so we shouldn't force a specific implementation

LayneHaber:
> Shouldn't the vault have a function for adding and removing liquidity? (native and ERC20 tokens) I think at least removing is a must have

you are completely right, missed that in my spec! created a new issue: #1175

LayneHaber (approved):

Issue #1059, opened 2022-04-27
As a router, I want the ability to borrow additional liquidity from AAVE (contract side)


LayneHaber:
### Discussed in https://github.com/connext/nxtp/discussions/939

<div type='discussions-op-text'>

<sup>Originally posted by **LayneHaber** April  5, 2022</sup>
# Summary

Using Aave v3 portals when there is no fast-liquidity available to facilitate a transaction increases the capital efficiency of Amarok as a liquidity layer on top of the nomad message passing infrastructure. The functionality should be restricted to a trusted set of routers, and those routers should only use the unbacked positions if there is no liquidity available. The fee they charge should be sufficient to cover the position + any interest when the full amount comes through

# Motivation

Aave v3 opens up a unique design space for bridges with their [portal feature](https://governance.aave.com/t/portals-powering-the-multi-blockchain-multiverse/6889). Using the portals, bridges have access to large amounts of collateral by taking on unbacked positions for the duration of the bridging. It would be incredible to be able to utilize this feature when providing fast liquidity across nomad.

## Design Requirements

- The `Connext.sol` contract should be whitelisted as a `BRIDGE` on Aave
- Should default to checking if Aave collateral is an option if there is no fast liquidity available
- Should be abstracted from the user
- Positions must be closed when funds are bridged

# Proposed Solution

![image](https://user-images.githubusercontent.com/12769011/161838906-cd546247-644f-4503-b03e-e596a735294c.png)

The changes should be isolated to the `execute` and `reconcile` function calls.

In `execute` if there is no fast liquidity available, the function should check a registry to see if (1) there is a corresponding `aToken` available on the chain and (2) if the given router has permissions to use Aave collateral. If both of these are true, then the router can call `mintUnbacked` to get an uncollateralized credit line and deliver the adopted asset to the user. Otherwise, the function reverts.

Once the mad assets are delivered by nomad, the contract should swap into the correct adopted asset and repay the loan + interest.

A step-by-step breakdown of this proposal is:

**`execute`**
1. Check to make sure there is sufficient liquidity
2. If there is, proceed with the normal transfer flow
3. If there is not, check to see if the router can support Aave liquidity. If the router is not approved for this feature, revert
4. Check to see if there is a corresponding `aToken` for the desired adopted asset. If there is not, revert
[...]

PR #1074, opened 2022-04-29
Aave Portals execute and reconcile logic

patitonar:
## The Problem
Closes https://github.com/connext/nxtp/issues/1059

## The Solution

Add the required logic for the Execute and reconcile methods. Includes some refactors introducing new structs to avoid `stack too deep` errors.

One thing I noticed while implementing this is that the `ConnextTokenRegistry` is actually not needed. All the interactions with Aave are done with only the Pool contract, we don't need to know the addresses of the aTokens. All parameters are specified in the underlying asset. If an adopted asset is not supported in Aave it will revert when trying to mintUnbacked.


packages/deployments/contracts/contracts/interfaces/IAavePool.sol
```diff
@@ -0,0 +1,48 @@
+// SPDX-License-Identifier: UNLICENSED
```
LayneHaber, line 1: is there a `wrap` / `unwrap` function here we can add?

packages/deployments/contracts/contracts/interfaces/IAavePool.sol
```diff
@@ -0,0 +1,48 @@
...
+pragma solidity ^0.8.11;
+
+interface IAavePool {
```
LayneHaber, line 4: do aave pools support wrapping and unwrapping?

packages/deployments/contracts/contracts/lib/Connext/ConnextLogic.sol
```diff
@@ -17,14 +18,16 @@ import {TypeCasts} from "../../nomad-core/contracts/XAppConnectionManager.sol";
...
-import {SafeERC20Upgradeable, AddressUpgradeable} from "@openzeppelin/contracts-upgradeable/token/ERC20/utils/SafeERC20Upgradeable.sol";
+import {SafeERC20Upgradeable, IERC20Upgradeable, AddressUpgradeable} from "@openzeppelin/contracts-upgradeable/token/ERC20/utils/SafeERC20Upgradeable.sol";
 
 library ConnextLogic {
   // ============ Libraries ============
   using TypedMemView for bytes;
   using TypedMemView for bytes29;
   using ConnextMessage for bytes29;
 
+  // TODO Get Aave referral code for Connext
```
LayneHaber, line 29: can we turn this into an issue?

packages/deployments/contracts/contracts/lib/Connext/ConnextLogic.sol
```diff
@@ -71,6 +75,24 @@ library ConnextLogic {
...
+    uint256 portalFeeDenominator;
+  }
+
+  struct ReconcilePortalVars {
+    uint256 portalTransferAmount;
+    uint256 adoptedAmountOut;
+    address adopted;
+    uint256 totalRepayAmount;
+    uint256 backUnbackedAmount;
+    uint256 portalFee;
```
LayneHaber, line 93: isn't this set by aave? is there no way to pull it from there instead of specifying it in the args?

packages/deployments/contracts/contracts/lib/Connext/ConnextLogic.sol
```diff
@@ -864,6 +878,177 @@ library ConnextLogic {
...
+
+    // Edge case with some tokens: Example USDT in ETH Mainnet, after the backUnbacked call there could be a remaining allowance if not the whole amount is pulled by aave.
+    // Later, if we try to increase the allowance it will fail. USDT demands if allowance is not 0, it has to be set to 0 first.
+    // TODO: Should we call approve(0) and approve(totalRepayAmount) instead? or with a try catch to not affect gas on all cases?
+    // Example: https://github.com/aave/aave-v3-periphery/blob/ca184e5278bcbc10d28c3dbbc604041d7cfac50b/contracts/adapters/paraswap/ParaSwapRepayAdapter.sol#L138-L140
+    SafeERC20Upgradeable.safeIncreaseAllowance(IERC20Upgradeable(vars.adopted), _args.aavePool, vars.totalRepayAmount);
+
+    // TODO: What if backUnbacked call fails? For example if the asset is inactive, paused or frozen?
+    // What is the impact of a reconcile to be reverted? Can be re-executed later?
+    IAavePool(_args.aavePool).backUnbacked(vars.adopted, vars.backUnbackedAmount, vars.portalFee);
```
LayneHaber, line 992: hmm if a message is handled (i.e. reconciled is called), it doesn't get rehandled by nomad and can't really be re-processed easily (i will have to double check this to be sure). we should probably have a separate method to repay the loans in case this fails (or if there is a problem with nomad so you can prevent getting into the hole by not closing out a position)

packages/deployments/contracts/contracts/lib/Connext/ConnextLogic.sol
```diff
@@ -864,6 +878,177 @@ library ConnextLogic {
...
+
+    // TODO should emit some event with the repayment info?
```
LayneHaber, line 994: yes :)

packages/deployments/contracts/contracts/lib/Connext/ConnextLogic.sol
```diff
@@ -864,6 +878,177 @@ library ConnextLogic {
...
+
+    // TODO: Do we need to check balance before and after to get the exact amount paid and give the routers the rest?
+    // Aave accounts a global unbacked variable per asset for all, not by address/bridge.
+    // Someone can repay more than it should, so then a the moment of calling backUnbacked()
+    // aave can pull a smaller amount than backUnbackedAmount. So there will be an extra amount of assets that needs to be assigned
+    // See https://github.com/aave/aave-v3-core/blob/feb3f20885c73025f40cc272b59e7eacfaa02fe4/contracts/protocol/libraries/logic/BridgeLogic.sol#L121
```
LayneHaber, line 1000: if we think this will be more than dust, we should probably assign it. however, its not clear to me where the excess would come from (i.e. who repaid more than they should). 

open to just distributing to all the routers though and leaving it as luck of the draw

packages/deployments/contracts/contracts/lib/Connext/ConnextLogic.sol
```diff
@@ -864,6 +878,177 @@ library ConnextLogic {
...
+      _token,
+      _amount
+    );
+
+    // Calculates the amount to be repaid to the portal
+    (vars.totalRepayAmount, vars.backUnbackedAmount, vars.portalFee) = _calculatePortalRepayment(
+      vars.adoptedAmountOut,
+      vars.portalTransferAmount,
+      _args.portalFeeNumerator,
+      _args.portalFeeDenominator
```
LayneHaber, line 981: where do these get assigned?

packages/deployments/contracts/contracts/lib/Connext/ConnextLogic.sol
```diff
@@ -864,6 +878,177 @@ library ConnextLogic {
...
+   * @dev Assumes the fee is proportional to the unbackedAmount.
+   * @return The total amount to be repaid
+   * @return The unbacked amount to be backed
+   * @return The fee amount to be paid
+   */
+  function _calculatePortalRepayment(
+    uint256 availableAmount,
+    uint256 portalTransferAmount,
+    uint256 portalFeeNumerator,
+    uint256 portalFeeDenominator
```
LayneHaber, line 1018: where are these assigned? ideally they should be pulling from the pool itself

packages/deployments/contracts/contracts/lib/Connext/ConnextLogic.sol
```diff
@@ -864,6 +878,177 @@ library ConnextLogic {
...
+        portalFee = availableAmount - backUnbackedAmount;
+      } else {
+        // Repay a partial amount of the transfer and no fees
+        backUnbackedAmount = availableAmount;
+        portalFee = 0;
+      }
+
+      totalRepayAmount = backUnbackedAmount + portalFee;
+
+      // TODO should emit an event to notify we are paying less than we should?
```
LayneHaber, line 1046: yes :)

packages/deployments/contracts/contracts/lib/Connext/ConnextLogic.sol
```diff
@@ -909,34 +1094,62 @@ library ConnextLogic {
...
+      // If router does not have enough liquidity, try to use Aave Portals
+      // TODO: check if this is the correct way to do this. Assumes 1 router
+      // Other option: In case of several routers, if one of them doesn't have enough liquidity,
+      // any of them would provide liquidity at all if taken from Aave Portals
+      if (
+        vars.pathLen == 1 &&
+        _routerBalances[_args.executeArgs.routers[0]][_args.executeArgs.local] < vars.fastTransferAmount &&
+        _args.aavePool != address(0)
+      ) {
+        // TODO: check if this is the correct way to do this. Assumes 1 router
```
LayneHaber, line 1129: yes -- if you're using portals only one router should take on the credit risk

packages/deployments/contracts/contracts/lib/Connext/ConnextLogic.sol
```diff
@@ -948,10 +1161,42 @@ library ConnextLogic {
...
+      _canonicalToAdopted,
+      _adoptedToLocalPools,
+      _args.tokenRegistry,
+      _args.executeArgs.local,
+      _fastTransferAmount
+    );
+
+    IAavePool(_args.aavePool).mintUnbacked(adopted, userAmount, address(this), AAVE_REFERRAL_CODE);
+
+    // Improvement: Instead of withdrawing to address(this), withdraw directly to the user or executor to save 1 transfer
```
LayneHaber, line 1190: 🤤

packages/deployments/contracts/contracts/lib/Connext/ConnextLogic.sol
```diff
@@ -864,6 +878,177 @@ library ConnextLogic {
...
+    // try to repay as much as unbacked as possible
+    if (totalRepayAmount > availableAmount) {
+      if (availableAmount > backUnbackedAmount) {
+        // Repay the whole transfer and a partial amount of fees
+        portalFee = availableAmount - backUnbackedAmount;
+      } else {
+        // Repay a partial amount of the transfer and no fees
+        backUnbackedAmount = availableAmount;
+        portalFee = 0;
+      }
```
LayneHaber, line 1042: will need to make sure we can pay as a router separately from the reconcile flow

also, can you add some comments around the asset that these numbers correspond to?

packages/deployments/contracts/contracts/nomad-xapps/contracts/connext/ConnextHandler.sol
```diff
@@ -294,6 +311,41 @@ contract ConnextHandler is
...
+
+  /**
+   * @notice Sets the Aave Portal fee numerator
+   * @param _aavePortalFeeNumerator The new value for the Aave Portal fee numerator
+   */
+  function setAavePortalFee(uint256 _aavePortalFeeNumerator) external onlyOwner {
+    if (_aavePortalFeeNumerator > LIQUIDITY_FEE_DENOMINATOR) revert ConnextHandler__setAavePortalFee_invalidFee();
+
+    aavePortalFeeNumerator = _aavePortalFeeNumerator;
+  }
```
LayneHaber, line 331: ideally we could pull this from aave directly. also, i think the fees differ per-pool?

packages/deployments/contracts/contracts/interfaces/IAavePool.sol
patitonar: Are you referring to wrapping/unwrapping native assets? At the protocol level, Aave only handles ERC20. They have[ this helper contract](https://github.com/aave/aave-v3-periphery/blob/ca184e5278bcbc10d28c3dbbc604041d7cfac50b/contracts/misc/WETHGateway.sol#L42-L49) to allow users to wrap/unwrap and enter/exit the pool in one transaction

packages/deployments/contracts/contracts/lib/Connext/ConnextLogic.sol
patitonar: Created the following issue https://github.com/connext/nxtp/issues/1084

packages/deployments/contracts/contracts/lib/Connext/ConnextLogic.sol
patitonar: Oh, I see. Then, I think it is important to correctly handle any external call failure to not revert the reconcile call and have a separate method to repay in case of any of these mentioned issues with Aave or Nomad.

packages/deployments/contracts/contracts/lib/Connext/ConnextLogic.sol
patitonar: > its not clear to me where the excess would come from (i.e. who repaid more than they should).

It could be any of our routers for example due to a miscalculation in a manual repayment on the separate method or could be any other bridge solution that is also using Aave Portals

packages/deployments/contracts/contracts/lib/Connext/ConnextLogic.sol
patitonar: These get assigned in the handle method before calling ConnextLogic here https://github.com/connext/nxtp/blob/71e7d7af0ac2111779c92dcbc33e64710a66528c/packages/deployments/contracts/contracts/nomad-xapps/contracts/connext/ConnextHandler.sol#L480-L481

packages/deployments/contracts/contracts/nomad-xapps/contracts/connext/ConnextHandler.sol
patitonar: I'm using this comment for all the comments related to fees.

Aave has only one Pool contract which is the entry point for the protocol which handles all the assets. 
In the contracts, there is no information about the portal fees, only the `BRIDGE_PROTOCOL_FEE()` which is used to [calculate](https://github.com/aave/aave-v3-core/blob/e46341caf815edc268893f4f9398035f242375d9/contracts/protocol/libraries/logic/BridgeLogic.sol#L123) the proportion of the paid fees that will be stored for the protocol instead of the Aave users. But there is no stored information about how to calculate the fees for each asset and/or for each bridge using Portals. That's why I created the `aavePortalFeeNumerator` variable.

I found this template for Portal whitelist https://governance.aave.com/t/arc-portals-governance-framework/7681
It looks like each bridge can propose a specific fee model for each asset they will use from Portals. It looks like in this V0 of Portals fees works more like a social/off-chain agreement than enforced via smart contracts. So, depending in our proposal, if we are going to use different pricing models and/or percentages on different assets, instead of storing one variable `aavePortalFeeNumerator` we may need to have a mapping relating the adopted asset a the feeNumerator.

It would be great to confirm all of this with someone from Aave.

packages/deployments/contracts/contracts/lib/Connext/ConnextLogic.sol
patitonar: Answered in this other comment https://github.com/connext/nxtp/pull/1074#discussion_r863933180

packages/deployments/contracts/contracts/lib/Connext/ConnextLogic.sol
patitonar: Added an event

packages/deployments/contracts/contracts/lib/Connext/ConnextLogic.sol
patitonar: Updated the docs and added a new method `repayAavePortal` to allow a router to manually repay

packages/deployments/contracts/contracts/lib/Connext/ConnextLogic.sol
patitonar: I thought about adding this, but it introduces some complexity in the code, and also it is only an improvement if the asset is not a native asset that needs to be unwrapped (so it also requires additional checks). So I won't include this for now

packages/deployments/contracts/contracts/interfaces/IAavePool.sol
LayneHaber: basically want a way to get the underlying asset instead of `aTokens` so you dont have to go through an AMM and the user doesn't need to know a loan was taken out. is there a way to do that?

packages/deployments/contracts/contracts/nomad-xapps/contracts/connext/ConnextHandler.sol
LayneHaber: i think you're right -- will confirm

packages/deployments/contracts/contracts/lib/Connext/ConnextLogic.sol
LayneHaber: it seems like we can discount this then, i dont see it being likely more than dust and im not sure its worth the added complexity

LayneHaber (approved):

packages/deployments/contracts/contracts/core/connext/facets/RoutersFacet.sol
```diff
@@ -319,6 +344,33 @@ contract RoutersFacet is BaseConnextFacet {
     emit LiquidityFeeNumeratorUpdated(_numerator, msg.sender);
   }
 
+  /**
+   * @notice Allow router to use Portals
+   * @param _router - The router address to approve
+   */
+  function approveRouterForPortal(address _router) external {
```
jakekidd, line 351: why is this not `onlyOwner`?

Is this intended to let any router approve themselves (in which case, this should be permissioned to onlyRouterOwner)?

packages/deployments/contracts/contracts/core/connext/facets/RoutersFacet.sol
```diff
@@ -319,6 +344,33 @@ contract RoutersFacet is BaseConnextFacet {
...
+    s.routerPermissionInfo.approvedForPortalRouters[_router] = true;
+
+    emit RouterApprovedForPortal(_router, msg.sender);
+  }
+
+  /**
+   * @notice Remove router access to use Portals
+   * @param _router - The router address to remove approval
+   */
+  function disapproveRouterForPortal(address _router) external {
```
jakekidd, line 365: 1. nit: should be `unapproveRouterForPortal` I think?

2. doesn't necessarily *need* to be onlyOwner so long as it's onlyRouterOwner.... buuuut, 1 huge problem here: if router owes debt to aave portal, and they are unapproved/disapproved they will be unable to pay back that debt via `PortalFacet.repayAavePortal`. I imagine that has some pretty bad consequences.

So first, **permission this pls**

next, either one of two things must happen:
- Can't unapprove routers with outstanding debt (not a fan)
- Remove the check to make sure router is approved for portals in `PortalFacet.repayAavePortal` (am a fan)

packages/deployments/contracts/contracts/core/connext/facets/PortalFacet.sol
```diff
@@ -0,0 +1,98 @@
...
+  /**
+   * @notice Used by routers to perform a manual repayment to Aave Portals to cover any outstanding debt
+   * @dev The router must be approved for Portal and with enough liquidity
+   */
+  function repayAavePortal(
+    address _local,
+    uint256 _backingAmount,
+    uint256 _feeAmount
+  ) external {
+    if (!s.routerPermissionInfo.approvedForPortalRouters[msg.sender])
```
jakekidd, line 70: Please see previous comments on permissioning unapproval for portal routers. Personally, I see no reason for this permissions check to exist here. Anyone should be able to pay back their debt; and in fact, why not allow anyone to pay back any routers debt? like a `repayAavePortalFor` method?

packages/deployments/contracts/contracts/core/connext/facets/PortalFacet.sol
```diff
@@ -0,0 +1,98 @@
...
+
+    // Need to swap into adopted asset or asset that was backing the loan
+    // The router will always be holding collateral in the local asset while the loaned asset
+    // is the adopted asset
+    (uint256 balanceInAdopted, address adopted) = AssetLogic.calculateSwapFromLocalAssetIfNeeded(_local, routerBalance);
+
+    if (balanceInAdopted < totalAmount) revert PortalFacet__repayAavePortal_insufficientFunds();
+
+    // Swap for exact `totalRepayAmount` of adopted asset to repay aave
+    (uint256 amountIn, ) = AssetLogic.swapFromLocalAssetIfNeededForExactOut(_local, totalAmount);
```
jakekidd, line 84: is it safe to not have any slippage maximums here? i.e. if `amountIn` goes over X, revert. to prevent someone from frontrunning/sandwiching this stableswap action

packages/deployments/contracts/contracts/core/connext/facets/RoutersFacet.sol
```diff
@@ -222,6 +239,14 @@ contract RoutersFacet is BaseConnextFacet {
     return s.routerBalances[_router][_asset];
   }
 
+  /**
+   * @notice Returns wether the router is approved for portals or not
```
jakekidd, line 243: nit: `wether` => `whether` lol

packages/deployments/contracts/contracts/core/connext/libraries/AssetLogic.sol
```diff
@@ -228,7 +252,114 @@ library AssetLogic {
...
+
+    uint8 tokenIndexIn = getTokenIndexFromStableSwapPool(_canonicalId, _assetIn);
+    uint8 tokenIndexOut = getTokenIndexFromStableSwapPool(_canonicalId, _assetOut);
+    if (stableSwapPoolExist(_canonicalId)) {
+      // if internal swap pool exists
+      return (s.swapStorages[_canonicalId].swapInternalOut(tokenIndexIn, tokenIndexOut, _amountOut, 0), _assetOut);
+    } else {
+      // Otherwise, swap via stable swap pool
+      IStableSwap pool = s.adoptedToLocalPools[_canonicalId];
+      uint256 _amountIn = pool.calculateSwapOut(tokenIndexIn, tokenIndexOut, _amountOut);
```
jakekidd, line 287: this is essentially the same thing as having *no* maximum amount in!

This `_swapAssetOut` is essentially a `swapExactOut` function. If the slippage and/or ratio is off-balance sufficiently, I could be getting a very bad deal here, and the only cap to my risk is however much funds I have available in the first place (bc obvi I cannot transfer more tokens than I own). This invites front-running/sandwiching to basically extract all of my tokens, so long as I own < the pool size.

jakekidd: I know this method is `internal` ofc, but anytime we are doing an exact-out-swap, we must have a route to exert control over the maximum amountIn to some degree

packages/deployments/contracts/contracts/core/connext/interfaces/IStableSwap.sol
```diff
@@ -40,7 +40,15 @@ interface IStableSwap {
   function swapExact(
     uint256 amountIn,
     address assetIn,
-    address assetOut
+    address assetOut,
+    uint256 minAmountOut
```
jakekidd, line 44: Hmm, maybe we should rename this method to `swapExactIn` to distinguish it from `swapExactOut`? Thoughts?

packages/deployments/contracts/contracts/core/connext/facets/BridgeFacet.sol
```diff
@@ -696,4 +708,220 @@ contract BridgeFacet is BaseConnextFacet {
...
+  }
+
+  /**
+   * @notice Parses the message and process the transfer
+   * @dev Will mint the tokens if the token originates on a remote origin
+   * @return The message amount
+   * @return The message token
+   * @return The message transfer id
+   */
+  function _reconcileProcessMessage(bytes memory _message)
```
jakekidd, line 743: Personally i'm not a fan of these "one-time-use" extension functions that we're using to avoid stack too deep errors. Would strongly prefer to find a way to scope this within the `_reconcile` method itself.

8 variables are declared here, and 7 are declared in the parent function including args - which keeps us just under the ref stack size limit of 16/

If we want (or need) to reduce it further, you could easily just scope this method's logic in a block, and keep only the 4 'return' variables `(uint256 amount, address local, bytes32 transferId, bytes32 canonical)` outside of the scope. At 11 vars, should be good.

cc @LayneHaber in case you have any thoughts on this. This is certainly IMHO, but i'm pushing to avoid this 'extension' pattern where possible

packages/deployments/contracts/contracts/core/connext/facets/BridgeFacet.sol
```diff
@@ -696,4 +708,220 @@ contract BridgeFacet is BaseConnextFacet {
...
+    // assert the action is valid
+    if (!action.isTransfer()) {
+      revert BridgeFacet__reconcile_invalidAction();
+    }
+
+    // load the transferId
+    bytes32 transferId = action.transferId();
+
+    // ensure the transaction has not been handled
+    if (s.reconciledTransfers[transferId]) {
```
jakekidd, line 766: Ik you did not write this code @liu-zhipeng but please move this check to be the very first thing in the method. Just in case there's a race to reconcile, somehow, we want the revert on the redundant tx to cost as little as possible

packages/deployments/contracts/contracts/core/connext/facets/BridgeFacet.sol
jakekidd: Just noticed we have this `_reconcileProcessPortal` method as well, which has 6 variables. That should also fit within the `_reconcile` method as long as we scope that method's logic entirely.

packages/deployments/contracts/contracts/core/connext/facets/BridgeFacet.sol
```diff
@@ -696,4 +708,220 @@ contract BridgeFacet is BaseConnextFacet {
...
+      emit AavePortalRepaymentDebt(_transferId, adopted, backUnbackedAmount, portalFee);
+    }
+
+    // TODO: Do we need to check balance before and after to get the exact amount paid and give the routers the rest?
+    // Aave accounts a global unbacked variable per asset for all, not by address/bridge.
+    // Someone can repay more than it should, so then a the moment of calling backUnbacked()
+    // aave can pull a smaller amount than backUnbackedAmount. So there will be an extra amount of assets that needs to be assigned
+    // See https://github.com/aave/aave-v3-core/blob/feb3f20885c73025f40cc272b59e7eacfaa02fe4/contracts/protocol/libraries/logic/BridgeLogic.sol#L121
+
+    // TODO: need to credit the router in the local asset, not the adopted asset
```
jakekidd, line 860: Is this TODO a blocker to merge this PR?

packages/deployments/contracts/contracts/core/connext/facets/BridgeFacet.sol
```diff
@@ -696,4 +708,220 @@ contract BridgeFacet is BaseConnextFacet {
...
+    } else {
+      // Update the amount repaid to 0, so the amount is credited to the router
+      amountIn = 0;
+      emit AavePortalRepaymentDebt(_transferId, adopted, backUnbackedAmount, portalFee);
+    }
+
+    // TODO: Do we need to check balance before and after to get the exact amount paid and give the routers the rest?
+    // Aave accounts a global unbacked variable per asset for all, not by address/bridge.
+    // Someone can repay more than it should, so then a the moment of calling backUnbacked()
+    // aave can pull a smaller amount than backUnbackedAmount. So there will be an extra amount of assets that needs to be assigned
```
jakekidd, line 857: I don't understand - is this comment referring to dust left over from repayment math?

packages/deployments/contracts/contracts/core/connext/facets/BridgeFacet.sol
jakekidd: oh wait there's also `_calculatePortalRepayment` LOL

ok i guess this is nbd for now. still think you could probs fit all of it in 1 method but nbd

packages/deployments/contracts/contracts/core/connext/facets/BridgeFacet.sol
```diff
@@ -421,86 +456,51 @@ contract BridgeFacet is BaseConnextFacet {
...
-    // - be reserved for execution (slow liquidity)
-    if (!s.tokenRegistry.isLocalOrigin(token)) {
-      IBridgeToken(token).mint(address(this), amount);
-      // Tell the token what its detailsHash is
-      IBridgeToken(token).setDetailsHash(details);
+    uint256 routersAmount = amount;
+    uint256 pathLen = routers.length;
+    if (portalTransferAmount != 0) {
+      // ensure a router took on credit risk
+      if (pathLen != 1) revert BridgeFacet__reconcile_noPortalRouter();
```
jakekidd, line 487: what if it's multiple routers? shouldn't this be reverting if pathLength == 0?

packages/deployments/contracts/contracts/core/connext/facets/BridgeFacet.sol
```diff
@@ -421,86 +456,51 @@ contract BridgeFacet is BaseConnextFacet {
...
+      routersAmount = _reconcileProcessPortal(amount, local, routers[0], transferId, canonical);
```
jakekidd, line 488: we're only processing for router[0]? wb other ones? i thought multipath using aave portal was possible

packages/deployments/contracts/contracts/core/connext/facets/RoutersFacet.sol
liu-zhipeng: It should be `onlyOwner`!

packages/deployments/contracts/contracts/core/connext/facets/PortalFacet.sol
liu-zhipeng: We will on `AssetLogic.sol`.

[https://github.com/connext/nxtp/issues/1227](url)

packages/deployments/contracts/contracts/core/connext/facets/BridgeFacet.sol
LayneHaber: This is resolved -- this is what @liu-zhipeng was working on

packages/deployments/contracts/contracts/core/connext/facets/BridgeFacet.sol
LayneHaber: the aave `backUnbacked` function may not take the entire `backUnbacked` amount you supply, and the remainder is on our contract not theirs. Important  to note

packages/deployments/contracts/contracts/core/connext/facets/BridgeFacet.sol
LayneHaber: hm? you only want to allow one router to take on the credit risk. since they're not supplying the liquidity you dont really need to multipath it.

packages/deployments/contracts/contracts/core/connext/facets/BridgeFacet.sol
LayneHaber: aave portals should only use a single router to simplify who holds the credit risk (also the router is trusted)

packages/deployments/contracts/contracts/core/connext/facets/BridgeFacet.sol
```diff
@@ -599,37 +599,49 @@ contract BridgeFacet is BaseConnextFacet {
...
-          i++;
+      // If router does not have enough liquidity, try to use Aave Portals
+      if (
+        pathLen == 1 && s.routerBalances[_args.routers[0]][_args.local] < fastTransferAmount && s.aavePool != address(0)
+      ) {
+        if (!s.routerPermissionInfo.approvedForPortalRouters[_args.routers[0]])
+          revert BridgeFacet__execute_notApprovedForPortals();
+
+        // Portal provides the adopted asset so we early return here
+        return _executePortalTransfer(_transferId, fastTransferAmount, _args.local, _args.routers[0]);
```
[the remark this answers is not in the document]
LayneHaber: hmm that is kind of an annoying one. you're basically introducing slippage you would have by allowing the aave portal to be used in a local transfer.

im not sure we should allow for that (i.e. revert intentionally), but curious to hear what @liu-zhipeng @rhlsthrm think

packages/deployments/contracts/contracts/core/connext/facets/PortalFacet.sol
LayneHaber: Yeah, all slippage enforcing stuff will go into #1320 (since it breaks the call params interface)

packages/deployments/contracts/contracts/core/connext/libraries/AssetLogic.sol
LayneHaber: right -- so i guess this is kind of weird. there's a few constraints we have to solve around:

- This is the only case where a transfer has to perform a swap on behalf of the router instead of the user
- There are no external parameters that can be supplied to enforce slippage
- Slippage cannot be strongly enforced because this is called via `handle`, which will mark the message as handled even if the handling fails (i.e. reconcile reverts because slippage is enforced)

So there's a couple of options here:

1. Use the amount the bridge minted as the default slippage/min received. If it exceeds that, pay as much as you can and emit the debt event
2. Have a global max slippage enforced on the contracts. If your transfer exceeds the slippage, then it would do the same as above.

Additionally, we can add a slippage param to the manual debt repay function that is included in the `PortalFacet`.

packages/deployments/contracts/contracts/core/connext/facets/BridgeFacet.sol
LayneHaber: i also think this could go into a separate PR of just code-styling and best practices clean up

packages/deployments/contracts/contracts_forge/core/connext/facets/BridgeFacet.t.sol
```diff
@@ -239,6 +239,40 @@ contract BridgeFacetTest is BridgeFacet, FacetHelper {
...
+      s.routerBalances[_args.routers[i]][_args.local] = 0 ether;
+    }
+
+    // set aave enabled
+    s.aavePool = _aavePool;
+
+    vm.expectRevert(abi.encodeWithSelector(BridgeFacet.BridgeFacet__execute_notApprovedForPortals.selector));
+    this.execute(_args);
+  }
+
```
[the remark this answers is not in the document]
LayneHaber: there shouldn't be multipath allowed for aave so having only the one router be unapproved should be sufficient

LayneHaber (changes requested):
This can't be merged until handling out-of-band repayments is addressed (i.e. the case where a router calls `repayAavePortal` before the transfer is reconciled)

LayneHaber (approved):

PR #1076, opened 2022-05-02
967 subgraph optimization

sanchaymittal:
## Checklist

- [ ] Run load tests.
- [ ] Update CHANGELOG.md.

### DIFF
--- a/packages/adapters/subgraph/src/lib/subgraphs/runtime/queries.ts
+++ b/packages/adapters/subgraph/src/lib/subgraphs/runtime/queries.ts
@@ -2,244 +2,139 @@ import { gql } from "graphql-request";
 
 // Contains all subgraph queries used by router
 
-export const getTransfers = gql`
-  query GetTransfers($originDomain: BigInt!, $destinationDomains: [BigInt!], $nonce: BigInt!) {
-    transfers(
-      where: { destinationDomain_in: $destinationDomains, nonce_gte: $nonce, originDomain: $originDomain }
-      orderBy: xcalledBlockNumber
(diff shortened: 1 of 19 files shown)


packages/adapters/subgraph/src/lib/subgraphs/runtime/queries.ts
```diff
@@ -250,116 +145,57 @@ export const getExecutedTransfersByIds = gql`
...
 
-export const getReconciledTransfersByIds = gql`
-  query GetReconciledTransfersByIds($transferIds: [Bytes!], $maxReconciledBlockNumber: BigInt!) {
-    transfers(
+export const getDestinationTransfersByIds = gql`
+  query GetDestinationTransfersByIds(
+    $transferIds: [Bytes!]
+    $maxExecutedBlockNumber: BigInt
+    $maxReconciledBlockNumber: BigInt
+    $status: TransferStatus
```
rhlsthrm, line 153: Why specify status if you're using IDs anyways?

rhlsthrm (approved):

packages/adapters/subgraph/src/lib/subgraphs/runtime/queries.ts
sanchaymittal: it's an optional param for functionality which we were using in one of the query

sanchaymittal: yeah, if we don't add `!` as suffix, then it's an optional param. Using this in code so it's all gucci

PR #1093, opened 2022-05-04
Only owner registry

LayneHaber:
## The Problem

Function was mistakenly placed as non-admin

## The Solution

Fix and redeploy

## Checklist

- [ ] Run load tests.
- [ ] Update CHANGELOG.md.


rhlsthrm (approved):
Why did the token registry address change? Do we need to change the staging and prod contracts?

LayneHaber:
> Why did the token registry address change? Do we need to change the staging and prod contracts?

the implementation changed, not the address of the upgradeable proxy

PR #1107, opened 2022-05-06
Choose which subgraph URL to use based on configured environment

jakekidd:


packages/adapters/subgraph/.graphclientrc.yml
```diff
@@ -1,5 +1,16 @@
 sources:
   - name: Connext_Kovan
+    handler:
+      graphql:
+        endpoint: https://api.thegraph.com/subgraphs/name/connext/nxtp-amarok-runtime-v0-kovan
+        retry: 5
+    transforms:
+      - prefix:
+          value: kovan_
```
wanglonghong, line 9: Staging prefix should be different from production.

wanglonghong (changes requested):

PR #1113, opened 2022-05-06
POC Diamond architecture

lmcorbalan:
## The Problem

No more bytecode available

## The Solution

Implement a proxy diamod architecture (https://eips.ethereum.org/EIPS/eip-2535) using [diamond-3 implentation](https://github.com/mudgen/diamond-3-hardhat) 

## Checklist

- [ ] Run load tests.
- [ ] Update CHANGELOG.md.


packages/deployments/contracts/contracts/facets/StableSwapFacet.sol
```diff
@@ -0,0 +1,458 @@
...
+  /**
+   * @notice Modifier to check deadline against current timestamp
+   * @param deadline latest timestamp to accept this transaction
+   */
+  modifier deadlineCheck(uint256 deadline) {
+    if (block.timestamp > deadline) revert StableSwapFacet__deadlineCheck_deadlineNotMet();
+    _;
+  }
+
+  /*** VIEW FUNCTIONS ***/
```
jakekidd, line 61: Lol ik this is such a nit but:
```
// ============ Modifiers ============
```
vs
```
 /*** VIEW FUNCTIONS ***/
 ```
 
 can we make these the same vibe

Comment (jakekidd) [reply packages/deployments/contracts/contracts/facets/StableSwapFacet.sol:61]: e.g.
```
// ========= View Functions =========
```

Comment (jakekidd) [reply packages/deployments/contracts/contracts/facets/StableSwapFacet.sol:61]: actually nvm, let's just come thru later and polish comments and stuff

Comment (liu-zhipeng) [reply packages/deployments/contracts/contracts/facets/StableSwapFacet.sol:61]: Okay ~

File: packages/deployments/contracts/contracts/facets/StableSwapFacet.sol [hunk 2]
```diff
@@ -0,0 +1,458 @@
...
+  /**
+   * @notice Add liquidity to the pool with the given amounts of tokens
+   * @param canonicalId the canonical token id
+   * @param amounts the amounts of each token to add, in their native precision
+   * @param minToMint the minimum LP tokens adding this amount of liquidity
+   * should mint, otherwise revert. Handy for front-running mitigation
+   * @param deadline latest timestamp to accept this transaction
+   * @return amount of LP token user minted and received
+   */
+  function addStableLiquidity(
```
jakekidd, line 259: I feel like even though it's not an ERC standard, we should conform with the same method names as Curve just to make this recognizable.

Like instead of `addStableLiquidity` just stick to `addLiquidity`

Besides, it's not like the function name is going to enforce that the liquidity is stablecoins lol

packages/deployments/contracts/contracts/facets/StableSwapFacet.sol
```diff
@@ -0,0 +1,458 @@
...
+    bytes32 _canonicalId,
+    IERC20[] memory _pooledTokens,
+    uint8[] memory decimals,
+    string memory lpTokenName,
+    string memory lpTokenSymbol,
+    uint256 _a,
+    uint256 _fee,
+    uint256 _adminFee,
+    address lpTokenTargetAddress
+  ) external onlyOwner {
```
jakekidd, line 357: needs ``initializer` modifier from openzeppelins `Initializable`, no?

otherwise, even though this is limited by `onlyOwner`, it can be reinitialized maliciously

packages/deployments/contracts/contracts/facets/StableSwapFacet.sol
liu-zhipeng: Yeah, i love that. `addLiquidity` but we have already such a function on `RoutersFacet`. so i had to think another name.
do you have any idea?

packages/deployments/contracts/contracts/facets/StableSwapFacet.sol
```diff
@@ -0,0 +1,458 @@
...
+  }
+
+  /**
+   * @notice Update the swap fee to be applied on swaps
+   * @param canonicalId the canonical token id
+   * @param newSwapFee new swap fee to be applied on future transactions
+   */
+  function setSwapFee(bytes32 canonicalId, uint256 newSwapFee) external onlyOwner {
+    s.swapStorages[canonicalId].setSwapFee(newSwapFee);
+  }
```
jakekidd, line 433: I'm assuming that these fee setter methods will be delegated to be only governance in the future

fee changes should be subject to a commitment period as well (like submit the change, and then commit it 72 hours later)

just something to keep in mind

packages/deployments/contracts/contracts/facets/StableSwapFacet.sol
liu-zhipeng: `initializer` modifier is for only `Initializable`. 

to prevent that, we have this.
`if (s.swapStorages[_canonicalId].pooledTokens.length != 0)
      revert StableSwapFacet__initializeSwap_alreadyInitialized();`

packages/deployments/contracts/contracts/facets/StableSwapFacet.sol
jakekidd: ah, so if that's the issue, we should probs just call it `addStableSwapLiquidity` lol

jakekidd: if there's really an overlap... maybe we should be calling `RoutersFacet.addLiquidity` `RoutersFacet.addRouterLiquidity` lol

but dwbi for rn

packages/deployments/contracts/contracts/facets/StableSwapFacet.sol
jakekidd: oh yeah, just realized the structure is different from typical stableswap in that this contract manages many pools indexed by canonical id

sounds good

packages/deployments/contracts/contracts/facets/StableSwapFacet.sol
```diff
@@ -0,0 +1,458 @@
...
+   */
+  function initializeStableSwap(
+    bytes32 _canonicalId,
+    IERC20[] memory _pooledTokens,
+    uint8[] memory decimals,
+    string memory lpTokenName,
+    string memory lpTokenSymbol,
+    uint256 _a,
+    uint256 _fee,
+    uint256 _adminFee,
```
jakekidd, line 355: admin fee should be global no? like, it should be the same value for every swap in this contract. that's typically how it works - like a protocol-wide fee, right?

we could leave this for now, but we may want to make that adminFee a separate storage item.

packages/deployments/contracts/contracts/interfaces/IConnextHandler.sol
```diff
@@ -86,51 +164,128 @@ interface IConnextHandler {
...
-    ConnextMessage.TokenId calldata canonical,
-    address adoptedAssetId,
-    address stableSwapPool
+  function addLiquidityFor(
+    uint256 _amount,
+    address _local,
+    address _router
+  ) external payable;
+
+  function addLiquidity(uint256 _amount, address _local) external payable;
```
jakekidd, line 175: From an earlier comment : we may want to change this to addRouterLiquidity and addRouterLiquidityFor and removeRouterLiquidity

to avoid confusion between LPing for router stuff from LPing for stableswap

packages/deployments/contracts/contracts/facets/StableSwapFacet.sol
```diff
@@ -0,0 +1,458 @@
...
+    s.swapStorages[canonicalId].withdrawAdminFees(msg.sender);
+  }
+
+  /**
+   * @notice Update the admin fee. Admin fee takes portion of the swap fee.
+   * @param canonicalId the canonical token id
+   * @param newAdminFee new admin fee to be applied on future transactions
+   */
+  function setAdminFee(bytes32 canonicalId, uint256 newAdminFee) external onlyOwner {
+    s.swapStorages[canonicalId].setAdminFee(newAdminFee);
```
jakekidd, line 423: imo, should have a MAX_ADMIN_FEE and a MAX_SWAP_FEE in place and require that this `newAdminFee` is always <= to it.

MAX_SWAP_FEE should be like 1% or 100 basis points maybe (so 10**8?)

Admin fee technically does not take any fee on top of swap fee, it takes a % of the swap fee iirc. So it should be capped at like, 50%? Does that sound right?

Just want to add rug protection for users here in case of unlikely event owner is compromised lol

PR #1128, opened 2022-05-10
StableSwapFacet

liu-zhipeng:
## Checklist

- [ ] Run load tests.
- [ ] Update CHANGELOG.md.


packages/deployments/contracts/contracts/diamond/facets/StableSwapFacet.sol
```diff
@@ -0,0 +1,457 @@
...
+   * eg 8 for WBTC. Cannot be larger than POOL_PRECISION_DECIMALS
+   * @param lpTokenName the long-form name of the token to be deployed
+   * @param lpTokenSymbol the short symbol for the token to be deployed
+   * @param _a the amplification coefficient * n * (n - 1). See the
+   * StableSwap paper for details
+   * @param _fee default swap fee to be initialized with
+   * @param _adminFee default adminFee to be initialized with
+   * @param lpTokenTargetAddress the address of an existing LPToken contract to use as a target
+   */
+  function _initializeSwap(
```
lmcorbalan, line 80: I think this logic needs to be moved to `DiamondInit.init` function

patitonar: I'm not sure exactly how it is supposed to work, but my guess is that there will be multiple initializations, one per each stableswap instance created? So probably this `_initializeSwap` should be called from an external method that allows to create a new/initialize a new stableSwap?

liu-zhipeng: > 

yes, exactly, initializeSwap is to init a swap for a canonical token.
I will fix

Issue #1007, opened 2022-04-19
[router] Do not execute if Nomad is down


rhlsthrm:
Check Nomad SDK:

> If you run 'checkHomes' on the SDK and provide it a list of networks, then you can access 'blacklist' value which is a set of Failed homes

alexwhte:
@rhlsthrm  Do we need this error handling in Connext Bridge UI?

rhlsthrm:
> @rhlsthrm Do we need error handling for this in Connext Bridge UI?
> 
> edit: nvm, commented in master #990

This needs a matching task in the SDK because the SDK should also implement this functionality.

base0010:
After looking at the nomad sdk not convinced that checkHomes is going to give us the correct state of if the destination chain is processing nominally. Needs clarification, but a shared utility that keeps track of the latest block hash of a updater committed root. Then we can have simple logic to compare that to block hash of now. If there is a significant n block lag (tbd what this measure exactly should be and if it needs be dynamic) we know the destination chain is down.

PR #1130, opened 2022-05-10
Sanity check #1007

base0010:
## Checklist

- [ ] Run load tests.
- [ ] Update CHANGELOG.md.


packages/agents/router/src/lib/operations/execute.ts
```diff
@@ -7,12 +7,29 @@ import {
...
+import { BridgeContext } from "@nomad-xyz/sdk-bridge";
+
 // fee percentage paid to relayer. need to be updated later
 export const RELAYER_FEE_PERCENTAGE = "1"; //  1%
+//helper function to match our config environments with nomads
+const nxtpEnvToNomadEnv = (envOverride?: string): string => {
+  const nxtpConfig = envOverride ? envOverride : getContext().config.environment;
+  switch (nxtpConfig) {
+    case "production":
+      return "production";
```
rhlsthrm, line 24: These don't map directly to our env. We use production for testnets also. We need to have a separate config item for `nomadEnv`.

rhlsthrm (changes requested):

packages/agents/router/src/config.ts
```diff
@@ -84,6 +84,7 @@ export const NxtpRouterConfigSchema = Type.Object({
   polling: TPollingConfig,
   subgraphPrefix: Type.Optional(Type.String()),
   environment: Type.Union([Type.Literal("staging"), Type.Literal("production")]),
+  nomad_environment: Type.Union([Type.Literal("staging"), Type.Literal("production")]),
```
rhlsthrm, line 87: Can we make this camel case?

rhlsthrm (changes requested):

packages/agents/router/src/config.ts
rhlsthrm: Also we need to add this config item to the below section where it parses it from env. And we need to add the correct mocks for it. Also need to add the config item to the terraform files inside `ops/` to make the live agents use the proper vars.

rhlsthrm (changes requested):

PR #1141, opened 2022-05-10
Update Connext hardhat tests to work with diamond proxy

patitonar:
## The Solution

Update hardhat tests and also some minor fixes on the contracts to make the diamond facets work together


packages/deployments/contracts/contracts/diamond/facets/BridgeFacet.sol
```diff
@@ -191,7 +191,7 @@ contract BridgeFacet is BaseConnextFacet {
    * @dev Can be called prior to or after `handle`, depending if fast liquidity is being
    * used.
    */
-  function execute(ExecuteArgs calldata _args) external returns (bytes32 transferId) {
+  function execute(ExecuteArgs calldata _args) external returns (bytes32) {
```
lmcorbalan, line 194: I think we need to fix the `return` statement also

PR #1224, opened 2022-05-23
Multipath support

wanglonghong:
## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes

### DIFF
--- a/packages/agents/router/src/config.ts
+++ b/packages/agents/router/src/config.ts
@@ -15,6 +15,7 @@ const DEFAULT_SUBGRAPH_POLL_INTERVAL = 15_000;
 const DEFAULT_CONFIRMATIONS = 3;
 const MIN_CACHE_POLL_INTERVAL = 2_000;
 const DEFAULT_CACHE_POLL_INTERVAL = 20_000;
+const DEFAULT_AUCTION_ROUND_DEPTH = 4;
 
 dotenvConfig();
 
@@ -82,6 +83,7 @@ export const NxtpRouterConfigSchema = Type.Object({
   mode: TModeConfig,
   network: Type.Union([Type.Literal("testnet"), Type.Literal("mainnet"), Type.Literal("local")]),
   polling: TPollingConfig,
+  auctionRoundDepth: Type.Number(),
   environment: Type.Union([Type.Literal("staging"), Type.Literal("production")]),
 });
 
@@ -173,6 +175,11 @@ export const getEnvConfig = (
         configFile.polling?.cach ||
         DEFAULT_CACHE_POLL_INTERVAL,
     },
+    auctionRoundDepth:
+      process.env.AUCTION_ROUND_DEPTH ||
+      configJson.auctionRoundDepth ||
+      configFile.auctionRoundDepth ||
+      DEFAULT_AUCTION_ROUND_DEPTH,
     environment: process.env.NXTP_ENVIRONMENT || configJson.environment || configFile.environment || "production",
   };
 
--- a/packages/agents/router/src/config.ts
+++ b/packages/agents/router/src/config.ts
@@ -83,7 +83,7 @@ export const NxtpRouterConfigSchema = Type.Object({
   mode: TModeConfig,
   network: Type.Union([Type.Literal("testnet"), Type.Literal("mainnet"), Type.Literal("local")]),
   polling: TPollingConfig,
-  auctionRoundDepth: Type.Number(),
+  auctionRoundDepth: Type.Optional(Type.Number()),
   subgraphPrefix: Type.Optional(Type.String()),
   environment: Type.Union([Type.Literal("staging"), Type.Literal("production")]),
 });
--- a/packages/agents/router/src/config.ts
+++ b/packages/agents/router/src/config.ts
@@ -83,7 +83,7 @@ export const NxtpRouterConfigSchema = Type.Object({
   mode: TModeConfig,
   network: Type.Union([Type.Literal("testnet"), Type.Literal("mainnet"), Type.Literal("local")]),
   polling: TPollingConfig,
-  auctionRoundDepth: Type.Optional(Type.Number()),
+  auctionRoundDepth: Type.Number(),
   subgraphPrefix: Type.Optional(Type.String()),
   environment: Type.Union([Type.Literal("staging"), Type.Literal("production")]),
 });
--- a/packages/agents/router/src/lib/operations/execute.ts
+++ b/packages/agents/router/src/lib/operations/execute.ts
@@ -10,6 +10,8 @@ import {
 import { CallDataForNonContract, MissingXCall, NotEnoughAmount, ParamsInvalid } from "../errors";
 import { getHelpers } from "../helpers";
 import { getContext } from "../../router";
+import { BigNumber } from "ethers";
+import { getAuctionAmount } from "../helpers/auctions";
 
 // fee percentage paid to relayer. need to be updated later
 export const RELAYER_FEE_PERCENTAGE = "1"; //  1%
@@ -23,6 +25,7 @@ export const execute = async (params: OriginTransfer): Promise<void> => {
   const { requestContext, methodContext } = createLoggingContext(execute.name);
 
   const {
+    config,
     logger,
     adapters: { wallet, subgraph, txservice },
     routerAddress,
@@ -75,29 +78,40 @@ export const execute = async (params: OriginTransfer): Promise<void> => {
   // based on a calculation of which rounds we can afford to bid on. For now, this is hardcoded to bid
   // only on the first auction round.
   // Produce the router path signatures for each auction round we want to bid on.
-  const signatures = {
-    "1": await signRouterPathPayload(transferId, "1", wallet),
-  };
 
-  logger.debug("Signed payloads", requestContext, methodContext, {
-    rounds: Object.keys(signatures),
-    // Sanitized with ellipsis.
-    sigs: Object.values(signatures).map((s) => s.slice(0, 6) + ".."),
-  });
-
-  // sanity check
+  // Make a list of signatures that reflect which auction rounds we want to bid on.
   const balance = await subgraph.getAssetBalance(destinationDomain, routerAddress, executeLocalAsset);
-  if (balance.lt(receivingAmount)) {
+  const signatures: Record<string, string> = {};
+  for (let roundIdx = 1; roundIdx <= config.auctionRoundDepth; roundIdx++) {
+    const amountForRound = getAuctionAmount(roundIdx, BigNumber.from(receivingAmount));
+    if (amountForRound.lte(balance)) {
+      signatures[roundIdx.toString()] = await signRouterPathPayload(transferId, roundIdx.toString(), wallet);
+    } else {
+      logger.debug(`Not enough balance for this round: ${roundIdx}. Skipping!`, requestContext, methodContext, {
+        balance: balance.toString(),
+        amountForRound: amountForRound.toString(),
+      });
+    }
+  }
+
+  if ([...Object.keys(signatures)].length == 0) {
     throw new NotEnoughAmount({
       balance: balance.toString(),
       receivingAmount: receivingAmount.toString(),
       executeLocalAsset,
       routerAddress,
       destinationDomain: destinationDomain,
+      maxRoundDepth: config.auctionRoundDepth,
       requestContext,
       methodContext,
     });
   }
+
+  logger.debug("Signed payloads", requestContext, methodContext, {
+    rounds: Object.keys(signatures),
+    // Sanitized with ellipsis.
+    sigs: Object.values(signatures).map((s) => s.slice(0, 6) + ".."),
+  });
 
   if (callData !== "0x") {
     const code = await txservice.getCode(+destinationDomain, to);
--- a/packages/agents/router/test/lib/operations/execute.spec.ts
+++ b/packages/agents/router/test/lib/operations/execute.spec.ts
@@ -52,6 +52,9 @@ describe("Operations:Execute", () => {
         router: mockRouter,
         signatures: {
           "1": mock.signature,
+          "2": mock.signature,
+          "3": mock.signature,
+          "4": mock.signature,
         },
       };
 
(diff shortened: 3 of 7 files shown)


packages/agents/router/src/config.ts
```diff
@@ -82,6 +83,7 @@ export const NxtpRouterConfigSchema = Type.Object({
   mode: TModeConfig,
   network: Type.Union([Type.Literal("testnet"), Type.Literal("mainnet"), Type.Literal("local")]),
   polling: TPollingConfig,
+  auctionRoundDepth: Type.Number(),
```
jakekidd, line 86: couple things:
1. nit: may want to consider renaming this to `maxPathLength` or `maxRoutersPerTransfer` for consistency
2. we should be consulting the on-chain value, `maxRoutersPerTransfer`, to cap whatever this value can be configured to be... or perhaps we should just set this value internally *to* `maxRoutersPerTransfer` by default. At the very least, before closing out this PR please leave a TODO with this information

wanglonghong: If its already configured on onchain, we should be able to read the on-chain value. correct.

packages/agents/router/src/lib/operations/execute.ts
```diff
@@ -108,7 +115,8 @@ export const execute = async (params: OriginTransfer): Promise<void> => {
 
   logger.debug("Sanity checks passed", requestContext, methodContext, { liquidity: balance.toString() });
 
-  const fee = DEFAULT_ROUTER_FEE;
+  // TODO. Needs to calculate fee amount in BigNumber and pass it into Bid.
+  const fee = "0";
```
jakekidd, line 119: This 'fee' can actually be removed from bids for now. It was a placeholder, as I initially thought routers would be sending competitive custom fees/rates sooner.

But this will actually stay hardcoded on-chain for now. The `LIQUIDITY_FEE_NUMERATOR` and `LIQUIDITY_FEE_DENOMINATOR` are what's used to determine the LP fee the router actually gets (approximately).

packages/agents/router/test/lib/operations/execute.spec.ts
```diff
@@ -1,6 +1,6 @@
 import { constants, utils, BigNumber } from "ethers";
 import { reset, restore } from "sinon";
-import { Bid, DEFAULT_ROUTER_FEE, expect, OriginTransfer } from "@connext/nxtp-utils";
+import { Bid, expect, OriginTransfer } from "@connext/nxtp-utils";
```
jakekidd, line 3: in line with the previous comment, pls remove `DEFAULT_ROUTER_FEE` from codebase, neither are needed

packages/agents/router/src/lib/operations/execute.ts
jakekidd: Or at least change the TODO here pls

wanglonghong: I don't think so. I think the fee should be there for off-chain calculation even though we use hardcoded values rn.

packages/agents/sequencer/src/lib/operations/auctions.ts
```diff
@@ -146,7 +146,6 @@ export const executeAuctions = async (_requestContext: RequestContext) => {
   // Handling each domain in parallel, but each individual transfer synchronously. This is to account
   // for the fact that one transfer's auction might affect another. For instance, a router might have
   // 100 tokens to LP, but bid on 2 100-token transfers. We shouldn't send both of those bids.
-  // RS: going to try out parallelizing it anyways, its okay if a few of them fail
```
jakekidd, line 149: This is worth discussing imo!! This could be parallelized, within reason - if we suspect that the majority of routers' bids won't get picked anyway (due to randomness).

Tradeoff is : routers polluting the network with overbidding vs routers underbidding conservatively. Arguably the sequencer should manage which routers don't have sufficient funds and filter them out... but if we pile all the work on the sequencer to do this, it won't scale very well

packages/agents/sequencer/src/lib/operations/auctions.ts
```diff
@@ -193,139 +192,187 @@ export const executeAuctions = async (_requestContext: RequestContext) => {
...
-                    destination,
-                    amount: amount.toString(),
-                  },
-                  assetBalanceId: `${asset.toLowerCase()}-${router.toLowerCase()}`,
-                  routerLiquidity,
+          for (const roundIdx of availableRoundIds) {
+            logger.debug(`Selecting the round ${roundIdx}`, requestContext, methodContext, { availableRoundIds });
+            const roundIdInNum = Number(roundIdx);
+            const totalBids = bidsRoundMap[roundIdInNum];
+            const combinatedBidsForRound = generateCombinations(
```
jakekidd, line 210: nit: `combinated` should be `combined` lol

packages/agents/sequencer/src/lib/helpers/auctions.ts
```diff
@@ -63,3 +64,83 @@ export const getDestinationLocalAsset = async (
...
+  return Math.pow(2, round - 1);
+};
+
+/**
+ * Generate all combinations of an array.
+ * @param sources - Array of input elements.
+ * @param length - Desired length of combinations.
+ * @return - Array of combination arrays.
+ */
+export const generateCombinations = (sources: any[], length: number): any[] => {
```
jakekidd, line 117: total nitpick but - this isn't generating arrays, it's generating **sets**. If I understand it correctly, this function is really more like `getAllSubsets`. I'd prefer a name like that.

difference ofc being that, for arrays, order matters, for sets, it doesn't

this doesn't need to return a ts Set object, should still just return an array, but yeah verbage here is important lol

packages/agents/sequencer/src/lib/operations/auctions.ts
```diff
@@ -193,139 +192,187 @@ export const executeAuctions = async (_requestContext: RequestContext) => {
...
+
+              try {
+                logger.debug("Sending bid to relayer", requestContext, methodContext, {
                   transferId,
-                  asset,
+                  bid: {
+                    // NOTE: Obfuscating signatures here for safety.
+                    routers: randomCombination.map((bid) => bid.router),
+                    fee: randomCombination.map((bid) => bid.fee),
+                    totalFee: randomCombination.reduce(
```
jakekidd, line 307: Again, can probably remove this `fee` stuff here. But ultimately, it's supposed to be a percentage (like `"5"` basis points aka 0.05%), *not* a specific value. So definitely should remove `totalFee` as makes no sense to add those percentages together lol.

packages/agents/sequencer/src/lib/operations/auctions.ts
```diff
@@ -193,139 +192,187 @@ export const executeAuctions = async (_requestContext: RequestContext) => {
...
-                  transfer: {
-                    transferId,
-                    asset,
-                    destination,
-                    amount: amount.toString(),
-                  },
-                  assetBalanceId: `${asset.toLowerCase()}-${router.toLowerCase()}`,
-                  routerLiquidity,
+          for (const roundIdx of availableRoundIds) {
+            logger.debug(`Selecting the round ${roundIdx}`, requestContext, methodContext, { availableRoundIds });
```
jakekidd, line 207: can we log the number of available bids here? like how many we have to choose from, how many combinations maybe?

packages/agents/sequencer/src/lib/helpers/auctions.ts
wanglonghong: it actually returns an array of combinations, not sets

packages/agents/router/src/lib/operations/execute.ts
jakekidd: any off-chain calculations should pull the current values from on-chain state as needed.

This value is 0.05% everywhere on-chain rn, and it's enforced on-chain. Nothing we record here is relevant.

In the future, we may allow routers to bid their own percentages (e.g. if a router wants to be competitive, they might bid 0.04% if it were an option). But that feature won't exist for a while. It's safe to remove `fee` from the `Bid` type for now.

jakekidd (approved):
awesome work. pls remove `fee` tho, if you have time! It's not needed for anything and won't be used for a long time.

packages/agents/sequencer/src/lib/helpers/auctions.ts
jakekidd: I know the type is an array, but mathematically it's a subset

every length-2 subset of A, B, C:
A, B
B, C
A, C

every length-2 sub-array of A, B, C:
A, B
B, A
B, C
C, B
A, C
C, A

not too important, just felt like it was worth noting in a comment at least, idk

jakekidd (changes requested):
hang on, spotted a potential issue here

packages/agents/router/src/lib/operations/execute.ts
```diff
@@ -75,30 +71,41 @@ export const execute = async (params: OriginTransfer): Promise<void> => {
...
 
-  // sanity check
+  // Make a list of signatures that reflect which auction rounds we want to bid on.
   const balance = await subgraph.getAssetBalance(destinationDomain, routerAddress, executeLocalAsset);
-  if (balance.lt(receivingAmount)) {
+  const signatures: Record<string, string> = {};
+  for (let roundIdx = 1; roundIdx <= config.maxPathLength; roundIdx++) {
+    const amountForRound = getAuctionAmount(roundIdx, BigNumber.from(receivingAmount));
+    if (amountForRound.lte(balance)) {
+      signatures[roundIdx.toString()] = await signRouterPathPayload(transferId, roundIdx.toString(), wallet);
```
jakekidd, line 81: this won't work. Routers can't sign bids with the round number - they need to sign the path length!

So for round 3, if the transfer amount is split 4-ways, the router should be signing "4" for the path length, not "3"!
```ts
signRouterPathPayload(transferId, "4", wallet);
```

packages/agents/router/src/config.ts
jakekidd: UPDATE: I was wrong! This should be changed back to `auctionRoundDepth` or `maxAuctionRounds` or something. Clearly, auction rounds != maximum path length! They scale separately.

packages/agents/router/src/lib/operations/execute.ts
wanglonghong: > this won't work. Routers can't sign bids with the round number - they need to sign the path length!
> 
> So for round 3, if the transfer amount is split 4-ways, the router should be signing "4" for the path length, not "3"!
> 
> ```ts
> signRouterPathPayload(transferId, "4", wallet);
> ```

good catch!

packages/agents/router/src/config.ts
```diff
@@ -15,6 +15,7 @@ const DEFAULT_SUBGRAPH_POLL_INTERVAL = 15_000;
 const DEFAULT_CONFIRMATIONS = 3;
 const MIN_CACHE_POLL_INTERVAL = 2_000;
 const DEFAULT_CACHE_POLL_INTERVAL = 20_000;
+const DEFAULT_MAX_PATH_LENGTH = 4;
```
jakekidd: Change this to 3 for now

current on-chain maxRoutersPerTransfer is 5

Round 4 is 8-router transfer, and will revert

We can update this default to be 4 later

packages/agents/router/src/lib/operations/execute.ts
jakekidd: bump. it looks like we're still passing in roundIdx.toString() ? We should be translating roundIdx => pathLength

packages/agents/router/test/lib/operations/execute.spec.ts
```diff
@@ -70,10 +72,32 @@ describe("Operations:Execute", () => {
         mockXTransfer.origin.assets.bridged.asset,
         mockXTransfer.destinationDomain,
       );
-      expect(mock.helpers.shared.signRouterPathPayload).to.be.calledOnce;
+      expect(mock.helpers.shared.signRouterPathPayload).to.be.callCount(4);
```
jakekidd, line 75: does this work? thought it should be:
```ts
expect(mock.helpers.shared.signRouterPathPayload.callCount).to.be.eq(4);
```

packages/agents/sequencer/test/lib/helpers/auctions.spec.ts
```diff
@@ -79,4 +88,163 @@ describe("Helpers:Auctions", () => {
...
+          fee: "0",
+          signatures: {
+            "2": mkSig("0xrouter3_2"),
+          },
+        },
+        {
+          transferId: transferId,
+          origin: "1111",
+          router: router4,
+          fee: "0",
```
jakekidd, line 198: side note, you can remove the `fee` param here lol

packages/agents/router/src/lib/operations/execute.ts
```diff
@@ -77,30 +73,42 @@ export const execute = async (params: OriginTransfer): Promise<void> => {
...
 
-  // sanity check
+  // Make a list of signatures that reflect which auction rounds we want to bid on.
   const balance = await subgraph.getAssetBalance(destinationDomain, routerAddress, executeLocalAsset);
-  if (balance.lt(receivingAmount)) {
+  const signatures: Record<string, string> = {};
+  for (let roundIdx = 1; roundIdx <= config.auctionRoundDepth; roundIdx++) {
+    const amountForRound = getAuctionAmount(roundIdx, BigNumber.from(receivingAmount));
+    if (amountForRound.lte(balance)) {
+      const pathLen = Math.pow(2, roundIdx - 1);
```
LayneHaber, line 83: why does the path length increase non-linearly?

LayneHaber: discussed in discord!

LayneHaber (approved):
Concerns addressed in discord, this lgtm

packages/agents/sequencer/src/config.ts
```diff
@@ -69,6 +70,7 @@ export const getEnvConfig = (
       configFile.supportedBidVersion ||
       version,
     subgraphPrefix: process.env.SEQ_SUBGRAPH_PREFIX || configJson.subgraphPrefix || configFile.subgraphPrefix,
+    auctionRoundDepth: process.env.AUCTION_ROUND_DEPTH || configJson.auctionRoundDepth || DEFAULT_AUCTION_ROUND_DEPTH,
```
LayneHaber, line 73: as discussed in discord -- will need some v
Screen Shot 2022-06-01 at 8 53 05 PM
alidation around the maximum routers you include with this bid path. context:

jakekidd: @wanglonghong's reply to this:
image

wanglonghong: resolve in https://github.com/connext/nxtp/issues/1343

Issue #1280, opened 2022-05-27
Production testnet subgraphs and DB are out of sync.


preethamr:
**Describe the bug**
Transfer data in the DB and subgraph are out of sync for a prolonged period. `poller` poll loop not functioning properly.

**Expected behavior**
`poller` should not stall, and should produce more debug logs while in process.

**Additional context**
Production testnet subgraphs and DB are out of sync.

sanchaymittal:
Was fixed and merged.

PR #1281, opened 2022-05-27
1280 production testnet subgraphs and db are out of sync

preethamr:
## Description

Transfer data in the DB and subgraph are out of sync for a prolonged period. poller poll loop not functioning properly.

## Type of change

- [ ] Refactoring
- [ x] Bug fix (non-breaking change which fixes an issue)

`Connextscan` UI should not be out of sync with subgraphs.

## Related Issue(s)

Fixes: https://github.com/connext/nxtp/issues/1280

### DIFF
--- a/packages/agents/backend/poller/src/lib/operations/transfers.ts
+++ b/packages/agents/backend/poller/src/lib/operations/transfers.ts
@@ -90,11 +90,17 @@ export const updateTransfers = async () => {
 
     if (xcalledTransfers.concat(executedTransfers).concat(reconciledTransfers).length == 0) {
       done = true;
-      logger.debug("Processed all pending transfers. Last page", requestContext, methodContext, { page });
+      logger.debug("Processed all pending transfers. Last page", requestContext, methodContext, {
+        lastPage: page,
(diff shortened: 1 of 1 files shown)


packages/agents/backend/poller/src/lib/operations/transfers.ts
```diff
@@ -90,9 +90,11 @@ export const updateTransfers = async () => {
 
     if (xcalledTransfers.concat(executedTransfers).concat(reconciledTransfers).length == 0) {
       done = true;
+      logger.debug("Processed all pending transfers. Last page", requestContext, methodContext, { page });
```
wanglonghong, line 93: I think that would be better to have pageSize as well here. so like this {page, pageSize}

wanglonghong (approved):
I think

packages/agents/backend/poller/src/lib/operations/transfers.ts
preethamr: Makes sense. Done

LayneHaber (approved):

jakekidd (approved):

PR #1284, opened 2022-05-29
Parallelize loops

### LINKED ISSUE #1279: As a team, we need to remove/optimize critical bottlenecks in order to support mainnet usage at scale
## Problem
Transactions are getting locked up longer than expected due to suboptimal code/use of our cache

This nested for loop at the router cache may be root prolbem for transfers to get stuck over 10min https://github.com/connext/nxtp/blob/main/packages/agents/router/src/bindings/cache/index.ts

## Ideas for solutions
We should
1. Review where we have bottlenecks or inefficiencies that will impact us on mainnet, and identify which ones are critical. NOTE: Be highly cognizant of scope, this is not an openended refactor task
2. Refactor i.e. parellelize as needed

## Acceptance criteria
[ ] Fast transfers complete within expected 30s-60s window at scale 
[ ] Cache does not get locked up 
[ ]

rhlsthrm:
Fixes #1279


rhlsthrm:
Tested locally against staging.

Issue #1307, opened 2022-05-30
As a user, I want to parse logs from the xcall TransactionReceipt


sanchaymittal:
**Problem**
A clear and concise description of what the problem is

utils function to allow parsing logs from xcall() TransactionReceipt to get the transferId.

**Ideas to solve this**
A clear and concise description of what you think could solve this problem.

**Acceptance Criteria**
[ ] X happens
[ ] Y happens

**Other**

PR #1308, opened 2022-05-30
1307 parselogs

sanchaymittal:
## Type of change

- [ ] Refactoring
- [X] New feature (non-breaking change which adds functionality)

## Related Issue(s)

Fixes: #1307


packages/agents/sdk/src/sdkUtils.ts
```diff
@@ -50,6 +51,16 @@ export class NxtpSdkUtils {
     return new NxtpSdkUtils(nxtpConfig, logger, chainData);
   }
 
+  public parseConnextTransactionReceipt(transactionReceipt: providers.TransactionReceipt): any {
+    const parsedlogs: any = [];
+    transactionReceipt.logs.forEach((log) => {
+      const l = this.contracts.connext.parseLog(log);
```
rhlsthrm, line 57: This is gonna throw an error. We should make this function more specific and make the function:
```ts
function getTransferIdFromXCallReceipt(...)
```
Then we know exactly which log in the receipt to parse for.

packages/agents/sdk/test/sdkUtils.spec.ts
```diff
@@ -65,6 +66,14 @@ describe("SdkUtils", () => {
     });
   });
 
+  describe("#parseConnextTransactionReceipt", () => {
+    it("happy: should work", async () => {
+      const res = nxtpUtils.parseConnextTransactionReceipt(mock.ethers.receipt());
```
rhlsthrm, line 71: Fix this test by using a real xcall receipt.

rhlsthrm (changes requested):

packages/agents/sdk/src/sdkUtils.ts
sanchaymittal: we can parse any log related to connext interface though, it doesn't have to be xcall right?

packages/agents/sdk/test/sdkUtils.spec.ts
sanchaymittal: yeah good idea, will add it.

PR #1314, opened 2022-05-30
1226 strict versioning

### LINKED ISSUE #1226: As a router, I need to know contract versioning and have the ability to upgrade/reference versions before submitting bids
## Background

If we upgrade contracts, routers should be able to check onchain and not bid on transfers if the contract version changes.

### Linked Issues & Documentation

N/A

## Definition Of Done

Routers and sequencers should be able to reference contract version number onchain before supporting transfers

## Tasks

- [x] Add version to contracts
- [ ] Add checks on router
- [ ] Add checks on sequencer

LayneHaber:
## Description

Routers should be prevented from bidding if the contracts are updated and they have not made prerequisite updates in their offchain logic. To fix this, we add versioning to all of the nomad router contracts. This can be leveraged in offchain checks to prevent bidding with non-compatible contract versions

## Type of change

- [ ] Refactoring

- Add the `VERSION()` getter to the `IConnextHandler` contract.

## Related Issue(s)

Fixes #1226


packages/deployments/contracts/contracts/core/promise/PromiseRouter.sol
```diff
@@ -13,6 +13,7 @@ import {IBridgeToken} from "../connext/interfaces/IBridgeToken.sol";
 
 import {Router} from "../shared/Router.sol";
 import {XAppConnectionClient} from "../shared/XAppConnectionClient.sol";
+import {Version0} from "../shared/Version0.sol";
```
liu-zhipeng, line 17: If the version of the `ConnextHandler` is upgraded to 1 or 2 ..., `PromiseRouter` and `RelayerFeeRouter` should import `Version1.sol` or `Version2.sol` ?

Why don't we import `Version.sol` instead of `Version0.sol`?

LayneHaber: The `Version0.sol` was what existed (ripped from nomad), but I agree with you. 

Will make the fix

liu-zhipeng (approved):

PR #1325, opened 2022-05-31
Rename to cartographer

rhlsthrm:
## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes


ops/testnet/prod/backend/main.tf
```diff
@@ -22,10 +22,10 @@ data "aws_route53_zone" "primary" {
   zone_id = "Z03634792TWUEHHQ5L0YX"
 }
 
-module "poller_db" {
-  domain                = "poller"
+module "cartographer_db" {
```
carlomazzaferro, line 25: note: this will destroy and re-create the db. maybe a good time to bump the db instance from `db.t2.small` to something larger

rhlsthrm: Yes let's make it larger and also update to the latest version. Can you set this up in this PR?

Issue #1220, opened 2022-05-23
Refactor backend poller polling loop


rhlsthrm:
Use PM2 restart strategy: https://pm2.keymetrics.io/docs/usage/restart-strategies/

Ensures system will not get locked up by a stuck loop.

PR #1340, opened 2022-06-02
Autorestart event loop using pm2

preethamr:
## Description
Use pm2 to externally monitor and manage `poller` event loop.
Update default config

## Type of change

- [x] Refactoring

Users are more likely to see the connextscan explorer is in sync with the subgraphs, and the chains.

## Related Issue(s)
https://github.com/connext/nxtp/issues/1220
Fixes


packages/agents/cartographer/poller/src/bindings/poller/index.ts
```diff
@@ -8,12 +8,31 @@ import { updateRouters } from "../../lib/operations/routers";
...
+
+const stopExternally = () => {
+  stoppedExternally = true;
+};
+
+process.on("SIGINT", function () {
+  stopExternally();
+});
+
+export const bindPoller = async (_pollInterval?: number, _pollIterations?: number) => {
```
rhlsthrm, line 21: Should we just remove this completely and run a single iteration? I don't see the need to loop in JS anymore.

preethamr: We could, I thought so too. Looping within JS removes startup costs, and hence would perform slightly better, as it would autorestart after the last loop. And this would be best of both worlds.

preethamr: I have updated it to do a single iteration, and not loop in JS anymore.

packages/agents/cartographer/poller/src/cartographer.ts
```diff
@@ -49,16 +49,9 @@ export const makeCartographer = async (_configOverride?: CartographerConfig) =>
   context.domains = Object.keys(supported).filter((domain) => supported[domain]);
 
   /// MARK - Bindings
-  await bindPoller();
-
   context.logger.info("Cartographer initialized!", requestContext, methodContext, {
     domains: context.domains,
   });
-  context.logger.info(`
```
rhlsthrm, line 57: :(

rhlsthrm (commented):
https://pm2.keymetrics.io/docs/usage/signals-clean-restart/#cleaning-states-and-jobs

Should we do this? To try to gracefully clean?

packages/agents/cartographer/poller/package.json
```diff
@@ -22,7 +25,7 @@
     "dbmate": "dbmate"
   },
   "nodemonConfig": {
-    "exec": "node --enable-source-maps ./dist/index.js | pino-pretty",
+    "exec": "pm2 start ./dist/index.js --source-map-support | pino-pretty",
```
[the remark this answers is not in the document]
preethamr: Yes. Done

packages/agents/cartographer/poller/src/cartographer.ts
preethamr: now its on again :)

PR #1344, opened 2022-06-02
Handle router's liquidity when the transfer gets executed/reconciled

wanglonghong:
## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes

### DIFF
--- a/packages/deployments/subgraph/src/amarok-runtime-staging/mapping.ts
+++ b/packages/deployments/subgraph/src/amarok-runtime-staging/mapping.ts
@@ -240,6 +240,7 @@ export function handleXCalled(event: XCalled): void {
  */
 export function handleExecuted(event: Executed): void {
   const num = event.params.args.routers.length;
+  const amount = event.params.args.amount;
   const routers: string[] = [];
   for (let i = 0; i < num; i++) {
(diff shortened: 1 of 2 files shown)


packages/deployments/subgraph/src/amarok-runtime-staging/mapping.ts
```diff
@@ -251,7 +252,13 @@ export function handleExecuted(event: Executed): void {
       router.isActive = true;
       router.save();
     }
+
     routers.push(router.id);
+
+    // Update router's liquidity
+    const assetBalance = getOrCreateAssetBalance(event.params.args.local, event.params.args.routers[i]);
+    assetBalance.amount = assetBalance.amount.minus(amount.div(BigInt.fromI32(num)));
```
sanchaymittal, line 260: When we are doing multipath, are all the routers contributing the same amount? Noticed we don't really ask for the amount in router signature.

sanchaymittal (approved):

packages/deployments/subgraph/src/amarok-runtime-staging/mapping.ts
wanglonghong: Yes. contributing the same amount

PR #1362, opened 2022-06-05
[DNM] Split bindings

rhlsthrm:


packages/agents/cartographer/poller/package.json
```diff
@@ -2,16 +2,14 @@
...
   "files": [
     "dist/**/*",
     "src/**/*"
   ],
   "scripts": {
-    "dev": "concurrently -i -k -n 'app,tsc' -c 'green,cyan' 'nodemon'",
-    "start": "pm2 start dist/index.js --source-map-support --kill-timeout 5000 --cron-restart=\"* * * * *\" --time",
-    "stop": "pm2 delete index",
+    "dev": "concurrently -i -k -n 'app,tsc' -c 'green,cyan' 'pm2-runtime --watch cartographer.config.js'",
+    "start": "pm2-runtime cartographer.config.js",
```
preethamr, line 12: We need to update the Dockerfile `CMD` accordingly. I can do that update.

rhlsthrm: Yes, we will need to modify Dockerfile. Eventually, we can split into separate Docker images for each pm2 app.

preethamr: Updated the Dockerfile.

Right, that would align with any auto-scaling we would use, and can scale them seperately

preethamr:
LGTM. And tests passing.

PR #1363, opened 2022-06-05
Use abort controller for timeout in fallback subgraph

jakekidd:
## Description

GraphQLClient supports using `AbortController`:
https://github.com/prisma-labs/graphql-request#cancellation
https://github.com/node-fetch/node-fetch#request-cancellation-with-abortsignal

This seems like a more functional way of aborting requests due to timeout than the current method (Promise.race).


jakekidd:
IS the current default timeout (10s) okay?

sanchaymittal:
too soon

PR #1383, opened 2022-06-08
Executor unit tests

wanglonghong:
## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes

### DIFF
--- a/packages/deployments/contracts/contracts_forge/core/connext/helpers/Executor.t.sol
+++ b/packages/deployments/contracts/contracts_forge/core/connext/helpers/Executor.t.sol
@@ -4,9 +4,7 @@ pragma solidity 0.8.14;
 import {IERC20} from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 
 import "../../../utils/ForgeHelper.sol";
-
 import "../../../../contracts/test/TestERC20.sol";
-
 import "../../../../contracts/core/connext/interfaces/IExecutor.sol";
 import "../../../../contracts/core/connext/helpers/Executor.sol";
 
@@ -106,8 +104,6 @@ contract ExecutorTest is ForgeHelper {
     // fund executor
     asset.mint(address(executor), 10 ether);
   }
-
-  // ============ Utils ============
 
   // ============ getConnext ============
 


packages/deployments/contracts/contracts/core/connext/helpers/Executor.sol
```diff
@@ -117,6 +117,8 @@ contract Executor is IExecutor {
 
     bool isNative = _args.assetId == address(0);
 
+    if (isNative) require(msg.value == _args.amount, "!equal");
+
```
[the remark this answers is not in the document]
LayneHaber: also we should handle this failure gracefully otherwise funds will get parked in the executor

LayneHaber (approved):

packages/deployments/contracts/contracts/core/connext/helpers/Executor.sol
```diff
@@ -26,12 +28,35 @@ contract Executor is IExecutor {
...
   // ============ Properties =============
 
+  /**
+   * @notice Address of associated connext
+   */
   address private immutable connext;
+
+  /**
+   * @notice Properties set for access via reentrancy
+   * @dev Contracts that are being called during `execute` may set
```
jakekidd, line 38: nit: Can we move this really great description of "properties" in general to be under the `// ============ Properties =============`

packages/deployments/contracts/contracts/core/connext/helpers/Executor.sol
```diff
@@ -26,12 +28,35 @@ contract Executor is IExecutor {
...
+   * elapsed).
+   *
+   * Callers can access these by using code:
+   * `IExecutor(msg.sender).originSender();`
+   */
   bytes private properties = LibCrossDomainProperty.EMPTY_BYTES;
+
+  /**
+   * @notice Amount transferred
+   * @dev This is always accessible via reentrancy. May *NOT* be the exact amount
```
jakekidd, line 50: This description feels misleading. This value *should* be accessed via reentrancy, and will be set only for the duration of the execution

jakekidd: Let's move this description to all be under that general "Properties" comment under `// ============ Properties =============`

packages/deployments/contracts/contracts/core/connext/helpers/Executor.sol
```diff
@@ -191,24 +238,39 @@ contract Executor is IExecutor {
...
+  function _sendToRecovery(
+    bool _isNative,
+    bool _hasIncreased,
     address _assetId,
     address payable _to,
     address payable _recovery,
     uint256 _amount
   ) private {
-    if (!isNative) {
+    if (_amount == 0) {
```
jakekidd, line 260: what happens if we got here with `msg.value != _args.amount` ??

If the amount is 0 but the msg.value has some extra (somehow, idk how), then that msg.value will get stuck here.

packages/deployments/contracts/contracts_forge/core/connext/helpers/Executor.t.sol
```diff
@@ -203,19 +239,190 @@ contract ExecutorTest is ForgeHelper {
         data
       )
     );
+  }
+
+  // Should gracefully handle failures if value != amount
```
jakekidd, line 244: On top of the previous comment about the edge case where value != amount, and amount == 0, would like to see that case tested here

jakekidd (changes requested):
need to handle the edge case I mentioned here

jakekidd (approved):
my bad, edge case is invalid - we're passing in msg.value as amount. looks good!

packages/deployments/contracts/contracts/core/connext/helpers/Executor.sol
```diff
@@ -1,8 +1,8 @@
 // SPDX-License-Identifier: UNLICENSED
 pragma solidity 0.8.14;
 
-import {AddressUpgradeable} from "@openzeppelin/contracts-upgradeable/utils/AddressUpgradeable.sol";
-import {SafeERC20Upgradeable, IERC20Upgradeable} from "@openzeppelin/contracts-upgradeable/token/ERC20/utils/SafeERC20Upgradeable.sol";
+import {Address} from "@openzeppelin/contracts/utils/Address.sol";
+import {SafeERC20, IERC20} from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
 // import {ExcessivelySafeCall} from "@nomad-xyz/excessively-safe-call/src/ExcessivelySafeCall.sol";
```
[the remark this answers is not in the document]
LayneHaber: remove what? the nomad import thats commented out?

i don't want to remove it, i would eventually like to use it straight from the package

packages/deployments/contracts/contracts/core/connext/helpers/Executor.sol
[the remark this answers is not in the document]
wanglonghong: Nice!

liu-zhipeng (approved):

PR #1424, opened 2022-06-16
Subgraph `handleExecute` mapping should use amount-after-fees for debiting router

jakekidd:
## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes


rhlsthrm (approved):

wanglonghong (approved):
lgtm

jakekidd:
Marked as DNM - using transactingAmount doesn't actually work since it's not the amount debited from the router, but the amount after **both** fee-taking *and* swap

jakekidd:
> Marked as DNM - using transactingAmount doesn't actually work since it's not the amount debited from the router, but the amount after **both** fee-taking _and_ (potentially) the swap to adopted asset

Hardcoded the fee calc in subg for now

Issue #1430, opened 2022-06-17
As a user, my transfer data should not have gaps


preethamr:
**Describe the bug**
Queries form `cartographer` to subgraphs use timestamps to checkpoint. This checkpoint mechanism because of quirks of how queries to subgraph work, are causing data gaps.

**To Reproduce**
1. Start with a fresh DB
2. Sync with prod testnet config
3. See data gaps

sanchaymittal:
Was completed in sprint 5

Issue #1324, opened 2022-05-31
Cartographer not showing some router balances


rhlsthrm:
https://api.thegraph.com/subgraphs/name/connext/nxtp-amarok-runtime-v0-rinkeby/graphql?query=query+MyQuery+%7B%0A++router%28id%3A+%220xc4eac3bfae3f3c951225c16848ddcab3d6a5abbf%22%29+%7B%0A++++assetBalances+%7B%0A++++++asset+%7B%0A++++++++adoptedAsset%0A++++++++blockNumber%0A++++++++canonicalDomain%0A++++++++canonicalId%0A++++++++id%0A++++++++local%0A++++++%7D%0A++++%7D%0A++%7D%0A%7D

This shows up on the subgraph but if I query cartographer:

```
GET {{postgrestUrl}}/routers_with_balances?address=eq.0xc4eac3bfae3f3c951225c16848ddcab3d6a5abbf
```

I only receive for domains Goerli and Kovan:

```json
[
  {
    "address": "0xc4eac3bfae3f3c951225c16848ddcab3d6a5abbf",
    "asset_canonical_id": "0x0000000000000000000000003ffc03f05d1869f493c7dbf913e636c6280e0ff9",
    "asset_domain": "2221",
    "router_address": "0xc4eac3bfae3f3c951225c16848ddcab3d6a5abbf",
    "balance": 10000000000000000000000000000,
    "local": "0x3ffc03f05d1869f493c7dbf913e636c6280e0ff9",
    "adopted": "0x3ffc03f05d1869f493c7dbf913e636c6280e0ff9",
    "canonical_id": "0x0000000000000000000000003ffc03f05d1869f493c7dbf913e636c6280e0ff9",
    "canonical_domain": "2221",
    "domain": "2221"
  },
  {
    "address": "0xc4eac3bfae3f3c951225c16848ddcab3d6a5abbf",
    "asset_canonical_id": "0x0000000000000000000000003ffc03f05d1869f493c7dbf913e636c6280e0ff9",
    "asset_domain": "3331",
    "router_address": "0xc4eac3bfae3f3c951225c16848ddcab3d6a5abbf",
    "balance": 10000000000000000000000000000,
    "local": "0x3ffc03f05d1869f493c7dbf913e636c6280e0ff9",
    "adopted": "0x3ffc03f05d1869f493c7dbf913e636c6280e0ff9",
[...]

PR #1435, opened 2022-06-19
1430 as a user my transfer data should not have gaps

preethamr:
Queries form cartographer to subgraphs use timestamps to checkpoint. This checkpoint mechanism because of quirks of how queries to subgraph work, are causing data gaps.

Adding a new mechanism to monitor for data gaps, and fetch from subgraph, to close the gaps.

## Type of change

- [x] Refactoring

## Related Issue(s)

Fixes 
https://github.com/connext/nxtp/issues/1430
https://github.com/connext/nxtp/issues/1324


packages/adapters/subgraph/src/lib/operations/queries.ts
```diff
@@ -368,19 +406,52 @@ export const getDestinationTransfersByExecuteTimestampQuery = (
...
+  const domains = Object.keys(config.sources);
+  for (const domain of domains) {
+    const prefix = config.sources[domain].prefix;
+    if (params.has(domain)) {
+      combinedQuery += destinationTransfersByIDsQueryString(
+        prefix,
+        params.get(domain)!.transferIDs.map((id) => `"${id}"`),
+      );
+    } else {
+      console.log(`No agents for domain: ${domain}`);
```
rhlsthrm, line 449: Prefer to not use console logs here.

packages/adapters/subgraph/src/lib/operations/queries.ts
```diff
@@ -294,72 +306,98 @@ export const getOriginTransfersQuery = (agents: Map<string, SubgraphQueryMetaPar
...
-  return `
-  ${prefix}_destinationTransfers(
-    where: { 
-      executedTimestamp_gte: ${fromTimestamp}, 
-      ${maxBlockNumber ? `, executedBlockNumber_lte: ${maxBlockNumber}` : ""} 
-    }, 
-    orderBy: executedTimestamp, 
+  return `${prefix}_destinationTransfers(
+    where: {
+      nonce_gte: ${fromNonce},
```
rhlsthrm, line 358: The main problem we run into here is that the reconciled transfers don't have a nonce. They only have a timestamp.

packages/adapters/subgraph/src/lib/operations/queries.ts
```diff
@@ -154,9 +159,16 @@ export const getAssetBalancesQuery = (prefix: string, router: string): string =>
...
-export const getAssetBalancesAllRoutersQuery = (prefix: string): string => {
+export const getAssetBalancesRoutersQuery = (
+  prefix: string,
+  offset: number,
+  orderDirection: "asc" | "desc" = "desc",
+): string => {
   const queryString = `
-    ${prefix}_routers {
+    ${prefix}_routers(
+    skip: ${offset},
```
rhlsthrm, line 169: We should also use the `first` param here so we can be explicit about how big the pages are.

packages/agents/cartographer/poller/db/schema.sql
```diff
@@ -182,6 +215,13 @@ ALTER TABLE ONLY public.transfers
     ADD CONSTRAINT transfers_pkey PRIMARY KEY (transfer_id);
 
 
+--
+-- Name: transfers update_time_on_transfers; Type: TRIGGER; Schema: public; Owner: -
+--
+
+CREATE TRIGGER update_time_on_transfers BEFORE UPDATE ON public.transfers FOR EACH ROW EXECUTE FUNCTION public.trigger_set_timestamp();
```
rhlsthrm, line 222: Love it.

packages/agents/cartographer/poller/src/lib/operations/transfers.ts
```diff
@@ -31,58 +49,122 @@ export const updateTransfers = async () => {
...
-    const transfers = await subgraph.getOriginTransfersByXCallTimestamp(subgraphXCallQueryMetaParams);
-    logger.info("Retrieved origin transfers by xcalled timestamp", requestContext, methodContext, {
+    const transfers = await subgraph.getOriginTransfersByNonce(subgraphOriginQueryMetaParams);
+    logger.info("Retrieved origin transfers by nonce", requestContext, methodContext, {
       transfers,
       count: transfers.length,
     });
     await database.saveTransfers(transfers);
+
+    for (const domain of domains) {
```
rhlsthrm, line 102: Nit, but this should be able to be parallelized i.e.

```ts
await Promise.all(domains.map(async (domain) => {
  ...
});
```

packages/agents/cartographer/poller/src/lib/operations/transfers.ts
```diff
@@ -31,58 +49,122 @@ export const updateTransfers = async () => {
...
+  if (subgraphDestinationQueryMetaParams.size > 0) {
+    // Get destination transfers for all domains in the mapping.
+    const transfers = await subgraph.getDestinationTransfersByNonce(subgraphDestinationQueryMetaParams);
+    logger.info("Retrieved destination transfers by nonce", requestContext, methodContext, {
       transfers,
       count: transfers.length,
     });
     await database.saveTransfers(transfers);
+
+    for (const domain of domains) {
```
rhlsthrm, line 121: Also can be parallelized.

packages/adapters/subgraph/src/lib/operations/queries.ts
preethamr: True, we will not get reconciled but not executed transfers here.
They would be fetched by the reconciled timestamp queries, separately

We could use timestamps everywhere, but using nonce is more robust, as it handles the case where there are more transfers that got executed in 1 second, than can fit in response, and cause us repeatedly process then same transfers and get stuck.

So, thought use nonce when we can, if not use timestamp.

rhlsthrm: This makes sense and is likely the best approach for now.

packages/agents/cartographer/poller/src/lib/operations/transfers.ts
preethamr: Yes!

packages/agents/cartographer/poller/src/lib/operations/transfers.ts
preethamr: Done

packages/adapters/subgraph/src/lib/operations/queries.ts
preethamr: Removed them.

packages/adapters/subgraph/src/lib/operations/queries.ts
preethamr: Changed it use first as well, with explicit page size.

PR #1438, opened 2022-06-21
Remove hardhat tests

LayneHaber:
## Description

- Removes hardhat tests in favor of foundry (also removes any unused helper contracts)
- Implements integration tests within foundry
- Fixes a few bugs (see below)

## Type of change

- [ ] Refactoring

Bugs fixed include:
- the `transferId` is different from `xcall` and `execute` when there is a swap involved
- you can create promise messages with 0-length return data, but you cannot process them
- portal fee debt getter was wrong
- portal was calculating the amount of debt to take out incorrectly when there is a swap

## Related Issue(s)

N/A


liu-zhipeng (approved):

packages/deployments/contracts/contracts/core/promise/libraries/PromiseMessage.sol
```diff
@@ -138,8 +138,8 @@ library PromiseMessage {
     }
     uint256 _length = lengthOfReturnData(_view);
     // before = 1 byte identifier + 32 bytes transferId + 20 bytes callback address + 1 byte success + 32 bytes length + x bytes data
-    // nonzero return data
-    return _length > 0 && (RETURNDATA_START + _length) == _len;
+    // allow zero-length return data
+    return _length >= 0 && (RETURNDATA_START + _length) == _len;
```
jakekidd, line 142: I'm wondering why this was here before... copy/paste error, or just wrong impression when it was written at the time? If it's the former, we should make sure it's not an issue elsewhere right?

jakekidd (approved):
absolute fooking legend mate

packages/deployments/contracts/contracts/core/promise/libraries/PromiseMessage.sol
jakekidd: nvm, just the one spot. interesting that we glossed over this!

PR #1494, opened 2022-07-12
Router msg queue

rhlsthrm:
## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes


packages/agents/router/src/config.ts
```diff
@@ -218,14 +198,6 @@ export const getEnvConfig = (
         })(),
     };
 
-    const maxLag = chainConfig.subgraph?.maxLag ?? MIN_SUBGRAPH_SYNC_BUFFER;
```
jakekidd, line 221: why are we removing max lag?

jakekidd: oh are we just removing the subgraph config, i c

rhlsthrm: Yep, it's not being used anywhere.

PR #1500, opened 2022-07-13
1483 estimate fee

### LINKED ISSUE #1483: As a user, I need relayer fees estimated for me and applied on destination
**Problem**
`relayerFee` is always set to 0. This works on testnet but will not work on mainnet. We need to be able to estimate the cost to send an execute transaction on the destination side (including calldata) and convert into origin side gas units to send with the transaction.

**Ideas to solve this**
"Easy" implementation is using an offchain oracle. We can lean on Gelato to help us add assets as needed so that we can convert the fees. What we need to do:
- Have a hardcoded gas amount for a basic `execute` (can use chainData and vary per chain if needed).
- Add any user specified calldata gas to the gas amount. This is required because external calls will mostly not be able to be directly estimated.
- Add a 15% buffer for relayer profit.
- Convert the gas amount to USD using the oracle API (Gelato). This can be displayed to the user.
- Calculate the amount of origin side gas needed to meet the USD-based requirement.
- Send with origin side `xcall`.

**Acceptance Criteria**
- [ ] SDK has a method called `estimateRelayerFee`.
- [ ] Method is able to take in calldata gas as a parameter to specify extra gas for calldata.
- [ ] Method will return the amount of origin gas required, as well as a USD estimation.
- [ ] Able to pass in relayer fee into the `xcall` SDK method.

**Other**

sanchaymittal:
## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes #1483


packages/utils/src/peripherals/chainData.ts
```diff
@@ -63,6 +63,7 @@ export type ChainData = {
     standard: string;
   }[];
   gasEstimates: {
+    execute: string;
     prepare: string;
     fulfill: string;
     cancel: string;
```
wanglonghong, line 69: we can remove prepare/fulfill/cancel because we don't use them anymore in amarok

wanglonghong (changes requested):

PR #1501, opened 2022-07-13
Debug deployments

liu-zhipeng:
## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes


jakekidd:
We don't need these deployment records, as we won't be using the deployments here in staging as per convo with @LayneHaber

PR #1527, opened 2022-07-18
Sanity checks

rhlsthrm:
## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes

### DIFF
--- a/packages/agents/router/src/errors/execute.ts
+++ b/packages/agents/router/src/errors/execute.ts
@@ -1,48 +1,60 @@
-import { NxtpError } from "@connext/nxtp-utils";
+import { NxtpError, Values } from "@connext/nxtp-utils";
 
-export class ParamsInvalid extends NxtpError {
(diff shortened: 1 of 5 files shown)


packages/agents/router/src/errors/execute.ts
```diff
@@ -52,50 +64,56 @@ export class NotEnoughLiquidity extends NxtpError {
...
 }
 
-export class InvalidAuctionRound extends NxtpError {
+export class InvalidAuctionRound extends ExecuteError {
   constructor(context: any = {}) {
     super("Invalid auction round", context, InvalidAuctionRound.name);
   }
 }
+
+export class UnableToGetAsset extends ExecuteError {
```
preethamr, line 115: Is this a `retryable = true` error in all cases ?

rhlsthrm: It's retryable = false.

PR #1529, opened 2022-07-19
reduce contract size of stableswap

liu-zhipeng:
## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes


LayneHaber (approved):
pending passing tests

LayneHaber:
also -- what's the final size here?

LayneHaber:
Screen Shot 2022-07-21 at 12 24 44 PM

final size seems to still be too large for mainnet on the `StableSwap`.

were going to be using the `StableSwapFacet` when we bootstrap our own AMMs and its likely any external AMM would write their own logic. should we just remove this from any deployments and explicitly mark it as oversized for mainnet so only an example implementation?

LayneHaber (changes requested):
still too large, see comment on PR

jakekidd:
> Screen Shot 2022-07-21 at 12 24 44 PM
> 
> final size seems to still be too large for mainnet on the `StableSwap`.
> 
> were going to be using the `StableSwapFacet` when we bootstrap our own AMMs and its likely any external AMM would write their own logic. should we just remove this from any deployments and explicitly mark it as oversized for mainnet so only an example implementation?

I actually thought `StableSwap` was just a deprecated contract from before our diamond upgrade

liu-zhipeng:
image

LayneHaber (approved):

LayneHaber:
> I actually thought `StableSwap` was just a deprecated contract from before our diamond upgrade

It's not deprecated, but it won't be used directly in prod deployments. It is useful to have for testing, and may as well be included since it's been in the audit scopes

PR #1534, opened 2022-07-19
Spearbit-45

liu-zhipeng:
## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes

### DIFF
--- a/packages/deployments/contracts/contracts/core/connext/helpers/Executor.sol
+++ b/packages/deployments/contracts/contracts/core/connext/helpers/Executor.sol
@@ -203,14 +203,14 @@ contract Executor is IExecutor {
     // Set the amount as well
     amnt = _args.amount;
 
-    // Ensure there is enough gas to handle failures
-    uint256 gas = gasleft() - FAILURE_GAS;
+    // Asset balance of this contract before call
+    uint256 beforeBalance = _assetBalance(_args.assetId);
 
     // Try to execute the callData
     // the low level call will return `false` if its execution reverts
     (success, returnData) = ExcessivelySafeCall.excessivelySafeCall(
       _args.to,
-      gas,
+      gasleft() - FAILURE_GAS,
       isNative ? _args.amount : 0,
       MAX_COPY,
       _args.callData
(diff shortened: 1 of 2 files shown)


packages/deployments/contracts/contracts/core/connext/helpers/Executor.sol
```diff
@@ -227,6 +227,19 @@ contract Executor is IExecutor {
...
+    if (hasValue) {
+      uint256 unclaimed = _args.amount - (beforeBalance - _assetBalance(_args.assetId));
+      if (unclaimed > 0) {
+        _sendToRecovery(isNative, hasValue, _args.assetId, payable(_args.to), payable(_args.recovery), unclaimed);
+      }
+    }
+
+    // Reset allowance to 0!
+    if (!isNative && hasValue) {
+      SafeERC20.safeApprove(IERC20(_args.assetId), _args.to, 0);
```
jakekidd, line 240: We should only have to reset allowance to 0 if there were "unclaimed" / unspent tokens, right?

If all the tokens were spent, then wouldn't the allowance now be 0?

packages/deployments/contracts/contracts/core/connext/helpers/Executor.sol
```diff
@@ -227,6 +227,19 @@ contract Executor is IExecutor {
       _sendToRecovery(isNative, hasValue, _args.assetId, payable(_args.to), payable(_args.recovery), _args.amount);
     }
 
+    // If there is unclaimed tokens, send them to the recovery address
+    if (hasValue) {
+      uint256 unclaimed = _args.amount - (beforeBalance - _assetBalance(_args.assetId));
+      if (unclaimed > 0) {
+        _sendToRecovery(isNative, hasValue, _args.assetId, payable(_args.to), payable(_args.recovery), unclaimed);
+      }
```
liu-zhipeng, line 235: This will decrease allowance as well. so i think we don't need `safeApprove` anymore

LayneHaber (changes requested):
Personally I don't think we should handle this case. I think it adds two different types of complexity:

(1) It adds more code complexity to the `Executor` contract in general
(2) It breaks the assumption that the only edgecase we handle is around your calldata failing. Calldata that leaves amounts within the `Executor` should be considered improperly implemented `calldata`, and outside of the handling scope of the protocol

I would change my mind if we decided to implement an interface for the `Executor` contract (which if implemented correctly, would remove the need for this handling). For now though, I think we should close this PR and treat this issue as "acknowledged"

liu-zhipeng:
> Personally I don't think we should handle this case. I think it adds two different types of complexity:
> 
> (1) It adds more code complexity to the `Executor` contract in general (2) It breaks the assumption that the only edgecase we handle is around your calldata failing. Calldata that leaves amounts within the `Executor` should be considered improperly implemented `calldata`, and outside of the handling scope of the protocol
> 
> I would change my mind if we decided to implement an interface for the `Executor` contract (which if implemented correctly, would remove the need for this handling). For now though, I think we should close this PR and treat this issue as "acknowledged"

Agree!

PR #1536, opened 2022-07-19
Spearbit-63: add recommended checks to `xcall`

jakekidd:
https://github.com/spearbit-audits/connext/issues/63

### DIFF
--- a/packages/deployments/contracts/contracts/core/connext/facets/BridgeFacet.sol
+++ b/packages/deployments/contracts/contracts/core/connext/facets/BridgeFacet.sol
@@ -41,12 +41,14 @@ contract BridgeFacet is BaseConnextFacet {
   error BridgeFacet__setPromiseRouter_invalidPromiseRouter();
   error BridgeFacet__setExecutor_invalidExecutor();
   error BridgeFacet__setSponsorVault_invalidSponsorVault();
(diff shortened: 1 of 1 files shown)


jakekidd:
Still missing the final check mentioned in the issue here:
> If _args.params.slippageTol is set to something greater than s.LIQUIDITY_FEE_DENOMINATOR then funds can be locked as xcall() allows for the user to provide the local asset, avoiding any swap while _handleExecuteLiquidity() in execute() may attempt to perform a swap on the destination chain.

jakekidd:
Added the final check as mentioned.

This check to make sure the slippageTol will add a fair bit of gas to the call as we would be doing an SLOAD to get the current denominator.

Something that could resolve this: have the denominator be a pre-defined constant value. It seems like that denominator value could be tracked separately from the `s.LIQUIDITY_FEE_DENOMINATOR`, afaik. It seems like the slippage tolerance could just have a hardcoded 10k as the denominator (that seems reasonable).

LayneHaber:
>  It seems like the slippage tolerance could just have a hardcoded 10k as the denominator (that seems reasonable).

There is no way to update the `LIQUIDITY_FEE_DENOMINATOR` so it essentially is hardcoded

LayneHaber (approved):

PR #1546, opened 2022-07-20
[DNM] Catch up unhandled errors

wanglonghong:
## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes

### DIFF
--- a/packages/agents/router/src/subscriber/bindings/exceptions/index.ts
+++ b/packages/agents/router/src/subscriber/bindings/exceptions/index.ts
@@ -1,0 +1,15 @@
+import { createLoggingContext } from "@connext/nxtp-utils";
+
+import { getContext } from "../../subscriber";
+
+export const bindUnhandledErrors = async () => {
+  const { logger } = getContext();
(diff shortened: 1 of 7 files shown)


packages/agents/router/src/subscriber/bindings/exceptions/index.ts
```diff
@@ -0,0 +1,15 @@
...
+import { getContext } from "../../subscriber";
+
+export const bindUnhandledErrors = async () => {
+  const { logger } = getContext();
+  const { requestContext, methodContext } = createLoggingContext(bindUnhandledErrors.name);
+  process.on("unhandledRejection", (error) => {
+    logger.warn("Unhandled rejection occured!", requestContext, methodContext, { error });
+  });
+
+  process.on("uncaughtException", (error) => {
```
rhlsthrm, line 12: This is very dangerous to do. We don't want to catch all unhandled errors, we specifically handle errors with try catch if we want them caught.

rhlsthrm (changes requested):
Do not do this, let's figure out why the errors are happening first.

PR #1550, opened 2022-07-20
Spearbit-65

liu-zhipeng:
## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes

### DIFF
--- a/packages/deployments/contracts/contracts/core/connext/helpers/Executor.sol
+++ b/packages/deployments/contracts/contracts/core/connext/helpers/Executor.sol
@@ -203,14 +203,14 @@ contract Executor is IExecutor {
     // Set the amount as well
     amnt = _args.amount;
 
-    // Ensure there is enough gas to handle failures
-    uint256 gas = gasleft() - FAILURE_GAS;
+    // Asset balance of this contract before call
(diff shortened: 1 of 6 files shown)


LayneHaber (approved):

LayneHaber:
I did some cleanup around the `AssetLogic` interfaces:

- All the `handle*` functions are now the ones that deal with wrapping and unwrapping native assets
- They return an `address asset` to show the asset that was transferred out of the contracts
- I removed any returning of `amount` -- because we don't support fee-on-transfer tokens (they revert when trying to transfer to the contract) we can always assume `amount in == amount specified`
- I cleaned up any code handling potential differences in amount

packages/deployments/contracts/contracts/core/connext/helpers/Executor.sol
```diff
@@ -222,7 +223,7 @@ contract Executor is IExecutor {
     // Unset amount
     amnt = 0;
 
-    // Handle failure cases
+    // Handle success cases
     if (!success) {
```
liu-zhipeng, line 227: `fix failure cases`?

LayneHaber: dammit

PR #1595, opened 2022-07-25
Spearbit 25

liu-zhipeng:
## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes 
https://github.com/spearbit-audits/connext/issues/25


packages/deployments/contracts/contracts/core/connext/helpers/SponsorVault.sol
```diff
@@ -316,4 +343,35 @@ contract SponsorVault is ISponsorVault, ReentrancyGuard, Ownable {
...
+  ) internal view returns (bool) {
+    uint256 ethPrice = priceOracle.getPriceFromChainlink(address(0));
+    if (ethPrice == 0) {
+      return false;
+    }
+
+    uint256 tokenPrice = priceOracle.getPriceFromChainlink(_token);
+    if (tokenPrice == 0) {
+      return false;
+    }
```
LayneHaber, line 369: Doesn't the `IPriceOracle` default to using chainlink -> oracle -> dex? maybe we could have a way to enforce the price feed source is acceptable instead of forcing it to come only from chainlink

LayneHaber: open to other arguments though! cc @wanglonghong

liu-zhipeng: > Doesn't the `IPriceOracle` default to using chainlink -> oracle -> dex? maybe we could have a way to enforce the price feed source is acceptable instead of forcing it to come only from chainlink

Our `ConnextPriceOracle` has a member function to get price feed from chainlink. `getPriceFromChainlink`

PR #1626, opened 2022-07-26
Spearbit-72: Require a signature from a whitelisted Sequencer agent in execute fast liquidity route

jakekidd:
See https://github.com/spearbit-audits/connext/issues/72

Adding a sequencer signature requirement, as well as a whitelist for sequencers to resolve this issue


LayneHaber (approved):
We should have a similar remove whitelist process for the sequencer whitelist.  I think it should be added in a future upgrade since there's no way removing the whitelist is valid at the moment, but just a note.

Approved!

PR #1636, opened 2022-07-27
Fix for nomad bridge context

rhlsthrm:


packages/agents/router/src/subscriber/subscriber.ts
```diff
@@ -11,6 +11,14 @@ import { Web3Signer } from "@connext/nxtp-adapters-web3signer";
 import { getContractInterfaces, TransactionService, contractDeployments } from "@connext/nxtp-txservice";
 import axios from "axios";
 import { BridgeContext } from "@nomad-xyz/sdk-bridge";
+import fetch, { Headers, Request, Response } from "node-fetch";
+
+if (!(globalThis as any).fetch) {
```
preethamr, line 16: Ah, that is interesting.

preethamr (approved):

PR #1663, opened 2022-07-29
Add key to swapUtils events

wanglonghong:
## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes


packages/deployments/contracts/contracts/core/connext/libraries/SwapUtils.sol
```diff
@@ -22,36 +22,47 @@ library SwapUtils {
 
   /*** EVENTS ***/
 
-  event TokenSwap(address indexed buyer, uint256 tokensSold, uint256 tokensBought, uint128 soldId, uint128 boughtId);
+  event TokenSwap(
+    bytes32 key,
```
jakekidd, line 26: let's make all of these `key` properties for each event `indexed`!

jakekidd (commented):
cc @LayneHaber . lgtm, but need to be sure it's okay to alter this library to be different than the standardized implementation

packages/deployments/contracts/contracts/core/connext/libraries/SwapUtils.sol
wanglonghong: https://github.com/connext/nxtp/pull/1663/commits/bb84f9d5f28908e1ee5e7a55dadd6f413c828f23

jakekidd (approved):

PR #1703, opened 2022-08-09
Process lighthouse txs in the sequencer

wanglonghong:
## Type of change

- [x] Refactoring

## Related Issue(s)

Fixes


packages/agents/sequencer/src/lib/operations/lighthouse.ts
```diff
@@ -0,0 +1,198 @@
...
+  await cache.lighthousetxs.storeLightHouseData(lighthouseData);
+  logger.info("Created a lighthouse tx", requestContext, methodContext, { transferId, lighthouseData });
+
+  const message: Message = {
+    transferId: transfer.transferId,
+    originDomain: transfer.xparams!.originDomain,
+    type: MessageType.SlowPath,
+  };
+
+  await mqClient.publish(config.messageQueue.publisher!, {
```
rhlsthrm, line 134: Hmm, is this using the same message queue keys as the fast path? I want to make sure it's fully separated so the type/routingKey should be properly namespaced, so we should make it so it always publishes with a separate name.

packages/agents/sequencer/src/lib/operations/relayer.ts
```diff
@@ -96,3 +105,81 @@ export const sendToRelayer = async (
   const taskId = await relayer.send(destinationChainId, destinationConnextAddress, encodedData, _requestContext);
   return taskId;
 };
+
+export const sendLightHouseDataToRelayer = async (
```
rhlsthrm, line 109: Let's break this file up into 2 files, `relayerFast` and `relayerSlow`.

packages/agents/sequencer/src/sequencer.ts
```diff
@@ -122,7 +124,12 @@ export const execute = async (_configOverride?: SequencerConfig) => {
     context.adapters.contracts = getContractInterfaces();
     context.adapters.relayer = await setupRelayer();
 
-    await executeAuction(transferId, requestContext);
+    if (messageType == MessageType.Auction) {
+      await executeAuction(transferId, requestContext);
+    } else if (messageType == MessageType.SlowPath) {
```
rhlsthrm, line 129: OK I see here these are namespaced, nice!

rhlsthrm (changes requested):
This is great, nice work on this! Main things are around naming in a less confusing way. Here's what would help:
- Don't have anything named `auctions` etc, it should be `executeFast` and `executeSlow`.
- Split up all the logic into separate files i.e. `relayer.ts` should definitely be split.

packages/agents/lighthouse/src/lib/operations/execute.ts
```diff
@@ -38,5 +35,5 @@ export const execute = async (args: ExecuteArgs, transferId: string): Promise<vo
   }
 
   const encodedData = contracts.connext.encodeFunctionData("execute", [args]);
-  await sendToRelayer(args, encodedData, transferId, requestContext);
+  await sendBidsToRelayer(args, encodedData, transferId, requestContext);
```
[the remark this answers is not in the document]
wanglonghong: oh you're right

packages/agents/sequencer/src/lib/operations/lighthouse.ts
wanglonghong: Yes. What I was thinking here is basically to make the message broker any kind of message, not just execute-fast ones. so we can integrate more like AMB off-chain stuff.

packages/agents/sequencer/src/lib/operations/auctions.ts
```diff
@@ -42,10 +42,10 @@ export const storeBid = async (bid: Bid, _requestContext: RequestContext): Promi
   }
 
   // check if bid router version is compatible with hosted sequencer
-  const checkVersion = compare(bid.routerVersion, config.supportedBidVersion!, "<");
+  const checkVersion = compare(bid.routerVersion, config.supportedVersion!, "<");
```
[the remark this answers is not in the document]
wanglonghong: Ok will go through the link you attached and update as needed

packages/agents/sequencer/src/sequencer.ts
wanglonghong: Nit: That would be better to update MessageType.Auction like MessageType.FastPath?

packages/agents/sequencer/src/lib/operations/lighthouse.ts
```diff
@@ -0,0 +1,198 @@
...
+      });
+    }
+    // Store the transfer locally. We will use this as a reference later when we execute this transfer
+    // in the cycle, for both encoding data and passing relayer fee to the relayer.
+    await cache.transfers.storeTransfers([transfer]);
+  }
+
+  if (!transfer.destination?.reconcile || transfer.destination?.execute) {
+    // This transfer has already been Executed or not Reconciled yet, so slow liquidity is no longer valid.
+    throw new InvalidSlowLiqTransfer({
```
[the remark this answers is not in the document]
wanglonghong: It just checks whether its a slow transfer or not. we don't need to process fast transfer here

packages/agents/sequencer/src/sequencer.ts
rhlsthrm: Yes, correct this would be better!

PR #1710, opened 2022-08-11
Improve root message handling and interface

jakekidd:
- Added immutable flag `IS_HUB` set on instantiation.
- Renamed `sendMessage` / `receiveMessage` to `sendRoot` / `receiveRoot` to make interface less confusing (we already have `process` methods)
- Renamed related events as well.
- `sendRoot` takes no arguments and just sends the current `outboundRoot`
- Virtual method `_receiveRoot`, which should be implemented by the specific <Domain>Connector, now returns the new updated root (`bytes32`). `receiveRoot` calls internal `_receiveRoot`, sets the `aggregateRoot` to the returned value, and if `IS_HUB` is true, will also update the RootManager (by calling `setOutboundRoot(mirrorDomain, aggregateRoot)`)
- Added method `propagateRoot` which can only be called by `RootManager` on the hub chain (i.e. if `IS_HUB` is true)


packages/deployments/contracts/contracts/core/messaging/interfaces/IConnector.sol
```diff
@@ -40,11 +40,23 @@ import {IProposedOwnable} from "../../shared/interfaces/IProposedOwnable.sol";
...
+   * chain out to the spoke/L2 chains.
+   */
+  function propagateRoot(bytes memory _data) external;
 
-  function processMessage(address _sender, bytes memory _data) external;
+  /**
+   * @notice This is called by relayers to trigger passing of current root to mainnet root manager
+   * @dev This is called at specific time intervals
+   */
+  function sendRoot() external;
```
LayneHaber, line 53: I don't love having two different functions that are designed to both send a `bytes32` to their mirror. it seems like we could make it a single function and use the `IS_HUB` tag

packages/deployments/contracts/contracts/core/messaging/connectors/MainnetConnector.sol
```diff
@@ -32,7 +43,7 @@ contract MainnetL1Connector is Connector {
    * @dev Messaging uses this function to send outboundRoot to root manager. On mainnet,
    * would just update the root manager directly
    */
-  function _sendMessage(bytes memory _data) internal override {
+  function _sendRoot(bytes memory _data) internal override {
```
LayneHaber, line 46: needs `msg.sender` checks as well

packages/deployments/contracts/contracts/core/messaging/connectors/Connector.sol
```diff
@@ -222,21 +236,51 @@ abstract contract Connector is ProposedOwnable, MerkleTreeManager, ConnectorMana
...
-  function sendMessage(bytes memory _data) external {
-    _sendMessage(_data);
+  /**
+   * @notice This should be called by RootManager to propagate an updated aggregate root on the Hub/L1
+   * chain out to the spoke/L2 chains.
+   */
+  function propagateRoot(bytes memory _data) external onlyRootManager onlyHub {
+    // RootManager has aggregated roots from all other Connectors on this hub chain.
+    // TODO: Should we update our current aggregateRoot to reflect the new one.
+    // aggregateRoot = bytes32(_data);
```
LayneHaber, line 246: yeah -- i don't see why not

packages/deployments/contracts/contracts/core/messaging/connectors/Connector.sol
```diff
@@ -40,9 +41,9 @@ abstract contract Connector is ProposedOwnable, MerkleTreeManager, ConnectorMana
 
   event Process(bytes32 leaf, bool success, bytes returnData);
 
-  event MessageSent(bytes data, address caller);
+  event RootUpdateSent(bytes data, address caller);
 
-  event MessageProcessed(address from, bytes data, address caller);
+  event RootUpdateReceived(address from, bytes data, address caller);
```
LayneHaber, line 46: do we want to be more specific with outbound / aggregate roots in the events if we have functions that are more opinionated?

packages/deployments/contracts/contracts/core/messaging/connectors/Connector.sol
```diff
@@ -222,21 +236,51 @@ abstract contract Connector is ProposedOwnable, MerkleTreeManager, ConnectorMana
...
+  /**
+   * @notice This should be called by RootManager to propagate an updated aggregate root on the Hub/L1
+   * chain out to the spoke/L2 chains.
+   */
+  function propagateRoot(bytes memory _data) external onlyRootManager onlyHub {
+    // RootManager has aggregated roots from all other Connectors on this hub chain.
+    // TODO: Should we update our current aggregateRoot to reflect the new one.
+    // aggregateRoot = bytes32(_data);
+
+    // TODO: This code block is partially redundant with the `sendRoot` method. Should we just make `sendRoot` public?
```
LayneHaber, line 248: `sendRoot` is public? im confuse

packages/deployments/contracts/contracts/core/messaging/interfaces/IConnector.sol
LayneHaber: is the issue the difference in args? if so, then lets be more explicit in the name about the roots we are sending. `sendOutboundRoot` / `sendAggregateRoot` or something

jakekidd: down for that! the second thing, that is. `sendOutboundRoot` and `sendAggregateRoot`

packages/deployments/contracts/contracts/core/messaging/connectors/Connector.sol
jakekidd: it's `external` and `external != public` lol.

I was thinking - should we just make sendRoot public, then we can technically call it externally *and* internally but meh

packages/deployments/contracts/contracts/core/messaging/connectors/Connector.sol
jakekidd: as in, with the naming here?

Like, `OutboundRootSent` or `AggregateRootReceived` ?

jakekidd:
Closing this out as we accomplished with similar issues - may come back later to reuse some things from this

Issue #1697, opened 2022-08-08
As a user, I want to see transaction metrics in Connextscan


rhlsthrm:
Recreate volume graphs from Connextscan.

<img width="1435" alt="image" src="https://user-images.githubusercontent.com/11512787/183446079-750e74c1-0aaf-411e-bdb3-7bece67cd2fa.png">

**Volume Metrics**

Get # transactions By Week. Get Total # Transactions by chain.

**Subtasks**

- [ ] Cartographer: #1695 
- [ ] SDK: #1686 
- [ ] UI integration

**Acceptance Criteria**

- [ ] Graph on Connextscan shows breakdown of Amarok # transactions by week.
- [ ] Visual shows Amarok total # transactions broken down by chain.

**Other**

Issue #1693, opened 2022-08-08
As a user, I want to see volume metrics in Connextscan


rhlsthrm:
Recreate volume graphs from Connextscan.

<img width="1435" alt="Screen Shot 2022-08-08 at 20 38 33" src="https://user-images.githubusercontent.com/11512787/183409713-553088ef-3cb5-4c7e-b341-219dbd2c3d58.png">

**Volume Metrics**

Get Volume By Week. Get Total Volume by chain.

**Subtasks**

- [ ] Cartographer: #1695 
- [ ] SDK: #1686 
- [ ] UI integration

**Acceptance Criteria**

- [ ] Graph on Connextscan shows breakdown of Amarok volume by week.
- [ ] Visual shows Amarok total volume broken down by chain.

**Other**

Issue #1684, opened 2022-08-03
As a user, I want to see Router TVL per chain


rhlsthrm:
**Problem**
We need to recreate the TVL graph from Connextscan for Amarok:

<img width="1071" alt="Screen Shot 2022-08-03 at 17 55 08" src="https://user-images.githubusercontent.com/11512787/182665654-12a9f923-d261-422b-84e2-1752cd862c12.png">

**Subtasks**

- [ ] UI integration
- [ ] SDK method to `getRouterTVL` #1685 
- [ ] Cartographer Postgrest endpoint to get total TVL #1686 

**Acceptance Criteria**

- [ ] Graph on Connextscan shows breakdown of Amarok router TVL

**Other**
We have 2 types of TVL, routers and StableSwap AMM. This is only for routers.

PR #1717, opened 2022-08-12
Feature/metrics vo

Jaynaik19:
## Addition of Metrics based on the below issues,

- https://github.com/connext/nxtp/issues/1684
- https://github.com/connext/nxtp/issues/1693
- https://github.com/connext/nxtp/issues/1697

Few important things:
- Given that the pricing data($) is missing in the database, Kept the aggregation daily instead of weekly. The $ value computation for volume and TVL will be more accurate when the pricing data is blended near time to transfer.
- Each row is a unique transfer
- For visibility into calculation added `max XCalled date` as `latest_transfer_day` in `router_tvl`


rhlsthrm:
Hey @Jaynaik19 can you please make a branch in the main repo instead of the fork? This is why the checks are failing. You should have access now.

rhlsthrm (approved):

PR #1744, opened 2022-08-19
Make a script for initializing and configuring protocol stack

jakekidd:
## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes


packages/deployments/contracts/src/config/configureProtocol.ts
```diff
@@ -0,0 +1,122 @@
...
+
+/**
+ * Handle configuration of the entire protocol, including messaging stack and connext diamond
+ * contracts, across all listed domains.
+ *
+ * Should effectively be a diagnostic on the whole protocol, making sure that everything that
+ * requires configuration and/or setup has been done so properly.
+ */
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+export const configureProtocol = async (protocol: ProtocolStack) => {
```
rhlsthrm, line 86: I don't think this should be in `src/config` just the name is a bit confusing since config means like configuration, but this is more like an initialization.

I like the idea of this, and I think this is something that should be able to be run by both hardhat tasks and integration tests so that we can run it on prod and local envs.

jakekidd: sure, I think init is a better name as well

We can work towards using both the `protocol` argument and a JSON config file on what to use here. Obviously I could just be grabbing deployments directly... But I'll leave that to the consumer (hardhat task or integration test, etc)

packages/deployments/contracts/src/init/initProtocol.ts
```diff
@@ -0,0 +1,122 @@
...
+  /// MARK - Deployer
+  // Get deployer mnemonic, which should be provided in env.
+  /// MARK - Deployments
+  // Get all deployments
+  /// ******************** MESSAGING ********************
+  /// MARK - Init
+  // TODO: Currently unused, as messaging init checks are not needed with the AMB-compatible stack.
+  // However, they will be useful as sanity checks for Nomad deployments in the future - thus, leaving
+  // this placeholder here for now...
+  /// MARK - Connector Mirrors
```
LayneHaber, line 102: there is a hardhat task for this, not sure if you can easily access those here

packages/deployments/contracts/src/init/initProtocol.ts
```diff
@@ -0,0 +1,122 @@
...
+  // Connectors should have their mirrors' address set; this lets them know about their counterparts.
+  /// MARK - Enroll Handlers
+  // Whitelist messaging routers as callers of dispatch?
```
LayneHaber, line 105: yep! `addSender` on connector -- this may be done in the deployment already but you will need to make sure all router (message sending) contracts are whitelisted on the connector

packages/deployments/contracts/src/init/initProtocol.ts
```diff
@@ -0,0 +1,122 @@
...
+  // Enroll messaging handlers.
+  /// ********************* CONNEXT *********************
+  /// MARK - Init
+  // Check to make sure Diamond Proxy is initialized.
+  /// MARK - Connextions
```
LayneHaber, line 110: we will likely be removing this once we combine the bridge router

packages/deployments/contracts/src/init/initProtocol.ts
```diff
@@ -0,0 +1,122 @@
...
+  //
+  /// ********************* ASSETS **********************
+  /// MARK - Register Assets
+  // Convert asset addresses
+  // Register assets in the TokenRegistry (enroll-custom).
```
LayneHaber, line 115: initialize swap can be done here as well

packages/deployments/contracts/src/init/initProtocol.ts
```diff
@@ -0,0 +1,122 @@
...
+
+type AssetStack = {
+  canonical: {
+    // The canonical domain of the asset.
+    domain: string;
+    // Address of the bridged asset on the canonical domain.
+    local: string;
+    // The address of the official canonical token.
+    // NOTE: If not specified here, then I guess the asset is natively cross-chain?
+    adopted?: string;
```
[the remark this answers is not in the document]
jakekidd: yeah, just wanted to make sure... I think it can just be `canonical: string` or whatever

packages/deployments/contracts/src/init/initProtocol.ts
```diff
@@ -0,0 +1,122 @@
...
+  sequencers: string[];
+  routers: string[];
+};
+
+type AssetStack = {
+  canonical: {
+    // The canonical domain of the asset.
+    domain: string;
+    // Address of the bridged asset on the canonical domain.
+    local: string;
```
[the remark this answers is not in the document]
jakekidd: yeah may have been a copy paste error or I just wasn't thinking lol

packages/deployments/contracts/src/init/initProtocol.ts
```diff
@@ -0,0 +1,122 @@
...
+  domain: string;
+
+  // RPC provider to use for this network.
+  rpc: string;
+
+  deployments: {
+    // Diamond.
+    Connext: string;
+    // Handlers.
+    BridgeRouter: string;
```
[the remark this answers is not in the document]
jakekidd: kk

packages/deployments/contracts/src/init/initProtocol.ts
jakekidd: trying to avoid hh tasks here as hh tasks use a specified network/chain, and this task should do init/config for all chains

PR #1754, opened 2022-08-23
add Dispatch event to subgraph; map to a Message entity that records the cross-chain message

### LINKED ISSUE #1753: Store `Dispatch` xchain messages in the subgraph
**Problem**
Subtask of https://github.com/connext/nxtp/issues/1714

**Ideas to solve this**
XCall is submitted, Dispatch event is emitted, subgraph mapping `handleDispatch` appends the message bytes to the correct `OriginTransfer` entry (use `messageHash` i.e. `leaf` for lookup!).

**Acceptance Criteria**

- [ ] `xcall` is called. `Dispatch` event is emitted from the transaction.
- [ ] `Message` is stored in subgraph.

jakekidd:
## Description

Fixes #1753 

## Type of change

- [ ] Refactoring

- Dispatch event now emits `destinationDomain`.
- Subgraph now records `Message` for cross-chain use by listening to targeted Connector contracts.
- We can consult the `Message` record in our polling loop for calling `proveAndProcess` (i.e. track which Messages need to be proven/delivered to which domains). 

- NOTE: Still need to add deployments/startBlock to the config for active deployments! Also, still need to redeploy the subgraph for staging!


packages/deployments/contracts/contracts/core/messaging/connectors/Connector.sol
```diff
@@ -36,7 +36,7 @@ abstract contract Connector is ProposedOwnable, MerkleTreeManager, ConnectorMana
 
   event SenderRemoved(address sender);
 
-  event Dispatch(bytes32 leaf, uint256 index, bytes32 root, bytes message);
+  event Dispatch(bytes32 leaf, uint256 index, bytes32 root, uint32 destinationDomain, bytes message);
```
LayneHaber, line 39: nit: we should make the domain indexed so we can find all messages to a specific domain

LayneHaber (approved):

jakekidd:
Closing this out as I believe this got replaced by other subg event logic

PR #1803, opened 2022-09-06
1786 intuitive defaults

### LINKED ISSUE #1786: As a dev, I want to use an intuitive default for transactingAsset for calldata-only xcalls
**Problem**

Add support for setting `transactingAsset: address(0)` in params for a calldata-only xcall. Right now we revert if it's set to `address(0)`. 

**Acceptance Criteria**

- [ ] Allow `address(0)` for `transactingAsset` param
- [ ] Enforce that `transactingAmount: 0` if `transactingAsset: address(0)`

**Other**

LayneHaber:
## Description

Fixes #1786 by adding a convention for 0-value transfers.

NOTE: when executing these transfers, you should use `address(0)` for the local asset on the router-side

## Type of change

- [ ] Refactoring

- Uses empty canonical info on transfer with `address(0) && amount == 0`
- Uses empty local when ^^ is true
- Tests

## Related Issue(s)

Fixes #1786


packages/deployments/contracts/contracts/core/connext/facets/BridgeFacet.sol
```diff
@@ -363,7 +366,13 @@ contract BridgeFacet is BaseConnextFacet {
     {
       // Check that the asset is supported -- can be either adopted or local.
       TokenId memory canonical = s.adoptedToCanonical[_args.transactingAsset];
-      if (canonical.id == bytes32(0)) {
+
```
jakekidd, line 369: we should avoid the `canonical` lookup here entirely if `_args.transactingAsset == address(0)`

jakekidd: so

```solidity
// empty token id
TokenId memory canonical;
if (_args.transactingAsset != address(0)) {
  // NOW we get the token id fr
  canonical = s.adoptedToCanonical[_args.transactingAsset];

  // do all these other checks that only matter if transactingAsset is specified anyway
  if (canonical.id == bytes32(0)) {

      if (s.tokenRegistry.isLocalOrigin(_args.transactingAsset)) {
        revert BridgeFacet__xcall_notSupportedAsset();
      }

      (uint32 canonicalDomain, bytes32 canonicalId) = s.tokenRegistry.getTokenId(_args.transactingAsset);
      canonical = TokenId(canonicalDomain, canonicalId);
    }
}

File: packages/deployments/contracts/contracts/core/connext/facets/BridgeFacet.sol [hunk 2]
```diff
@@ -363,7 +366,13 @@ contract BridgeFacet is BaseConnextFacet {
...
+      // NOTE: above we check that you can only have `address(0)` as a transacting asset when
+      // you are sending 0-amounts. Because 0-amount transfers shortcircuit all checks on
+      // mappings keyed on hash(canonicalId, canonicalDomain), this is safe even when the
+      // address(0) asset is not whitelisted. These values are only used for the `transactionId`
+      // generation
+      if (canonical.id == bytes32(0) && _args.transactingAsset != address(0)) {
```
jakekidd, line 375: what happens if the _args.transactingAsset == address(0) ?? should we still be calling `transferAssetToContract` and `swapToLocalAssetIfNeeded` below?

jakekidd: oh, I see, we do nothing in that method if the amount is 0... but still, we should just include the whole "transfer asset to contract and do asset stuff" step within a larger conditional block that checks to make sure this isn't a transfer with `_args.transactingAsset == address(0)`... at least, that seems like the more readable and efficient way to do it imo

packages/deployments/contracts/contracts/core/connext/facets/BridgeFacet.sol
LayneHaber: yeah that works too, ill make the change

packages/deployments/contracts/contracts/core/connext/facets/BridgeFacet.sol
LayneHaber: reasonable

packages/deployments/contracts/contracts/core/connext/facets/BridgeFacet.sol
```diff
@@ -362,31 +365,47 @@ contract BridgeFacet is BaseConnextFacet {
...
+      TokenId memory canonical;
+
+      // NOTE: above we check that you can only have `address(0)` as a transacting asset when
+      // you are sending 0-amounts. Because 0-amount transfers shortcircuit all checks on
+      // mappings keyed on hash(canonicalId, canonicalDomain), this is safe even when the
+      // address(0) asset is not whitelisted. These values are only used for the `transactionId`
+      // generation
+      if (_args.transactingAsset == address(0)) {
+        s.adoptedToCanonical[_args.transactingAsset];
+
```
liu-zhipeng, line 377: need this?

PR #1850, opened 2022-09-12
Fix connector deploy

liu-zhipeng:
## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes


packages/deployments/contracts/deploy/01_deployMessaging.ts
```diff
@@ -111,11 +111,9 @@ const handleDeployHub = async (
       continue;
     }
 
-    const contract = getConnectorName(protocol, mirrorChainId);
-    if (
-      (!connectorName.includes("Optimism") && !connectorName.includes("Polygon")) ||
-      connectorName.includes("Mainnet")
-    ) {
+    const contract = getConnectorName(protocol, mirrorChainId, protocol.hub);
+    console.log(contract);
```
jakekidd, line 115: remove console.log? or add details to it if we really wanna log here

packages/deployments/contracts/deploy/01_deployMessaging.ts
```diff
@@ -111,11 +111,9 @@ const handleDeployHub = async (
...
+    if ((!contract.includes("Optimism") && !contract.includes("Polygon")) || contract.includes("Mainnet")) {
```
jakekidd, line 115: a really helpful update for this would be to put these names in some kind of constant... we'll want to eventually make it so these names are less hardcoded in the future...

packages/deployments/contracts/tasks/addConnectors.ts
```diff
@@ -26,7 +27,7 @@ export default task("add-connectors", "Add all connectors to the root manager")
     const deployer = Wallet.fromMnemonic((networkConfig.accounts as any).mnemonic as unknown as string);
 
     const env = mustGetEnv(_env);
-    const remove = _remove ?? true;
+    const remove = _remove ? _remove === "true" : true;
```
jakekidd, line 30: instead of having this line, just change
```
.addOptionalParam("remove", "Whether or not to remove connectors that exist")
```
to
```
.addFlag("remove", "Whether or not to remove connectors that exist")
```

that way we can just use it as a flag, like:
```
ywc run hardhat addConnectors --remove
```

packages/deployments/contracts/tasks/setMirrorConnectors.ts
```diff
@@ -39,15 +39,15 @@ export default task("set-mirror-connectors", "Add a remote router")
           console.log(`tx mined:`, receipt.transactionHash);
         }
 
-        if (name.includes("PolygonL1Connector")) {
+        if (name.includes("PolygonHubConnector")) {
```
jakekidd, line 42: wait why are we hardcoded *just* for polygon here?

wb optimism, etc

jakekidd: oh bc polygon is special case with this FxChildTunnel thing

LayneHaber (approved):

PR #1908, opened 2022-09-19
Transactions support

rhlsthrm:
## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes


packages/agents/cartographer/poller/src/adapters/database/client.ts
```diff
@@ -262,69 +252,75 @@ export const getTransfersWithDestinationPending = async (
...
-    `.run(poolToUse);
-
-      await db.sql<s.asset_balances.SQL, s.asset_balances.JSONSelectable>`
-      INSERT INTO ${"asset_balances"} (${db.cols(balance.balance)}) VALUES (${db.vals(
-        balance.balance,
-      )}) ON CONFLICT ("asset_canonical_id", "asset_domain", "router_address") DO UPDATE SET (${db.cols(
-        balance.balance,
-      )}) = (${db.vals(balance.balance)}) RETURNING *
-    `.run(poolToUse);
+  await db.repeatableRead(poolToUse, async (txnClient) => {
```
aandreakis, line 285: It looks like this function is called from `routers.ts`: https://github.com/connext/nxtp/blob/96086f0270349aab37d9636d7b5f54e8ce3acf5b/packages/agents/cartographer/poller/src/lib/operations/routers.ts#L23

If a transactional scope is ever needed within `routers.ts`, it would be better to define the transaction there. This would allow to have a transactional scope that spans multiple db commands within `routers.ts`.

It would also keep this file cleaner so that the responsibility of defining transactions is moved to whoever is calling functions from `client.ts`.

aandreakis (changes requested):

PR #1913, opened 2022-09-20
Make ownable configurable by deployment

### LINKED ISSUE #1736: As a dev user, I need to be able to set custom acceptance delays for proposing/renouncing ownership
**Problem**

The acceptance delay parameter in `LibDiamond.sol` was lifted out into the constructor for `DiamondInit.sol` so that a customizeable delay could be used when deploying. We need to do a similar change for the `ProposedOwnable*` contracts.

**Ideas to solve this**

Make delay parameter a constructor argument instead of the hardcoded 7 days.

**Acceptance Criteria**

- [ ] ProposedOwnable.sol delay parameter is customizeable
- [ ] ProposedOwnableFacet.sol delay parameter is customizeable

**Other**

rhlsthrm:
## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes #1736


rhlsthrm:
> it seems we should have a single delay constant thats relevant for both upgrading the contracts and making any major ownership changes. is there a reason why they are different? open to other arguments here
> 
> other than that it lgtm

Hmm, I had thought it could be useful to have them separate tbh, you might want to upgrade quickly but make ownership changes slower. I could see our launch having these properties initially.

liu-zhipeng:
> 

Agree

liu-zhipeng (approved):

LayneHaber:
merging then!!

LayneHaber (approved):

Issue #1804, opened 2022-09-06
As a user, I want an easy to interact with BridgeToken


LayneHaber:
**Problem**
`permit` is a commonly used extension to the ERC20 interface that allows for improved UX. All users of local tokens minted by the router would benefit from having this included in the interface.

**Ideas to solve this**
Ensure the `BridgeToken` minted by the `TokenRegistry` includes `permit` in its interface.

**Acceptance Criteria**

- [ ] `OZERC20` implementation is upgraded

**Other**

LayneHaber:
merged, all done

PR #1937, opened 2022-09-20
permit implementation

preethamr:
## Type of change

- [x] Refactoring

## Related Issue(s)

Fixes 
https://github.com/connext/nxtp/issues/1804


packages/deployments/contracts/contracts/test/OZERC20.sol
```diff
@@ -287,4 +318,67 @@ contract ERC20 is IERC20 {
...
+  }
+
+  /**
+   * @dev See {IERC20Permit-DOMAIN_SEPARATOR}.
+   * This is ALWAYS calculated at runtime
+   * because the token name may change
+   */
+  function DOMAIN_SEPARATOR() external view override returns (bytes32) {
+    return _domainSeparatorV4();
+  }
```
LayneHaber, line 374: so, in the [eip712](https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/ac5931e2afbff4a66faede4c830998ee56a301c9/contracts/utils/cryptography/EIP712Upgradeable.sol#L63-L65) implementation called here it uses a stored variable of `_HASHED_NAME`. since we can update the name via the `_detailsHash` mechanism we have to call [`_buildDomainSeparator `](https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/ac5931e2afbff4a66faede4c830998ee56a301c9/contracts/utils/cryptography/EIP712Upgradeable.sol#L67-L73) directly using the `keccak256(bytes(token.name))` as the name hash, and the rest of the arguments remaining the same

LayneHaber (changes requested):
just need to change the default invocation used and its all good!

LayneHaber (approved):

preethamr:
Resolved in this branch:
https://github.com/connext/nxtp/pull/1942

Issue #1805, opened 2022-09-06
As a dev user, I want consistent decimals on tokens across chains


LayneHaber:
**Problem**
Right now, the token registry uses the `detailsHash` mechanism to pass through the `name`, `symbol`, and `decimals` for a representation token. On deployment, the `detailsHash` (hash of the details from the origin chain) is set but the token is always deployed with a `decimals` of 18. The decimals/other details can be updated to the correct values at any point (correctness is checked against the details hash), but this results in the number of decimals (potentially) changing while the token is deployed.

**Ideas to solve this**
Pass the raw text of the details through the transfer message, and deploy the token with the correct details up front.

**Acceptance Criteria**

- [x]  Remove `detailsHash` from `BridgeToken`, or at least remove decimals
- [x]  Update `BridgeMessage` to include original token details
- [x]  Deploy `BridgeToken` with the proper number of decimals
- [x]  Ensure the decimals can never be updated once deployed

**Other**

PR #1938, opened 2022-09-21
Fix 1805 - Consistent decimals on tokens across chains

liu-zhipeng:
## Type of change

- [x] Refactoring

## Related Issue(s)

Fixes 
https://github.com/connext/nxtp/issues/1805


packages/deployments/contracts/contracts/test/BridgeToken.sol
```diff
@@ -33,6 +33,8 @@ contract BridgeToken is IBridgeToken, OwnableUpgradeable, ERC20 {
 
   function initialize() public override initializer {
     __Ownable_init();
+
+    token.decimals = 18;
```
LayneHaber, line 37: here we need to make sure the `BridgeToken` decimals are the same as the canonical token they are representing, not just that they are always 18.

that means we have to pass that information in plaintext as part of the `BridgeMessage`, instead of only the `detailsHash` so when we deploy the token we can ensure it is set correctly

LayneHaber (commented):
the change requested here is to format the message contents so that it passes the correct number of decimals between chains. this would involve updating the `BridgeMessage` to include a `decimals` field, and ensuring the right decimals are set when the `TokenRegistry` deploys a new representation.

that being said, the motivation for this change was to *not* have the decimals change after a token is deployed / our contracts are initialized. this PR does that in a simpler way (enforcing the use of 18 decimals always), but means there would potentially be some inconsistencies between the representation and canonical/adopted assets.

im open to leaving this as-is and always using 18 decimals, but curious what your thoughts are. cc @jakekidd @rhlsthrm 

i would be a little wary to introduce this complexity, but this does alleviate my immediate concern.

packages/deployments/contracts/contracts/test/BridgeToken.sol
liu-zhipeng: Got it
I will update

LayneHaber (approved):
fixed all my qualms and tests in latest commits, ready to go if you like em!

PR #1940, opened 2022-09-21
remove: nonReentrant from xcall

### LINKED ISSUE #1704: As a dev user, I want to nest xcalls within xcalls
**Problem**

The target function of an xcall could potentially have a nested xcall in it. For example, delivering assets to a "final" domain even if most of the operational logic happens in an intermediate one. 

In general, this is needed for a multi-domain call flow like `Domain A -> Domain B -> Domain C` where a user initiates the first xcall on Domain A, some operations are done in Domain B, and some additional operations are needed on Domain C.

The problem right now is that we have a non-reentrancy modifier that prevents this.

**Ideas to solve this**

Figure out how to only allow re-entrant calls from Connext contracts.

**Acceptance Criteria**

- [ ] Xcall within an xcall works

**Other**

sanchaymittal:
## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes 
Fixes #1704

### DIFF
--- a/packages/deployments/contracts/contracts/core/connext/facets/BridgeFacet.sol
+++ b/packages/deployments/contracts/contracts/core/connext/facets/BridgeFacet.sol
@@ -376,7 +376,7 @@ contract BridgeFacet is BaseConnextFacet {
     CallParams memory _params,
     address _asset,
     uint256 _amount
-  ) internal nonReentrant whenNotPaused returns (bytes32) {
+  ) internal whenNotPaused returns (bytes32) {
     // Sanity checks.
     bytes32 remoteInstance;
     {


ArjunBhuptani:
What are the risks of doing this? We need to be *extremely* careful about reentrancy bugs given that xcall touches pretty much the entire contract state.

Perhaps as an alternative approach, we should simply stop reentrant xcalls into themselves on the same chain, but not xReceive->xcall?

LayneHaber (approved):
should make sure we add some more comments around this when were doing documentation

LayneHaber:
> What are the risks of doing this? We need to be extremely careful about reentrancy bugs given that xcall touches pretty much the entire contract state.

> Perhaps as an alternative approach, we should simply stop reentrant xcalls into themselves on the same chain, but not xReceive->xcall?

This is something im much more comfortable discussing with security experts. Ideally, we push this as the initial implementation, and outline other options we've thought about in the docs delivered to the auditors.

I'm not entirely sure I understand your suggestion though.

PR #1946, opened 2022-09-22
Condense logic for xcall messaging helper(s); reduce excessive delegate calls

jakekidd:
## Description

- reduce redundant token registry calls
- deduce whether the input asset is a canonical asset, and use this to determine whether we should burn the local asset (i.e. whether it is representative)

### DIFF
--- a/packages/deployments/contracts/contracts/core/connext/facets/BridgeFacet.sol
+++ b/packages/deployments/contracts/contracts/core/connext/facets/BridgeFacet.sol
@@ -47,6 +47,7 @@ contract BridgeFacet is BaseConnextFacet {
   error BridgeFacet__xcall_emptyTo();
   error BridgeFacet__xcall_notSupportedAsset();
   error BridgeFacet__xcall_invalidSlippage();
+  error BridgeFacet__xcall_canonicalAssetNotReceived();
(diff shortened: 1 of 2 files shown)


packages/deployments/contracts/contracts/core/connext/facets/BridgeFacet.sol
```diff
@@ -513,17 +513,23 @@ contract BridgeFacet is BaseConnextFacet {
...
+          // Input asset is an adopted asset.
+          // It could be the canonical asset, or it could be a representational asset if the
+          // representational asset is adopted on this domain.
+
+          // Retrieve the local asset address.
           local = s.tokenRegistry.getLocalAddress(canonical.domain, canonical.id);
+          // Determine whether this is the canonical asset. If the TokenRegistry returned the input
+          // asset as the local asset, and we're on the canonical domain, then this must be the
+          // canonical asset.
+          isCanonical = _params.originDomain == canonical.domain && local == _asset;
```
jakekidd, line 532: pls make sure this logic makes sense to you

jakekidd: @rhlsthrm

rhlsthrm (approved):

PR #1947, opened 2022-09-22
1932 protocol roles

### LINKED ISSUE #1932: As a team, we need admin roles for secure product rollout
**Problem**

A safe launch is the best launch, and we need to ensure we have the proper tools in place to:
- Allow bots to easily pause the contracts
- Cap the amount of a given asset that goes into the system
- Allow routers to self-manage

**Ideas to solve this**
We need the following feature in this story:

*Admin Roles vs Owner*
There should be different roles within the protocol. Specifically there should be a role for protocol admin (manages protocol parameters, upgrades, and whitelists), watchers (disconnect connectors, pause contracts), and routers.

See the policies outlined in [Security Requirements](https://www.notion.so/Security-Requirements-218f692c9cf84e79931e871b2efc4b77) for further details.

**Acceptance Criteria**

- [ ] Admin Roles
   - [x]  Create roles within the `ProposedOwnableFacet`
   - [x]  Create function for bestowing `watcher` and `router` role
   - [x]  Permission router whitelist to `router` role
   - [x]  Permission pausing to `watcher` role
   - [x]  Permission all else to `protocolAdmin` role

**Other**
Note liquidity caps handled in #1806

sanchaymittal:
## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes #1932


packages/deployments/contracts/contracts/core/connext/facets/ProposedOwnableFacet.sol
```diff
@@ -245,12 +254,48 @@ contract ProposedOwnableFacet is BaseConnextFacet, IProposedOwnable {
     _setOwner(s._proposed);
   }
 
-  function pause() public onlyOwner {
+  /**
+   * @notice Assign candidate -> router role
+   */
```
LayneHaber, line 259: there are not enough comments here about what the role does or who is allowed to assign it, and why it exists. also missing natspec

packages/deployments/contracts/contracts/core/connext/facets/ProposedOwnableFacet.sol
```diff
@@ -245,12 +254,48 @@ contract ProposedOwnableFacet is BaseConnextFacet, IProposedOwnable {
...
+    if (s.routerRole[routerRoleCandidate]) revert ProposedOwnableFacet__assignRouterRole_alreadyAdded();
+
+    s.routerRole[routerRoleCandidate] = true;
+    emit AssignRouterRole(routerRoleCandidate, msg.sender);
+  }
+
+  /**
+   * @notice Assign candidate -> watcher role
+   */
+  function assignWatcherRole(address watcherRoleCandidate) public onlyOwnerOrWatcher {
```
LayneHaber, line 272: this should be `onlyOwnerOrAdmin`

LayneHaber (changes requested):
needs tests and additional documentation. ideally at the facet level you could mention the purpose for the roles, and the breakdown of responsibilities

packages/deployments/contracts/contracts/core/connext/facets/ProposedOwnableFacet.sol
sanchaymittal: missed the comments big time, fixed

packages/deployments/contracts/contracts/core/connext/facets/ProposedOwnableFacet.sol
sanchaymittal: fixed!

sanchaymittal:
@LayneHaber  pushed the unit test and made changes as per review. Not able to check the coverage, so may need to check if more unit test is needed.

LayneHaber (changes requested):
should have an enum a la OZ roles implementation instead of 3 mappings

LayneHaber (approved):

packages/deployments/contracts/contracts/core/connext/facets/ProposedOwnableFacet.sol
```diff
@@ -245,12 +264,61 @@ contract ProposedOwnableFacet is BaseConnextFacet, IProposedOwnable {
...
+  /**
+   * @notice Use to assign an address Watcher role
+   * Address with Watcher role has permission to pause
+   * Can only be called by Owner or Role.Admin
+   * @dev requested address will be whitelisted as Role.Watcher under roles mapping
+   * @param _watcher - The address to be assigned as Role.Watcher under roles
+   */
+  function assignRoleWatcher(address _watcher) public onlyOwnerOrAdmin {
+    // Use contract as source of truth
+    // Will fail if candidate is already added
```
liu-zhipeng, line 293: We don't need the functions to renounce or remove roles, once assign roles to watcher or router, we can't remove role from them.

packages/deployments/contracts/contracts/core/connext/facets/ProposedOwnableFacet.sol
```diff
@@ -245,12 +264,61 @@ contract ProposedOwnableFacet is BaseConnextFacet, IProposedOwnable {
...
-  function pause() public onlyOwner {
+  /**
+   * @notice Use to assign an address Router role
+   * Address with Router has permission to add new router
+   * and assign addresses as Role.Router
+   * Can only be called by Owner or Role.Router
+   * @dev requested address will be whitelisted as Role.Router under roles mapping
+   * @param _router - The address to be assigned as Role.Router under roles
+   */
+  function assignRoleRouter(address _router) public onlyOwnerOrRouter {
```
jakekidd, line 275: Under this construction, the same EOA can't have multiple roles (i.e. be a router AND a watcher)... is that okay? Is that what we want? Is there a good reason as to why a singular EOA shouldn't be able to have multiple roles? (Seems like it's just inconvenient, but idk)

packages/deployments/contracts/contracts/core/connext/facets/ProposedOwnableFacet.sol
jakekidd: Was this comment intended for a different line? This function seems to be adding the role of watcher.

Regardless, with this setup, an owner *could* "renounce" or "remove" a role simply by assigning something else. For example, an owner could assign the role of router to an EOA with `assignRoleRouter(EOA)` and then later call `assignRoleWatcher(EOA)`. This would make the EOA a watcher instead of a router, effectively revoking their router permissions (even if it's for a diff permission set, it could have destructive effects).

You could fix this by changing the checks at the beginning of these assign methods to
```solidity
if (s.roles[_watcher] != address(0)) revert ProposedOwnableFacet__assignRoleWatcher_alreadyAdded();
```

but there's another problem with that - if we do it that way, and there's no way to *actually* revoke the role, then an EOA will permanently be made a router or watcher (you can't change it back).

packages/deployments/contracts/contracts/core/connext/facets/ProposedOwnableFacet.sol
jakekidd: Also, it seems weird that a router can just add any other arbitrary router (since the permissions are `onlyOwnerOrRouter`)? Is that normal?

packages/deployments/contracts/contracts/core/connext/facets/ProposedOwnableFacet.sol
```diff
@@ -245,12 +264,61 @@ contract ProposedOwnableFacet is BaseConnextFacet, IProposedOwnable {
...
+  }
+
+  /**
+   * @notice Use to assign an address Admin role
+   * Address with Admin role has permission to all else of Router & Watcher role
+   * Can only be called by Owner or Role.Admin
+   * @dev requested address will be whitelisted as Role.Admin under roles mapping
+   * @param _admin - The address to beassigned as Role.Admin under roles
+   */
+  function assignRoleAdmin(address _admin) public onlyOwnerOrAdmin {
```
jakekidd, line 307: I feel like we should have role acceptance methods (i.e. use the pending -> accepted pattern).

If not, at the very least, we probs wanna sanity check that the input address isn't `address(0)` at least, since having s.roles[0x0] = Roles.Admin could have bad consequences lol

packages/deployments/contracts/contracts_forge/core/connext/facets/ProposedOwnableFacet.t.sol
```diff
@@ -132,6 +141,45 @@ contract ProposedOwnableFacetTest is ProposedOwnableFacet, FacetHelper {
     utils_acceptNewOwnerAndAssert(_proposed);
   }
 
+  function utils_assignRoleRouter(address routerAgent, address caller) public {
```
jakekidd, line 144: this isn't a great application for fuzzing imho - I think the only real addresses we need to test for to cover all possible cases are:
- caller == randomAddressA and routerAgent == randomAddressB
- caller == routerAgent == randomAddressA
- caller == 0x0
- routerAgent == 0x0
- caller == routerAgent == 0x0

fuzzing uses a decent amount of resources, so we should pick and choose carefully where we do it

lmk tho if you have some other reasoning here!

packages/deployments/contracts/contracts_forge/core/connext/facets/ProposedOwnableFacet.t.sol
```diff
@@ -132,6 +141,45 @@ contract ProposedOwnableFacetTest is ProposedOwnableFacet, FacetHelper {
...
+    vm.prank(caller);
+    this.assignRoleAdmin(adminAgent);
+
+    if (this.queryRole(adminAgent) == Role.Admin) {
+      assertTrue(true);
+    } else {
+      assertTrue(false);
+    }
+  }
+
```
jakekidd, line 182: Ideally should have tests for fail cases, as well as have some tests around the intended/desired behavior when a candidate address is switched to a different role from a previous one OR we attempt to assign a candidate to multiple roles (not sure what the desired behavior is out of the two of these...)

packages/deployments/contracts/contracts/core/connext/facets/ProposedOwnableFacet.sol
sanchaymittal: hmm, we should definitely have a function revoke the added role. Will add a function for it.

Also, the above-shared check is correct, will fix it in the code.

packages/deployments/contracts/contracts/core/connext/facets/ProposedOwnableFacet.sol
sanchaymittal: I think it makes sense to keep, a 1-1 relation between address and role.

Hmm, I think it would make more sense if the permission to add router falls under admin only.

packages/deployments/contracts/contracts/core/connext/facets/ProposedOwnableFacet.sol
sanchaymittal: will add sanity check for addressZero

packages/deployments/contracts/contracts_forge/core/connext/facets/ProposedOwnableFacet.t.sol
sanchaymittal: once we addressZero sanity check, then it should cover all the fuzz cases.

packages/deployments/contracts/contracts_forge/core/connext/facets/ProposedOwnableFacet.t.sol
sanchaymittal: we do have test for failed cases, will add more after making the above changes.

liu-zhipeng (approved):

liu-zhipeng (approved):

LayneHaber (approved):

PR #1977, opened 2022-09-28
1966 gelato 1balance

preethamr:
## Description

Use Gelato SDK in Lighthouse

## Type of change

- [ ] Refactoring

### DIFF
--- a/packages/agents/lighthouse/src/tasks/prover/adapters/relayer/gelato.ts
+++ b/packages/agents/lighthouse/src/tasks/prover/adapters/relayer/gelato.ts
@@ -1,9 +1,12 @@
 import { createLoggingContext } from "@connext/nxtp-utils";
-import { AxiosError } from "axios";
+import { GelatoRelaySDK } from "@gelatonetwork/relay-sdk";
 
 import { RelayerSendFailed } from "../../../../errors";
-import { getGelatoRelayerAddress, isChainSupportedByGelato, gelatoSend } from "../../../../mockable";
+import { getGelatoRelayerAddress, isChainSupportedByGelato } from "../../../../mockable";
 import { getContext } from "../../prover";
+
+// TODO: Use secret
+const sponsorApiKey = "xxx";
 
 export const getRelayerAddress = async (chainId: number): Promise<string> => {
   const { logger } = getContext();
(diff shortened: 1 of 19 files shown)


packages/agents/lighthouse/src/tasks/prover/adapters/relayer/gelato.ts
```diff
@@ -12,37 +12,29 @@ export const getRelayerAddress = async (chainId: number): Promise<string> => {
...
+    throw new RelayerSendFailed({ response: response });
   } else {
-    const { taskId } = result;
-    logger.info("Sent to Gelato network", requestContext, methodContext, {
-      result,
-      taskId,
-      // response: response.data,
-    });
+    logger.info("Sent to Gelato network", requestContext, methodContext, response);
     // TODO: replace this with transactionHash eventually
```
rhlsthrm, line 37: I think it's okay to return `taskId` here instead of tranasactionHash. We should have another method in this adapter to query transactionHash where we can use that to monitor progress.

packages/agents/lighthouse/src/tasks/prover/adapters/relayer/gelato.ts
```diff
@@ -1,8 +1,8 @@
 import { createLoggingContext } from "@connext/nxtp-utils";
-import { AxiosError } from "axios";
+import { GelatoRelaySDK } from "@gelatonetwork/relay-sdk";
```
rhlsthrm, line 2: I think it would be much better to refactor this into a separate adapter package, since it's used across multiple agents and it can be generalized.

rhlsthrm (changes requested):
Looks good, requesting some changes.

packages/agents/lighthouse/src/tasks/prover/adapters/relayer/gelato.ts
preethamr: Done.
Added a method to get transactionHash

packages/agents/lighthouse/src/tasks/prover/adapters/relayer/gelato.ts
preethamr: Done.
Refactored this into relayer in utils, where all generalized gelato helpers are.

Issue #1806, opened 2022-09-07
As a team, we need liquidity caps for secure product rollout


LayneHaber:
**Problem**

A safe launch is the best launch, and we need to ensure we have the proper tools in place to:
- Allow bots to easily pause the contracts
- Cap the amount of a given asset that goes into the system
- Allow routers to self-manage

**Ideas to solve this**
We need the following features:

*Liquidity Caps*
There should be a cap on the amount of liquidity custodied within the connext contracts. These caps should be set when an asset is whitelisted, and should be able to be adjusted by protocol admin independently.

*Admin Roles vs Owner*
There should be different roles within the protocol. Specifically there should be a role for protocol admin (manages protocol parameters, upgrades, and whitelists), watchers (disconnect connectors, pause contracts), and routers.

See the policies outlined in [Security Requirements](https://www.notion.so/Security-Requirements-218f692c9cf84e79931e871b2efc4b77) for further details.

**Acceptance Criteria**

- [ ] Liquidity Caps
   - [ ]  Add `cap` to `setupAsset`
   - [ ]  Enforce custodying of canonical assets on `xcall` (or `BridgeRouter` if this is done pre-combination)
   - [ ]  Add `updateAssetCap` function

- [ ] Admin Roles
   - [ ]  Create roles within the `ProposedOwnableFacet`
   - [ ]  Create function for bestowing `watcher` and `router` role
   - [ ]  Permission router whitelist to `router` role
   - [ ]  Permission pausing to `watcher` role
   - [ ]  Permission all else to `protocolAdmin` role

**Other**

PR #1991, opened 2022-09-29
1806 token registry

LayneHaber:
## Description

Precursor to #1806 

- Moves `TokenRegistry` logic to diamond
- Sets details by owner, not via details hash
- Removes details hash mechanism (from `BridgeMessage`, `BridgeToken`, etc.)
- Removes `decimals` from message
- Deploys tokens on whitelist, not on `reconcile`

## Type of change

- [x] Refactoring

## Related Issue(s)

Fixes


packages/deployments/contracts/contracts/core/connext/facets/BaseConnextFacet.sol
```diff
@@ -135,26 +137,27 @@ contract BaseConnextFacet {
   }
 
   /**
-   * @notice Calculates a transferId
+   * @notice Returns the adopted assets for given canonical information
    */
-  function _calculateTransferId(CallParams memory _params) internal pure returns (bytes32) {
-    return keccak256(abi.encode(_params));
+  function _getRepresentationAsset(bytes32 _canonicalId, uint32 _canonicalDomain) internal view returns (address) {
+    return _getRepresentationAsset(AssetLogic.calculateCanonicalHash(_canonicalId, _canonicalDomain));
```
jakekidd, line 143: nit: imo it'd be better to require that the caller calculate the canonical hash separately ahead of time (i.e. *only* provide the `_getRepresentationAsset(bytes32 _key)` function below and no convenience functions like this). That way we don't end up doing redundant work.

more of a nit since it's really not that expensive to hash things.. just kinda seems best practice to me

packages/deployments/contracts/contracts/core/connext/facets/BridgeFacet.sol
```diff
@@ -965,7 +917,7 @@ contract BridgeFacet is BaseConnextFacet {
     }
 
     // Format hook action.
-    bytes29 _action = BridgeMessage.formatTransfer(_amount, _detailsHash, _transferId, decimals);
+    bytes29 _action = BridgeMessage.formatTransfer(_amount, _transferId);
```
jakekidd, line 920: we don't need the details hash and decimals anymore??? love itttt

packages/deployments/contracts/contracts/core/connext/facets/RoutersFacet.sol
```diff
@@ -543,16 +543,13 @@ contract RoutersFacet is BaseConnextFacet {
     if (_amount == 0) revert RoutersFacet__addLiquidityForRouter_amountIsZero();
 
     // Get the canonical asset ID from the representation.
-    (uint32 domain, bytes32 canonicalId) = s.tokenRegistry.getTokenId(_local);
-    bytes32 key = _calculateCanonicalHash(canonicalId, domain);
+    TokenId memory canonical = _getApprovedCanonicalId(_local);
```
jakekidd, line 546: what is the difference between `_getApprovedCanonicalId` and `_getCanonicalTokenId`

jakekidd: o i see, it also checks to see if it's been approved

packages/deployments/contracts/contracts/core/connext/facets/BaseConnextFacet.sol
```diff
@@ -168,4 +171,35 @@ contract BaseConnextFacet {
...
+  {
+    _local = AssetLogic.getLocalAsset(_id, _domain, s);
+    _adopted = _getAdoptedAsset(_id, _domain);
+  }
+
+  function _isLocalOrigin(address _token) internal view returns (bool) {
+    return AssetLogic.isLocalOrigin(_token, s);
+  }
+
+  function _getApprovedCanonicalId(address _candidate) internal view returns (TokenId memory _canonical) {
```
jakekidd, line 196: functions like this are really useful if we're calling another contract - they do a bunch of needed things at once.

however, in what is essentially a library (an inherited contract), you may as well have two separate functions - one for getting the canonical token ID and the other for checking whether the asset is approved. all the logic is effectively in-lined anyway, so we might as well just make this function for checking whether the canonical ID is approved, and enforce that the caller has already called `_getCanonicalTokenId` separately

jakekidd: tldr IMO avoid over-specialized helpers idk. this is a nit lol

packages/deployments/contracts/contracts/core/connext/libraries/LibConnextStorage.sol
```diff
@@ -137,10 +136,10 @@ struct AppStorage {
   // 4
   uint32 domain;
   /**
-   * @notice The local nomad token registry.
+   * @notice UpgradeBeacon from which new token proxies will get their implementation
    */
   // 5
-  ITokenRegistry tokenRegistry;
+  address tokenBeacon;
   /**
```
liu-zhipeng, line 143: where will `tokenBeacon` be initialized? maybe we should deploy separately and init it with diamond init contract

LayneHaber: it is used as the implementation address used to deploy new bridge token proxies, which are now deployed on `setupAsset`

liu-zhipeng: Yes, but i couldn't find where `tokenBeacon` address is initialized.

LayneHaber: ahh good catch -- ill add it to the deployer

packages/deployments/contracts/contracts/core/connext/facets/BaseConnextFacet.sol
LayneHaber: agree but really its to share the helper between two facets -- the code path would be the same so i think putting them in the same helper is okay

Issue #1963, opened 2022-09-27
If validation period elapsed, use pending aggregate root in proveAndProcess


jakekidd:
**Describe the bug**
When we call `proveAndProcess`, we should be able to accept using the `pendingAggregateRoot` for proofs IFF the root has already been validated (i.e. the validation period in blocks has elapsed).

There isn't a clean cut resolution here, but a good compromise is probably in order. Some constraints:
- We shouldn't use the pendingAggregateRoot just because the validation period has elapsed. What if a relayer sends a proof tx just before the validation period elapsed? The proof - which will probs be for the currentAggregateRoot - will fail to get the pendingAggregateRoot.
- We shouldn't just switch up pendingAggregateRoot and currentAggregateRoot blindly, since that will potentially rug/grief other relayer txs in the mempool (and therefore users)
- It would suck to have a ton of separate calls just check to see that the current pendingAggregateRoot is verified, and then not do anything about it... at that point we might as well go ahead and label it as proven.

TL;DR: Some racey problems with state updating and relayer txs.

One option:
- Keep a mapping of `provenAggregateRoots`
- Caller of `proveAndProcess` supplies `bytes32 _targetAggregateRoot`, indicating which they plan on using.
- We look up the target aggregate root in `provenAggregateRoots`
- If it's not there, then we can check whether it's the `pendingAggregateRoot`; if so, check to see if the validation period has elapsed for that pending root.
    - If we use the pending and the validation period has elapsed, go ahead and add the `pendingAggregateRoot` to `provenAggregateRoots`

Additionally: would be nice to have a separate public method that literally just does that last bit: check to see if validation period has elapsed and if so, move pending root -> proven roots.

(Finally: would be great to have an optimizer that allows the caller of the proveAndProcess call to specify that the root is *definitely the pending one* so we can skip checking the proven roots and assume it's not there... but that could be racey, so idk)
    
One lingering concern: is it okay to have a mapping of proven roots for the aggregate roots on the spoke chains?

See discussion:
https://discord.com/channels/992549151134982234/1022516954986184794/1024079262489247917

Issue #1964, opened 2022-09-27
Gas optimizations


jakekidd:
## What is?
Various gas optimizations in the contract that @jakekidd found. This issue could be broken down into multiple issues for each one.

### `propagate` domains and connectors

Right now, we iterate over the `domains` in storage as a way to determine which domains to propagate the aggregateRoot to - this is costly, as we are doing 2 SLOADs per domain (1 for domain itself from the array, 1 for the corresponding connector addresses).

Instead, I propose `propagate` method take the list of domains as an argument to the method. To make sure that the list provided reflects the actual array in storage (i.e. isn't made-up by the caller), we regularly hash the array of domains in storage whenever that array (i.e. connectors) is updated and save the hash separately, so we can validate any `domains` argument for `propagate` against that hash.

The gas cost of loading the whole array from storage, which is 1 SLOAD per item, would instead be just 1 SLOAD for the hash, which represents a scalable list of domains.

There would still be O(n) calls to get the *connector* for each domain but at least we could cut everything overall in half... unless...

... unless we could also have an arg for list of connector addresses? And in storage, keep an array of hub connectors *instead* of a mapping? It would technically make sense in a way - the arrays of connectors and domains would match (i.e. a connector at index 3 in connectors would belong to domain at index 3 in domains), and we could actively track a hash/digest of both for verifying that the args provided reflect the real values stored!

### AssetLogic storage inefficiencies

There are a number of storage inefficiency issues in AssetLogic.sol.

Many instances are like the following example in `_swapAsset`:
```c++
  function _swapAsset(
    bytes32 _key,
    address _assetIn,
    address _assetOut,
    uint256 _amount,
    uint256 _minOut
  ) internal returns (uint256, address) {
    AppStorage storage s = LibConnextStorage.connextStorage();

    if (stableSwapPoolExist(_key)) {
      // if internal swap pool exists
      return (
        s.swapStorages[_key].swapInternal(
          getTokenIndexFromStableSwapPool(_key, _assetIn),
          getTokenIndexFromStableSwapPool(_key, _assetOut),
          _amount,
          _minOut
        ),
        _assetOut
      );
    }
```
We are calling `stableSwapPoolExist()` which does the lookup `s.swapStorages[_key]` ... then we are doing the same lookup again to call `swapInternal`.

Basically, we need to go through AssetLogic and evaluate which helpers are doing redundant actions.

In the example, we should just do the SLOAD once, and write it to memory:
```c++
[...]

PR #1992, opened 2022-09-29
Handle validation period for incoming roots on SpokeConnector, RootManager

jakekidd:
Fixes https://github.com/connext/nxtp/issues/1963

A subtask of the gas optimizations listed here:
https://github.com/connext/nxtp/issues/1964
(the second one in that list ^)

## Type of change

- [x] Refactoring

### Verification for aggregateRoots in SpokeConnector

- Using two mappings, `pendingAggregateRoots` and `provenAggregateRoots`
- when we call `proveAndProcess`, caller specifies a target aggregate root. If the target is in pending, we check to see if delay blocks have been surpassed for verification; if so, we move to proven

### Verification for inboundRoots in RootManager

- Added a custom `QueueLib` (similar to existing one, but with key optimizations)
- Using a queue to verify incoming inboundRoots.
- roots get dequeued and inserted into the aggregate Merkle tree on `propagate` if they have surpassed verification period
- pending roots can be removed by owner if paused

### DomainIndexer

- DomainIndexer = abstract contract used by RootManager to manage domains/connectors mutex
- There is now an array for both connectors and domains in storage.
- There is also a `connectorsHash` and `domainsHash` in storage : these are hash digests of the two respective arrays.
- The connector at a given index will belong to the domain at the same index.
- When new connectors get added/removed, we modify existing arrays as needed and then update the hashes in storage.
- Now when we call propagate, we pass in the current domains and connectors arrays.

**Result/Improvement: Instead of consuming 1600 * N gas per call (where N = number of supported domains) retrieving domain/connector info, we will consume a fixed 1600 gas per call by just checking that the given domains and connectors arrays match what we currently have in storage.**

(*1600 above comes from the cost of an SLOAD of a pre-written value (i.e. warm slot), 800 * 2, since we read once from `domains` and once from `connectors`.) 

This is just more scalable - our messaging layer should ideally grow as little as possible in gas cost with each domain we add to the system.

#### ~~Important Last Question!~~

EDIT: on further thought, this is probs not worth it- it's not much saved for a method we'll call infrequently and adds complexity to off-chain dependencies:

> One thing we could do here that would greatly simplify on-chain storage is *get rid of the domains and connectors arrays entirely*. Instead, we *just store the hashes* and any time the current domains/connectors array needs to be referenced, we just provide it in arguments. While it's convenient to have the on-chain storage for the sake of reference, both the domain and the connector are already emitted in events in `addConnector` and `removeConnector`, so we could just store the current arrays' states in the subgraph.
>
> This would reduce cost of `addConnector` and `removeConnector` (which is currently linear to N domains) *and* reduce mutex risk of having redundant data structures / references that all need updates whenever we do those operations.


packages/deployments/contracts/contracts/messaging/RootManager.sol
```diff
@@ -14,122 +16,171 @@ import {WatcherClient} from "./WatcherClient.sol";
...
+    return pendingInboundRoots.length();
+  }
+
+  // ============ Admin Functions ============
+
+  /**
+   * @notice Set the `delayBlocks`, the period in blocks over which an incoming message
+   * is verified.
+   */
+  function setDelayBlocks(uint256 _delayBlocks) public onlyOwner {
```
LayneHaber, line 100: should have an event

packages/deployments/contracts/contracts/messaging/libraries/DomainIndexer.sol
```diff
@@ -0,0 +1,160 @@
...
+  function getConnectorForDomain(uint32 _domain) public view returns (address) {
+    return connectors[getDomainIndex(_domain)];
+  }
+
+  /**
+   * @notice Validate given domains and connectors arrays are correct (i.e. they mirror what is
+   * currently saved in storage).
+   * @dev Reverts if domains or connectors do not match.
+   */
+  function validateDomains(uint32[] calldata _domains, address[] calldata _connectors) public view {
```
LayneHaber, line 85: just note that it will error if the ordering of the domains / connectors as well

packages/deployments/contracts/contracts/messaging/libraries/DomainIndexer.sol
```diff
@@ -0,0 +1,160 @@
...
+    require(!isDomainSupported(_domain), "exists");
+    // Sanity check: connector is reasonable.
+    require(_connector != address(0), "!connector");
+
+    // Push domain and connector to respective arrays.
+    domains.push(_domain);
+    connectors.push(_connector);
+    // Set reverse lookup.
+    uint256 _indexPlusOne = domains.length;
+    domainToIndexPlusOne[_domain] = _indexPlusOne;
```
LayneHaber, line 110: why not just use the index instead of `domainToIndexPlusOne` ? or is it to avoid the `0` case altogether? (if so, we should add a comment to the mapping declaration)

packages/deployments/contracts/contracts/messaging/RootManager.sol
```diff
@@ -14,122 +16,171 @@ import {WatcherClient} from "./WatcherClient.sol";
...
   // ============ Events ============
 
   event RootAggregated(uint32 domain, bytes32 receivedRoot, uint256 index);
 
-  event RootPropagated(bytes32 aggregate, uint32[] domains);
+  event RootPropagated(bytes32 aggregate, uint32[] domains, uint256 count);
 
   event ConnectorAdded(uint32 domain, address connector);
 
   event ConnectorRemoved(uint32 domain, address connector);
```
LayneHaber, line 32: because the hashes stored in the indexer change, we should make sure we emit the newly updated domain and connector arrays

LayneHaber (commented):
need to change some events, but otherwise lgtm (though we should have more tests, esp for the domain indexer)

packages/deployments/contracts/contracts/messaging/libraries/DomainIndexer.sol
jakekidd: Yep, it's a sneaky shortcut to avoid 0-case so it's a dual purpose mapping. There is a comment in the mapping declaration about it, but lemme make sure it's highlighted

LayneHaber: oh must've missed it in the diff -- its in there, nice

LayneHaber (approved):

PR #2027, opened 2022-10-08
Add zksync connector

rhlsthrm:
## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes


LayneHaber (approved):
made some fixes directly before reviewing. summary:

- make `processMessage` a null-op on the hub connector to ensure all messages are proven
- make `verifySender` always return false on hub connector to reflect the fact that cross domain senders cannot be verified in that way
- make `verifySender` check the `msg.sender` on the spoke connector, as suggested by tutorials in zk sync docs

Issue #2133, opened 2022-10-18
RootAggregated event is misleading for indexing leaves (index off-by-X)


jakekidd:
**Describe the bug**
In a sparse merkle tree, each leaf has an index in the tree.

In our naive implementation of the messaging layer, incoming message roots were immediately aggregated into the aggregate tree on the hub; now they go through a validation period by sitting in a pendingInboundRoots queue for a set number of `delayBlocks`.

This means the `RootAggregated` event no longer emits the index of the leaves in the tree, but the index in the queue. The index in the queue will be the index in the tree +1 (we currently account for this in our off-chain code). However! If any of the message roots turn out to be fraudulent, that off-by-one issue will become off-by-two.

Basically, the core issue is that the subgraph treats the `RootAggregated` event's index as the tree index.

**How to fix?**

First thing: we should change the `RootPropagated` event to give us the info we need. This will unfortunately increase gas costs, but it really is the simplest solution:
```
-  event RootPropagated(bytes32 aggregate, uint32[] domains, uint256 count);
+  event RootPropagated(bytes32 aggregateRoot, uint256 count, uint32[] domains, bytes32[] aggregatedMessageRoots);
```
From here it's easy enough to determine the tree index of each aggregated message root. There's some hidden complexity though - since we just changed the subgraph key for OriginMessage to be queueIndex from originally being the message root itself (due to redundant/duplicate message roots being committed)... that will have to be resolved.

From here we have 2 options:

1. Keep mapping `RootAggregated` event in subg into subg entity OriginMessage, but change `index` property to `queueIndex` property, and add a `treeIndex` property that gets set on `RootPropagated` event (it will be null until the message root has been propagated!).
2. Delete mapping for `RootAggregated` event. Instead, we don't create an OriginMessage entity *until* that message gets propagated (i.e. included in the tree). That way the OriginMessage.index reflects the *actual* index in the tree. Queue index is no longer relevant.

I think I'm in favor of (2) since it seems the clearest to me (and we can circumvent the aforementioned hidden complexity).

**To Reproduce**
1. Submit fraudulent root
2. Remove from queue
3. Indices for OriginMessages in subg will no longer be off-by-1, they will be off-by-2.

**Expected behavior**
In the subgraph, OriginMessage.index (or OriginMessage.treeIndex) should reflect the tree index exactly - no off-by-X issues.

PR #2136, opened 2022-10-19
Emit aggregated message roots in RootManager.propagate; track AggregatedMessageRoots in subg

jakekidd:
## Description
- index that is tracked in our DB and subgraph is off by 1 since it's the queue index not the tree index
- that off-by-1 issue becomes off-by-2 if there's ever a fraudulent root removed from the queue
- simplest solution: don't track message root indices in the subgraph until they are actually inserted into the tree

## Type of change

- [x] Refactoring

Used solution 2 in the named issue below

## Related Issue(s)

Fixes https://github.com/connext/nxtp/issues/2133


jakekidd:
should have been draft, DNM

Current status: made the on-chain fixes in RootManager and attempted to update subgraph mapping under amarok-hub-staging/mapping/rootManager.ts (and schema and abi updated)

Still need to probs fix subg integration with DB / etc - idk how we were querying RootAggregated and RootPropagated entries in the subg but we should now query AggregatedMessageRoot to get the **accurate index** of the message root in the tree.

packages/deployments/subgraph/src/amarok-hub-staging/schema.graphql
```diff
@@ -13,6 +13,12 @@ type RootPropagated @entity {
   count: BigInt!
 }
 
+type AggregatedMessageRoot @entity {
+  id: ID! # Same as index.
```
wanglonghong, line 17: if we make id same as index here, duplication issue may happen.

wanglonghong (changes requested):

packages/deployments/subgraph/src/amarok-hub-staging/schema.graphql
wanglonghong: will update them in next commits

rhlsthrm (approved):

wanglonghong (approved):

PR #2193, opened 2022-10-22
Spearbit 7 fixes

jakekidd:
## Type of change

- [x] Refactoring
- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Related Issue(s)

Fixes

### DIFF
--- a/packages/deployments/contracts/contracts/core/connext/facets/SwapAdminFacet.sol
+++ b/packages/deployments/contracts/contracts/core/connext/facets/SwapAdminFacet.sol
@@ -33,6 +33,8 @@ contract SwapAdminFacet is BaseConnextFacet {
   error SwapAdminFacet__initializeSwap_feeExceedMax();
   error SwapAdminFacet__initializeSwap_adminFeeExceedMax();
   error SwapAdminFacet__initializeSwap_failedInitLpTokenClone();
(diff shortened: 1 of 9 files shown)


packages/deployments/contracts/contracts/core/connext/facets/SwapAdminFacet.sol
```diff
@@ -176,6 +185,34 @@ contract SwapAdminFacet is BaseConnextFacet {
...
+      if (s.swapStorages[_key].adminFees[i] > 0) {
+        s.swapStorages[_key].withdrawAdminFees(msg.sender);
+      }
+
+      unchecked {
+        ++i;
+      }
+    }
+
+    delete s.swapStorages[_key];
```
jakekidd, line 212: I think this `delete` operation won't clear the nested array properties in `Swap`, so removing a swap and recreating one with the same key could result in reusing old references (such as for `IERC20[] pooledTokens`) unless they are explicitly overwritten when we do the recreation.

In the loop above (`for (uint8 i; i < numPooledTokens; ) {`) we should probs just delete entries in all of the nested arrays (pooledTokens, tokenPrecisionMultipliers, balances, and adminFees) to make sure everything is cleaned out.

PR #2199, opened 2022-10-24
Spearbit 36

jakekidd:
## Description

See spearbit comment here: https://github.com/spearbit-audits/connext-nxtp/pull/5#discussion_r998950182

## Type of change

- [ ] Refactoring

Prevent DoSing redundant (unnecessary) message roots being sent to hub. Implement a rate limiting feature that allows owner to set a number of blocks messages must elapse between send events.


packages/deployments/contracts/contracts/messaging/libraries/RateLimited.sol
```diff
@@ -0,0 +1,61 @@
...
+  // ============ Modifiers ============
+
+  /**
+   * @notice Checks to see if we can send this block, given the current rate limit
+   * setting and the last block we sent a message. If rate limit has been surpassed,
+   * we update the `lastSentBlock` to be the current block.
+   */
+  modifier rateLimited() {
+    // Check to make sure we have surpassed the number of rate limit blocks.
+    if (lastSentBlock + rateLimitBlocks > block.number) {
```
wanglonghong, line 41: it looks wrong to me. shouldn't it be like the following?!?!
`if (lastSentBlock + rateLimitBlocks < block.number)`

wanglonghong (changes requested):

packages/deployments/contracts/contracts/messaging/libraries/RateLimited.sol
rhlsthrm: Actually I think @jakekidd is correct. If the last sent block plus the rate limit blocks is larger than the current block it means not enough blocks have passed, so you want to revert.

packages/deployments/contracts/contracts/messaging/connectors/SpokeConnector.sol
```diff
@@ -255,7 +272,7 @@ abstract contract SpokeConnector is Connector, ConnectorManager, WatcherClient,
    * @notice This returns the root of all messages with the origin domain as this domain (i.e.
    * all outbound messages)
    */
-  function send() external whenNotPaused {
+  function send() external whenNotPaused rateLimited {
     bytes memory _data = abi.encodePacked(MERKLE.root());
```
rhlsthrm, line 276: Don't we need to actually check against `sentMessageRoots`? I think this is incomplete.

rhlsthrm (changes requested):

packages/deployments/contracts/contracts/messaging/connectors/SpokeConnector.sol
rhlsthrm: I fixed this, blz check @jakekidd

wanglonghong (approved):

packages/deployments/contracts/contracts/messaging/connectors/SpokeConnector.sol
jakekidd: yep, gob job, ty

rhlsthrm (approved):

packages/deployments/contracts/contracts/messaging/libraries/Queue.sol
```diff
@@ -54,38 +54,57 @@ library QueueLib {
...
+
+    {
+      // If we would otherwise be searching beyond the maximum amount we are allowed to dequeue in this
+      // call, reduce `last` to artificially shrink the available queue within the scope of this method.
+      uint128 highestAllowed = first + max - 1;
+      if (last > highestAllowed) {
+        last = highestAllowed;
+      }
+    }
+
```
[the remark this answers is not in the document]
csanuragjain: @gpersoon 
In this case since last>highestAllowed (100_000>(1+100-1)) {last=100}
So loop should run 100 times

packages/deployments/contracts/contracts/messaging/RootManager.sol
```diff
@@ -184,6 +185,32 @@ contract RootManager is ProposedOwnable, IRootManager, WatcherClient, DomainInde
...
+   * @return bytes32 The new aggregate root.
+   * @return uint256 The updated count (number of leaves).
+   */
+  function dequeue() public whenNotPaused returns (bytes32, uint256) {
+    // Get all of the verified roots from the queue.
+    bytes32[] memory _verifiedInboundRoots = pendingInboundRoots.dequeueVerified(delayBlocks, DEQUEUE_MAX);
+
+    // Sanity check: there must be some verified roots to aggregate and send: otherwise we would be
+    // propagating a redundant aggregate root.
+    require(_verifiedInboundRoots.length != 0, "no verified roots");
```
gpersoon, line 206: This can result in a DOS vector.
Suppose someone repeatedly calls `dequeue()` until the `queue` is empty (which can be done because its permissionless).

Then a relayer calls `propagate()`, which calls `dequeue()`.
As `dequeue()` reverts (because the queue is empty), `propagate()` will also revert.

Possible solution: just return when `_verifiedInboundRoots.length == 0`

jakekidd: Oh excellent point! Yes, I think we can just return here when the queue is empty.

One issue this presents - propagate can be called even if no new roots have been added since the last propagate. I think I may need to add a flag to ensure we aren't calling a redundant propagate, since this check here:
```require(_verifiedInboundRoots.length != 0, "no verified roots");```
was previously the thing that prevented that from happening.

PR #2211, opened 2022-10-25
Spearbit 164

jakekidd:
## Description

See https://github.com/spearbit-audits/connext-nxtp/pull/7#discussion_r999939295


jakekidd:
NOTE: tests still need to be updated. Also, double-check the assembly code implementation! I left a TODO at a possible error in the code?

packages/deployments/contracts/contracts/messaging/libraries/Merkle.sol
```diff
@@ -16,6 +16,48 @@ library MerkleLib {
   uint256 internal constant TREE_DEPTH = 32;
   uint256 internal constant MAX_LEAVES = 2**TREE_DEPTH - 1;
 
+  /**
+   * @dev Z_i repsesent the hash values at different heights for a binary tree with leaf values equal to `0`.
+   * (e.g. Z_1 is the keccak256 hash of (0x0, 0x0), Z_2 is the keccak256 hash of (Z_1, Z_1), etc...)
+   * Z_0 is the bottom of the 32-layer tree, Z_31 is the top.
```
Saw-mon-and-Natalie, line 22: If we would consider the root layer to be the top layer than we have `33` layers and the top one would be `Z_32`.

packages/deployments/contracts/contracts/messaging/libraries/Merkle.sol
```diff
@@ -16,6 +16,48 @@ library MerkleLib {
...
+  bytes32 internal constant Z_24 = hex"cdc72595f74c7b1043d0e1ffbab734648c838dfb0527d971b602bc216c9619ef";
+  bytes32 internal constant Z_25 = hex"0abf5ac974a1ed57f4050aa510dd9c74f508277b39d7973bb2dfccc5eeb0618d";
+  bytes32 internal constant Z_26 = hex"b8cd74046ff337f0a7bf2c8e03e10f642c1886798d71806ab1e888d9e5ee87d0";
+  bytes32 internal constant Z_27 = hex"838c5655cb21c6cb83313b5a631175dff4963772cce9108188b34ac87c81c41e";
+  bytes32 internal constant Z_28 = hex"662ee4dd2dd7b2bc707961b1e646c4047669dcb6584f0d8d770daf5d7e7deb2e";
+  bytes32 internal constant Z_29 = hex"388ab20e2573d171a88108e79d820e98f26c0b84aa8b2f4aa4968dbb818ea322";
+  bytes32 internal constant Z_30 = hex"93237c50ba75ee485f4c22adf2f741400bdf8d6a9cc7df7ecae576221665d735";
+  bytes32 internal constant Z_31 = hex"8448818bb4ae4562849e949e17ac16e0be16688e156b5cf15e098c627c0056a9";
+  bytes32 internal constant Z_32 = hex"27ae5ba08d7291c96c8cbddcc148bf48a6d68c7974b94356f53754ef6171d757";
+
```
Saw-mon-and-Natalie, line 58: add differential tests for these values.

packages/deployments/contracts/contracts/messaging/libraries/Merkle.sol
```diff
@@ -34,52 +76,11 @@ library MerkleLib {
...
-   * @dev Reverts if the tree is already full.
-   * @param node Element to insert into tree.
-   * @return uint256 Updated count (number of nodes in the tree).
-   **/
-  function insert(Tree storage tree, bytes32 node) internal returns (uint256) {
-    uint256 size = tree.count + 1; // Add 1 since we'll be including a new node.
+  function insert(Tree memory tree, bytes32 node) internal pure returns (Tree memory) {
+    // Update tree.count to increase the current count by 1 since we'll be including a new node.
+    uint256 size = ++tree.count;
     if (size >= MAX_LEAVES) revert MerkleLib__insert_treeIsFull();
```
Saw-mon-and-Natalie, line 82: We can change the condition to `size > MAX_LEAVES` since `MAX_LEAVES` is $2^{32}-1$.

Also note in the deposit contract the check for `MAX_LEAVES` comes before updating the `tree.count` which can save gas on paths that would revert. If we move this check to the top of the function then can replace the condition with `size > (MAX_LEAVES - 1)`.

https://github.com/ethereum/deposit_contract/commit/e357f3797bd37116564fb1e41605864d6fffb45f

Saw-mon-and-Natalie:
forge tests fail:

```bash
error[9553]: TypeError: Invalid type for argument in function call. Invalid implicit conversion from bytes32 to bytes32[] memory requested.
   --> contracts/messaging/connectors/SpokeConnector.sol:296:53:
    |
296 |     (bytes32 _root, uint256 _count) = MERKLE.insert(_messageHash);
    |                                                     ^^^^^^^^^^^^



error[9553]: TypeError: Invalid type for argument in function call. Invalid implicit conversion from bytes32 to bytes32[] memory requested.
  --> contracts_forge/messaging/Merkle.t.sol:22:34:
   |
22 |       (, _count) = merkle.insert(_messageHash);
   |                                  ^^^^^^^^^^^^



error[9553]: TypeError: Invalid type for argument in function call. Invalid implicit conversion from bytes32 to bytes32[] memory requested.
   --> contracts_forge/messaging/PingPong.t.sol:227:79:
    |
227 |     (bytes32 expectedRoot, uint256 expectedCount) = referenceSpokeTree.insert(messageHash);
    |                                                                               ^^^^^^^^^^^



error[9582]: TypeError: Member "zeroHashes" not found or not visible after argument-dependent lookup in type(library MerkleLib).
   --> contracts_forge/messaging/PingPong.t.sol:383:39:
    |
383 |     bytes32[32] memory messageProof = MerkleLib.zeroHashes();
    |                                       ^^^^^^^^^^^^^^^^^^^^

```

`MERKLE.insert` only accepts an array of leaves, the single leaf endpoint has been removed. And `MerkleLib.zeroHashes()` does not exist anymore.

packages/deployments/contracts/contracts/messaging/libraries/Merkle.sol
```diff
@@ -104,36 +105,727 @@ library MerkleLib {
...
+
   /**
    * @notice Calculates and returns tree's current root.
-   * @return bytes32 root.
+   * @return _current bytes32 root.
    **/
-  function root(Tree storage tree) internal view returns (bytes32) {
-    return rootWithCtx(tree, zeroHashes());
-  }
+  function root(Tree storage tree) internal view returns (bytes32 _current) {
```
Saw-mon-and-Natalie, line 114: add differential tests for `root`.

packages/deployments/contracts/contracts/messaging/libraries/Merkle.sol
```diff
@@ -151,91 +843,170 @@ library MerkleLib {
...
+
+      f := shl(5, and(_index, shl(30, 1)))
+      mstore(f, _current)
+      mstore(sub(0x20, f), mload(add(BRANCH_DATA_OFFSET, shl(5, 30))))
+      _current := keccak256(0, 0x40)
+
+      f := shl(5, and(_index, shl(31, 1)))
+      mstore(f, _current)
+      mstore(sub(0x20, f), mload(add(BRANCH_DATA_OFFSET, shl(5, 31))))
+      _current := keccak256(0, 0x40)
```
Saw-mon-and-Natalie, line 1009: This part of the assembly block needs to be modified. For more info check https://github.com/spearbit-audits/connext-nxtp/issues/164#issuecomment-1307507146

packages/deployments/contracts/contracts/messaging/libraries/Merkle.sol
jakekidd: Gas cost on path that would revert due to going over max leaves is negligible. We have far greater concerns in place - namely, recycling this tree. But I will definitely change the condition to be `size > MAX_LEAVES`

packages/deployments/contracts/contracts/messaging/libraries/Merkle.sol
jakekidd: @Saw-mon-and-Natalie just fixed this - pls take a look when you get a chance!

packages/deployments/contracts/contracts/messaging/libraries/MerkleLib.sol
```diff
@@ -16,6 +16,48 @@ library MerkleLib {
   uint256 internal constant TREE_DEPTH = 32;
   uint256 internal constant MAX_LEAVES = 2**TREE_DEPTH - 1;
 
+  /**
+   * @dev Z_i repsesent the hash values at different heights for a binary tree with leaf values equal to `0`.
```
gpersoon, line 20: typo repsesent

gpersoon:
@LayneHaber / @jakekidd I assume YUL won't be used (as in https://github.com/spearbit-audits/connext-nxtp/issues/167?)
This PR still contains YUL but also other improvements like building the `tree` in memory first.

jakekidd:
> @LayneHaber / @jakekidd I assume YUL won't be used (as in [spearbit-audits/connext-nxtp#167](https://github.com/spearbit-audits/connext-nxtp/issues/167)?) This PR still contains YUL but also other improvements like building the `tree` in memory first.

If you're referring to `insert`, we've acknowledged that issue and plan on moving forward without YUL for that method for now. Short-term timeline and cost/benefit analysis resulted in us sidelining that one vs others like the modifications we've made (included in the "cost" here is the fact that we need to understand this assembly code well in order to maintain it!).

This PR contains a bunch of overlapping fixes since a lot of them were dependent on one another.

packages/deployments/contracts/contracts/messaging/libraries/MerkleLib.sol
```diff
@@ -104,36 +105,734 @@ library MerkleLib {
...
 
-  /**
-   * @notice Calculates and returns tree's current root given array of zero hashes.
-   * @param _zeroes Array of zero hashes.
-   * @return _current Calculated root of tree.
-   **/
-  function rootWithCtx(Tree storage tree, bytes32[TREE_DEPTH] memory _zeroes) internal view returns (bytes32 _current) {
-    uint256 _index = tree.count;
+    uint256 i;
+    assembly {
```
gpersoon, line 122: So this should be replaced with the original Solidity code  (assuming YUL won't be used)

jakekidd: @gpersoon I'm a bit confused. Why would we not use the YUL / assembly code here? We want to use it. Assuming you are referring to the `root` method here, we are calling it in the codebase. Sry, but if you could clarify what you mean here that would be great!

gpersoon: Hi Jake,
I was referring to: https://github.com/spearbit-audits/connext-nxtp/issues/167#issuecomment-1317415966
"We don't have the internal capability to maintain Yul safely, so adding a large portion to core code paths seems inadvisable"
So that is why I was confused :-(

PR #2274, opened 2022-10-28
Add health checks

rhlsthrm:
## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes


carlomazzaferro:
How often are the healthcheck pings sent? They will have to be run at a pace faster than what's configured to be received on betteruptime.

Otherwise, it is also essential that we set the configuration for the endpoint URLs in the `config.tf` files!

rhlsthrm:
> How often are the healthcheck pings sent? They will have to be run at a pace faster than what's configured to be received on betteruptime.
> 
> Otherwise, it is also essential that we set the configuration for the endpoint URLs in the `config.tf` files!

Yep, I have the betteruptime configs set up properly! And correct, the config.tf files are the last change needed.

PR #2298, opened 2022-11-01
Spearbit 179

jakekidd:
## Description
see https://github.com/spearbit-audits/connext-nxtp/pull/4#discussion_r1000673896


LayneHaber (approved):
ideally tests

PR #2342, opened 2022-11-04
Spearbit 31

liu-zhipeng:
## Description
Fix https://github.com/spearbit-audits/connext-nxtp/issues/31

## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes


LayneHaber (commented):
who is able to upgrade the LP token once it is deployed? it seems it points to an implementation, then clones it. this implementation may not have the ownable implemented properly, so could set the owner to a random address.

instead, we should ensure that the `owner()` of the `LPToken` is the main contract

LayneHaber:
concern of LP token should be addressed in a separate PR, merging

PR #2353, opened 2022-11-04
Spearbit 38

liu-zhipeng:
## Description
Fix https://github.com/spearbit-audits/connext-nxtp/issues/38

## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes 
https://github.com/spearbit-audits/connext-nxtp/issues/38


LayneHaber (commented):
should add in references described in [issue](https://github.com/spearbit-audits/connext-nxtp/issues/38):

> And make sure any values of  for init functions or as an input to swap functions are scaled using  from the amplification coefficient.

> Looks like the original source for SwapUtils is from here:
https://github.com/saddle-finance/saddle-contract/blob/master/contracts/SwapUtils.sol

> which is a Solidity rewrite of the curve finance pool template contracts in Vyper:
https://github.com/curvefi/curve-contract/tree/master/contracts/pool-templates

> It is recommend adding header NatSpec comments and mentioning authors or original/modified sources for all files including the ones mentioned in this issue.

PR #2354, opened 2022-11-04
Spearbit 76, 125, 124

liu-zhipeng:
## Description
Fix https://github.com/spearbit-audits/connext-nxtp/issues/76

## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes 
https://github.com/spearbit-audits/connext-nxtp/issues/76


packages/deployments/contracts/contracts/core/connext/facets/SwapAdminFacet.sol
```diff
@@ -176,6 +185,34 @@ contract SwapAdminFacet is BaseConnextFacet {
...
+
+    if (numPooledTokens == 0) revert SwapAdminFacet__removeSwap_notInitialized();
+
+    for (uint8 i; i < numPooledTokens; ) {
+      if (s.swapStorages[_key].balances[i] > 0) {
+        revert SwapAdminFacet__removeSwap_NonZeroBalance();
+      }
+
+      if (s.swapStorages[_key].adminFees[i] > 0) {
+        s.swapStorages[_key].withdrawAdminFees(msg.sender);
```
LayneHaber, line 204: should emit event

LayneHaber (changes requested):
should emit events, and we should test this

packages/deployments/contracts/contracts/core/connext/facets/SwapAdminFacet.sol
```diff
@@ -176,6 +185,35 @@ contract SwapAdminFacet is BaseConnextFacet {
...
+
+    if (numPooledTokens == 0) revert SwapAdminFacet__removeSwap_notInitialized();
+
+    for (uint256 i; i < numPooledTokens; ) {
+      if (s.swapStorages[_key].balances[i] == 0) {
+        revert SwapAdminFacet__removeSwap_NonZeroBalance();
+      }
+
+      if (s.swapStorages[_key].adminFees[i] > 0) {
+        s.swapStorages[_key].withdrawAdminFees(msg.sender);
```
gpersoon, line 204: `withdrawAdminFees()` also loops over `pooledTokens`, so could be outside of the `for` loop

packages/deployments/contracts/contracts/core/connext/facets/SwapAdminFacet.sol
```diff
@@ -176,6 +185,35 @@ contract SwapAdminFacet is BaseConnextFacet {
...
+   *
+   * @param _key the hash of the canonical id and domain for token
+   */
+  function removeSwap(bytes32 _key) external onlyOwnerOrAdmin {
+    uint256 numPooledTokens = s.swapStorages[_key].pooledTokens.length;
+
+    if (numPooledTokens == 0) revert SwapAdminFacet__removeSwap_notInitialized();
+
+    for (uint256 i; i < numPooledTokens; ) {
+      if (s.swapStorages[_key].balances[i] == 0) {
```
gpersoon, line 199: Shouldn't it be the other way around?
e.g. only allow removal of swap when all the balances are 0?
If the balances are >0 then the tokens would be lost.

liu-zhipeng: Yes, correct!

liu-zhipeng: But there is another problem here. Once swap pool is initialized and liquidity added, it is very hard to make all balances zero. It means that in this case, swap pool can't be removed!
Do you have any idea for this?

liu-zhipeng: Remove all total supply, can make balances zero. anyway fixed, please check again

packages/deployments/contracts/contracts/core/connext/facets/SwapAdminFacet.sol
```diff
@@ -176,6 +185,35 @@ contract SwapAdminFacet is BaseConnextFacet {
     emit SwapInitialized(_key, entry, msg.sender);
   }
 
+  /**
+   * @notice remove Swap Struct for key
+   *
+   * @param _key the hash of the canonical id and domain for token
+   */
+  function removeSwap(bytes32 _key) external onlyOwnerOrAdmin {
```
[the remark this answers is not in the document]
gpersoon: This one is fixed

packages/deployments/contracts/contracts/core/connext/facets/SwapAdminFacet.sol
liu-zhipeng: Yes, correct, but `addSwapLiquidity` is not owner function, so anyone can add liquidity to swap pool. In this case, someone doesn't remove their liquidity from swap pool, owner can't remove that swap. I think that would be problem and we need a `forceRemoveSwap` function.

LayneHaber: we could take a similar approach as 

in that case:
- `forceRemoveSwap` (rename to `disableSwap`) would set a flag that indicates whether it is usable
- `initializeSwap` could still be gated to admins, no need to leave that open

then you would be able to safely remove liquidity when the pool is disabled. you wouldn't be able to clean up the storage quite as well, though maybe that could be a separate function (and is called with some enforced delay from `disableSwap`). i think that may be overkill.

LayneHaber (approved):
see above

LayneHaber (commented):
see above (didn't mean to approve previously)

packages/deployments/contracts/contracts/core/connext/facets/SwapAdminFacet.sol
```diff
@@ -33,6 +33,8 @@ contract SwapAdminFacet is BaseConnextFacet {
   error SwapAdminFacet__initializeSwap_feeExceedMax();
   error SwapAdminFacet__initializeSwap_adminFeeExceedMax();
   error SwapAdminFacet__initializeSwap_failedInitLpTokenClone();
+  error SwapAdminFacet__removeSwap_notInitialized();
+  error SwapAdminFacet__removeSwap_NonZeroBalance();
```
LayneHaber, line 37:
```suggestion
  error SwapAdminFacet__removeSwap_nonZeroBalance();
```

packages/deployments/contracts/contracts/core/connext/facets/SwapAdminFacet.sol
```diff
@@ -177,6 +186,36 @@ contract SwapAdminFacet is BaseConnextFacet {
...
+   * @param _key the hash of the canonical id and domain for token
+   */
+  function removeSwap(bytes32 _key) external onlyOwnerOrAdmin {
+    uint256 numPooledTokens = s.swapStorages[_key].pooledTokens.length;
+
+    if (numPooledTokens == 0) revert SwapAdminFacet__removeSwap_notInitialized();
+
+    for (uint256 i; i < numPooledTokens; ) {
+      if (s.swapStorages[_key].balances[i] > 0) {
+        revert SwapAdminFacet__removeSwap_NonZeroBalance();
```
LayneHaber, line 201:
```suggestion
        revert SwapAdminFacet__removeSwap_nonZeroBalance();
```

packages/deployments/contracts/contracts/core/connext/facets/SwapAdminFacet.sol
```diff
@@ -177,6 +186,36 @@ contract SwapAdminFacet is BaseConnextFacet {
...
+    }
+
+    s.swapStorages[_key].withdrawAdminFees(msg.sender);
+    emit AdminFeesWithdrawn(_key, msg.sender);
+
+    delete s.swapStorages[_key];
+
+    emit SwapRemoved(_key, msg.sender);
+  }
+
```
LayneHaber, line 218:
```suggestion
    _withdrawAdminFees(_key, msg.sender);

    delete s.swapStorages[_key];

    emit SwapRemoved(_key, msg.sender);
  }

  /**
  * @notice Withdraws all admin fees for pool at key to provided address and emits event
  * @param _key Hash of the canonical domain and id
  * @param _to Recipient of fees
  */
  function _withdrawAdminFees(bytes32 _key, address _to) internal {
     s.swapStorages[_key].withdrawAdminFees(_to);
     emit AdminFeesWithdrawn(_key, _to);
  }
```

LayneHaber: use same internal function for `withdrawSwapAdminFees`

LayneHaber (changes requested):
still curious about just setting the flag instead of clearing all of the storage, and allowing re-initializing

packages/deployments/contracts/contracts/core/connext/facets/SwapAdminFacet.sol
gpersoon: That seems like a solid approach.
When the swap is disabled it can't be used of course, i presume the system should use an external amm then.

0xbok: we can add a functionality so that Connext Diamond can still let liquidity providers to remove their balance from a disabled AMM.

0xbok: what @gpersoon mentioned can make it work. Always use an external AMM. The liquidity providers deposit tokens directly to that AMM. Now Connext Diamond can just connect and disconnect from it without worrying about the token balance in AMM.

packages/deployments/contracts/contracts/core/connext/facets/SwapAdminFacet.sol
```diff
@@ -177,13 +186,51 @@ contract SwapAdminFacet is BaseConnextFacet {
...
+      delete s.tokenIndexes[_key][address(s.swapStorages[_key].pooledTokens[i])];
+
+      unchecked {
+        ++i;
+      }
+    }
+
+    _withdrawAdminFees(_key, msg.sender);
+
+    delete s.swapStorages[_key];
```
LayneHaber, line 213: if you delete this you will no longer be able to remove liquidity, even after the pool is disabled, see comments above for best approach

LayneHaber (changes requested):

gpersoon:
@LayneHaber @liu-zhipeng 
The current change is:
- disable the swap (with a wait period then LP can retrieve their tokens)
- after the wait period : send the remaining tokens to the owner and remove the swap

One improvement suggestion: during the wait period the swap is still selected but it also immediately reverts.
So a solution would be: on the (4) locations where  `if (ipool.exists())...` is called, also check the pool isn't disabled.
Or maybe better integrate this in the `exists()` function.

liu-zhipeng:
> exists

Good idea! fixed here ^^

PR #2385, opened 2022-11-07
Spearbit-i29: OptiHubConnector, verify sender in _processMessage

jakekidd:
## Description
Fixes https://github.com/spearbit-audits/connext-nxtp/issues/29

## Type of change

- [ ] Refactoring

### DIFF
--- a/packages/deployments/contracts/contracts/messaging/connectors/optimism/OptimismHubConnector.sol
+++ b/packages/deployments/contracts/contracts/messaging/connectors/optimism/OptimismHubConnector.sol
@@ -67,6 +67,8 @@ contract OptimismHubConnector is HubConnector, BaseOptimism {
    * spoke)
    */
   function _processMessage(bytes memory _data) internal override {
+    require(_verifySender(mirrorConnector), "!mirrorConnector");
+
     // sanity check root length
     require(_data.length == 32, "!length");
 


packages/deployments/contracts/contracts/messaging/connectors/optimism/OptimismHubConnector.sol
```diff
@@ -67,6 +67,8 @@ contract OptimismHubConnector is HubConnector, BaseOptimism {
    * spoke)
    */
   function _processMessage(bytes memory _data) internal override {
+    require(_verifySender(mirrorConnector), "!mirrorConnector");
```
LayneHaber, line 70: right now, this will cause a revert if it is not called by the AMB.

The `BaseOptimism` defines `_verifySender` as:

```js
function _verifySender(address _amb, address _expected) internal view returns (bool) {
    require(msg.sender == _amb, "!bridge");
    return OptimismAmb(_amb).xDomainMessageSender() == _expected;
}
```

which is correct for when the message is processed after the 7 day delay from the optimism bridge interface. however, it will fail when `_processMessage` is called by `processMessageFromRoot`, where the `msg.sender` is a relayer. the mirror connector is still asserted in `processMessageFromRoot`, but we can't call `_verifySender` within that flow

LayneHaber (changes requested):

LayneHaber:
also tests

LayneHaber:
see: https://github.com/spearbit-audits/connext-nxtp/issues/29#issuecomment-1310734368

jakekidd:
Issue should be dismissed as no longer valid!

See https://github.com/spearbit-audits/connext-nxtp/issues/29#issuecomment-1317300040

PR #2387, opened 2022-11-07
Spearbit-i137: make set fx root/child tunnel method internal

jakekidd:
## Description
fixes https://github.com/spearbit-audits/connext-nxtp/issues/137

## Type of change

- [ ] Refactoring


sanchaymittal (approved):
looking good!

jakekidd:
> ideally, we would compute the address in advance and provide it on the constructor, then assert it is `address(0)`. however, the `mirrorConnector` on mainnet (or any hub) would always be empty.

agreed, this seemed like the best compromise

LayneHaber (approved):

PR #2388, opened 2022-11-08
Spearbit 105

liu-zhipeng:
## Description
Fix https://github.com/spearbit-audits/connext-nxtp/issues/105

## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes 
https://github.com/spearbit-audits/connext-nxtp/issues/105

### DIFF
--- a/packages/deployments/contracts/contracts/core/connext/libraries/SwapUtils.sol
+++ b/packages/deployments/contracts/contracts/core/connext/libraries/SwapUtils.sol
@@ -643,6 +643,7 @@ library SwapUtils {
     uint256[] memory multipliers = self.tokenPrecisionMultipliers;
 
     uint256 numBalances = balances.length;
+    require(amounts.length != numBalances, "invalid length of amounts");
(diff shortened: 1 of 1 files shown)


packages/deployments/contracts/contracts/core/connext/libraries/SwapUtils.sol
```diff
@@ -643,6 +643,7 @@ library SwapUtils {
     uint256[] memory multipliers = self.tokenPrecisionMultipliers;
 
     uint256 numBalances = balances.length;
+    require(amounts.length != numBalances, "invalid length of amounts");
```
gpersoon, line 646: I think it should be:
```solidity
require(amounts.length == numBalances, "invalid length of amounts");
```

Saw-mon-and-Natalie, line 646: length check can be moved to the very to of the function.

LayneHaber (changes requested):
needs test

LayneHaber (approved):

PR #2390, opened 2022-11-08
Spearbit i37: add removal vector for root manager queue

jakekidd:
## Description
Fixes https://github.com/spearbit-audits/connext-nxtp/issues/37

## Type of change

- [x] Refactoring


jakekidd:
included changes from `rate-limit-messaging` due to dependent changes / to avoid conflicts

packages/deployments/contracts/contracts/messaging/RootManager.sol
```diff
@@ -25,16 +26,24 @@ contract RootManager is ProposedOwnable, IRootManager, WatcherClient, DomainInde
 
   event DelayBlocksUpdated(uint256 previous, uint256 updated);
 
-  event RootAggregated(uint32 domain, bytes32 receivedRoot, uint256 queueIndex);
+  event RootReceived(uint32 domain, bytes32 receivedRoot, uint256 queueIndex);
```
LayneHaber, line 29: we should make an issue flagging these event changes

LayneHaber (commented):
this pr should be separated as it seems like it may be tacking a few different issues, not just the one referenced (no way of removing fraudulent roots)

packages/deployments/contracts/contracts/messaging/RootManager.sol
jakekidd: agreed

jakekidd:
> this pr should be separated as it seems like it may be tacking a few different issues, not just the one referenced (no way of removing fraudulent roots)

We can separate them by merging `rate-limit-messaging`. There's dependent changes, so this PR had to be based on that branch.

packages/deployments/contracts/contracts/messaging/libraries/Queue.sol
```diff
@@ -106,8 +109,15 @@ library QueueLib {
     uint256 index; // Cursor for index in the batch of `items`.
     // NOTE: `first <= last` rephrased here to `!(first > last)` as it's a cheaper condition.
     while (!(first > last)) {
-      items[index] = queue.data[first];
+      bytes32 item = queue.data[first];
+      // Check to see if the item has been removed before appending it to the array.
+      if (!queue.removed[item]) {
+        items[index] = item;
```
gpersoon, line 115: This leaves empty spaces in `items[]`. (note the 0 values could be tricky)
Because `index` is increased even when `removed[]` is `true`

packages/deployments/contracts/contracts/messaging/libraries/Queue.sol
```diff
@@ -106,8 +109,15 @@ library QueueLib {
...
+      }
```
csanuragjain, line 116: If item is removed then items[index] will not be updated. But in this case also index will get updated `++index; at L125`

This raises 2 problems:

1. This means returned items will have 0 address roots in middle of the indexes (if some root were blacklisted)
2. Since items size was defined beforehand `last + 1 - first`, it will always contain 0 address at end indexes if any of the root was blacklisted

packages/deployments/contracts/contracts/messaging/libraries/Queue.sol
```diff
@@ -121,6 +131,14 @@ library QueueLib {
     return items;
   }
 
+  /**
+   * @notice Sets a certain value to be ignored (skipped) when dequeuing.
+   */
+  function remove(Queue storage queue, bytes32 item) internal {
+    require(!queue.removed[item], "already removed");
+    queue.removed[item] = true;
+  }
```
csanuragjain, line 140: Seems like fix for RootManager.sol who will call this function is missing in this pull request. Kindly check

packages/deployments/contracts/contracts/messaging/libraries/Queue.sol
jakekidd: hmm. Open to any suggestions here, cc @csanuragjain since your comment seems to be saying the same thing.

I definitely don't want to leave gaps, but the idea of copying over into another array in memory does not sound cost effective

packages/deployments/contracts/contracts/messaging/libraries/Queue.sol
jakekidd: We could skip over an index in this event, and essentially move all the invalid items to the end of the array. We could track the count of items that were invalid, and then copy over into a new array the `<length of the old array> - <invalid item count>`, but that's not a very clean operation

packages/deployments/contracts/contracts/messaging/libraries/Queue.sol
jakekidd: Alternatively we could just *not* increment index if `removed[]` is `true` and instead return an array with some potential subset of 0-values at the end of it, and document that in the natspec here. We could then handle this particular arrangement wherever the array ends up being consumed.

Otherwise, there might be a hacky way to truncate the array in assembly of its 0-values...

jakekidd: Actually, I might just elect for simpler solution of copying over to a second return-value array IFF there was a removed item. Basically, track whether an item was removed in this loop, then recopy accordingly. That way if we're going gas-inefficient route, at least it will only be inefficient in edge case

packages/deployments/contracts/contracts/messaging/libraries/Queue.sol
```diff
@@ -104,21 +107,60 @@ library QueueLib {
...
-        ++index;
         ++first;
       }
     }
+
     // Update the value for `first` in our queue object since we've dequeued a number of elements.
     queue.first = first;
-    return items;
+
+    if (removedCount == 0) {
```
jakekidd, line 141: cc @gpersoon @csanuragjain please check the solution here.

packages/deployments/contracts/contracts/messaging/RootManager.sol
```diff
@@ -141,6 +143,20 @@ contract RootManager is ProposedOwnable, IRootManager, WatcherClient, DomainInde
...
+  /**
+   * @notice Removes (effectively blacklists) a given (fraudulent) root from the queue of pending
+   * inbound roots.
+   * @dev The given root does NOT have to currently be in the queue. It isn't removed from the queue
+   * directly, but instead is filtered out when dequeuing is done for the sake of aggregation.
+   * @dev Can only be called by the owner when the protocol is paused.
+   *
+   * @param _root The root to be discarded.
+   */
+  function discardRoot(bytes32 _root) public onlyOwner whenPaused {
```
jakekidd, line 155: @csanuragjain See method here added as the interface to discard (i.e. remove / ban) a given root.

Note that I'm not adding a method for "undiscarding" a given root! That root will be banned indefinitely. That seems perfectly acceptable to me, but lmk if you have any concerns.

cc @LayneHaber

packages/deployments/contracts/contracts/messaging/libraries/Queue.sol
```diff
@@ -104,21 +107,60 @@ library QueueLib {
...
+      return items;
+    } else {
+      // If some items were removed, there will be a number of trailing 0 values we need to truncate
+      // from the array. Create a new array with all of the items up until these empty values.
+      uint256 amendedLength = last + 1 - first - removedCount;
```
csanuragjain, line 146: This will fail since first has now incremented to last+1 due to earlier while loop.
I think we could use amendedLength=index since index contains the actual number of items being inserted

packages/deployments/contracts/contracts/messaging/libraries/Queue.sol
csanuragjain: The solution wont work as highlighted at comment https://github.com/connext/nxtp/pull/2390/files#r1020708564

packages/deployments/contracts/contracts/messaging/RootManager.sol
csanuragjain: Only thing is you need to be really careful while removing such that genuine roots are never removed by mistake. Otherwise I think it is fine

LayneHaber: Yeah this is a failure-recovery scenario, so we can ensure the correct roots are banned

packages/deployments/contracts/contracts/messaging/libraries/Queue.sol
jakekidd: Fixed!

packages/deployments/contracts/contracts_forge/messaging/libraries/Queue.t.sol
```diff
@@ -138,4 +138,33 @@ contract QueueLibTest is ForgeHelper {
...
+    uint256 testEntryCount = 100;
+    Entry[] memory entries = utils_generateEntries(testEntryCount, 1);
+    for (uint256 i; i < testEntryCount; i++) {
+      Entry memory entry = entries[i];
+      utils_enqueueEntry(entry);
+    }
+
+    // Remove one of the entries.
+    uint256 badRootIndex = 54;
+    queue.remove(entries[badRootIndex].message);
```
LayneHaber, line 153: event emission

packages/deployments/contracts/contracts_forge/messaging/libraries/Queue.t.sol
```diff
@@ -138,4 +138,33 @@ contract QueueLibTest is ForgeHelper {
...
+
+    // Delay blocks will be 10 below. Let's roll ahead 20 blocks and we can be sure that
+    // normally, the whole queue would be verified.
+    vm.roll(block.number + 20);
+    // If one of the roots was removed we should get 99 roots back.
+    bytes32[] memory verified = queue.dequeueVerified(10, 10000);
+    assertEq(verified.length, 99);
```
LayneHaber, line 160: should add an additional test case with partial root removal

LayneHaber (approved):
some nits but lgtm

PR #2407, opened 2022-11-08
Spearbit 17, 20

rhlsthrm:
## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes


packages/deployments/contracts/contracts/messaging/RootManager.sol
```diff
@@ -144,7 +144,11 @@ contract RootManager is ProposedOwnable, IRootManager, WatcherClient, DomainInde
    * @param _connectors Array of connectors: should match exactly the array of `connectors` in storage
    * (see `_domains` param's info on reducing gas costs).
    */
-  function propagate(uint32[] calldata _domains, address[] calldata _connectors) external whenNotPaused {
+  function propagate(
+    uint32[] calldata _domains,
+    address[] calldata _connectors,
+    uint256[] calldata _fees
+  ) external whenNotPaused {
```
sanchaymittal, line 151: needs to be payable i think

packages/deployments/contracts/contracts/messaging/RootManager.sol
```diff
@@ -163,7 +170,7 @@ contract RootManager is ProposedOwnable, IRootManager, WatcherClient, DomainInde
     (bytes32 _aggregateRoot, uint256 _count) = MERKLE.insert(_verifiedInboundRoots);
 
     for (uint32 i; i < _numDomains; ) {
-      IHubConnector(_connectors[i]).sendMessage(abi.encodePacked(_aggregateRoot));
+      IHubConnector(_connectors[i]).sendMessage{value: _fees[i]}(abi.encodePacked(_aggregateRoot), _encodedData[i]);
```
jakekidd, line 173: the sum value of all fees must add up to what this contract received in ETH when `propagate` was called, otherwise 1 of 2 things could happen:
- ETH could get left behind in the contract (w/ no dusting methods here)
- someone could call `propagate` and send fees to the wrong connectors as a griefing vector

jakekidd: > someone could call propagate and send fees to the wrong connectors as a griefing vector

Also, this problem kind of arises here in general. With every connector, a different fee requirement will exist. Without a way to enforce that fee requirement on-chain, and with this method being external w/o accessibility restrictions, someone could grief the messaging layer by calling propagate with 0 fees provided when fees are in fact needed.

We need to either restrict this method to relayers for now OR pre-configure fees ahead of time for each connector (e.g. each connector expects a certain fee amount in the `sendMessage` call, and the owner configures that amount separately)

jakekidd (changes requested):
see comment regarding griefing vector

packages/deployments/contracts/contracts/messaging/RootManager.sol
LayneHaber: fixed by decrementing the sum, should error on underflow

sanchaymittal (approved):

packages/deployments/contracts/contracts/messaging/connectors/arbitrum/ArbitrumHubConnector.sol
```diff
@@ -47,20 +97,47 @@ contract ArbitrumHubConnector is HubConnector {
...
+    address _outbox,
+    uint256 _maxSubmissionCostCap,
+    uint256 _maxGasCap,
+    uint256 _gasPriceCap
+  ) HubConnector(_domain, _mirrorDomain, _amb, _rootManager, _mirrorConnector) {
     outbox = IArbitrumOutbox(_outbox);
     rollup = IArbitrumRollup(outbox.rollup());
+
+    // Set initial caps for L1 -> L2 messages
+    maxSubmissionCostCap = _maxSubmissionCostCap;
```
rhlsthrm, line 109: Is there a reason for these caps in the contract? Whoever calls the function will be funding with their own gas so I don't see the need to enforce a cap here.

LayneHaber: it depends on how the relayer fees are being repaid -- i think best practice for these things is to allow runtime info to be passed in up to a point, but theres always a risk relayer could bid up to the cap constantly.

LayneHaber (approved):

PR #2413, opened 2022-11-09
Spearbit 114

LayneHaber:
## Description

Fixes https://github.com/spearbit-audits/connext-nxtp/issues/114

## Type of change

- [x] Refactoring
- [x] Breaking change (fix or feature that would cause existing functionality to change)

- Changes the storage structure of the router configuration to:
```js
/**
 * @notice Contains configs for each router
 * @param approved Whether the router is whitelisted, settable by admin
 * @param portalApproved Whether the router is whitelisted for portals, settable by admin
 * @param routerOwners The address that can update the `recipient`
 * @param proposedRouterOwners Owner candidates
 * @param proposedRouterTimestamp When owner candidate was proposed (there is a delay to acceptance)
 */
struct RouterConfig {
  bool approved;
  bool portalApproved;
  address owner;
  address recipient;
  address proposed;
  uint256 proposedTimestamp;
}

// ... in AppStorage ... 

mapping(address => RouterConfig) routerConfigs;
```

- Adds an `initializeRouter` function that sets the `owner` and `recipient`. This function is callable by anyone, and uses the `msg.sender` as the router
- Removes `setupRouter`
- Adds an `addRouter` function that only updates approval of a given router, not config
- Updates `removeRouter` to only update approval for a given router
- Ensures renouncing router ownership means `recipient` cannot be set
- Simplifies authentication on router config

## Related Issue(s)

Fixes https://github.com/spearbit-audits/connext-nxtp/issues/114


packages/deployments/contracts/contracts/core/connext/facets/RoutersFacet.sol
```diff
@@ -249,86 +237,47 @@ contract RoutersFacet is BaseConnextFacet {
...
-    address recipient
-  ) external onlyOwnerOrRouter {
+  function addRouter(address _router) external onlyOwnerOrRouter {
     // Sanity check: not empty
-    if (router == address(0)) revert RoutersFacet__setupRouter_routerEmpty();
+    if (_router == address(0)) revert RoutersFacet__addRouter_routerEmpty();
 
     // Sanity check: needs approval
-    if (s.routerPermissionInfo.approvedRouters[router]) revert RoutersFacet__setupRouter_alreadyAdded();
+    if (s.routerConfigs[_router].approved) revert RoutersFacet__addRouter_alreadyAdded();
```
gpersoon, line 254: Perhaps check the router is initialized?
e.g. `s.routerConfigs[_router].owner != address(0)`

liu-zhipeng (approved):

packages/deployments/contracts/contracts/core/connext/facets/RoutersFacet.sol
LayneHaber: i thought about it, but dont see a reason that admins cant preapprove a router. if the EOA was added incorrectly, that's a concern regardless (i.e. typos).

jakekidd: agreed. approval should we be a separate switch to configuration, which should be done at the EOA owner's discretion. i.e. those steps should be basically async

packages/deployments/contracts/contracts/core/connext/facets/RoutersFacet.sol
```diff
@@ -249,86 +237,47 @@ contract RoutersFacet is BaseConnextFacet {
...
-   * @param recipient Initial Recipient of router
+   * @notice Used to whitelist a given router
+   * @param _router Router address to setup
    */
-  function setupRouter(
-    address router,
-    address owner,
-    address recipient
-  ) external onlyOwnerOrRouter {
+  function addRouter(address _router) external onlyOwnerOrRouter {
```
jakekidd, line 249: maybe we should change method names to `approveRouter` and `revokeRouter` but idk, open to either. Just important to disambiguate the two things as much as possible

jakekidd: or `unapproveRouter` - we should follow the pattern for what we have for portals `approveRouterForPortals`, `unapproveRouterForPortals`...

packages/deployments/contracts/contracts/core/connext/facets/RoutersFacet.sol
```diff
@@ -249,86 +237,47 @@ contract RoutersFacet is BaseConnextFacet {
...
+    emit RouterAdded(_router, msg.sender);
   }
 
   /**
    * @notice Used to remove routers that can transact crosschain
-   * @param router Router address to remove
+   * @param _router Router address to remove
    */
-  function removeRouter(address router) external onlyOwnerOrRouter {
+  function removeRouter(address _router) external onlyOwnerOrRouter {
```
jakekidd, line 267: see note to change name to unapproveRouter

packages/deployments/contracts/contracts/core/connext/facets/RoutersFacet.sol
```diff
@@ -501,9 +477,10 @@ contract RoutersFacet is BaseConnextFacet {
     address payable _to,
     address _router
   ) external nonReentrant whenNotPaused {
-    // Caller must be the router owner
-    if (msg.sender != getRouterOwner(_router)) revert RoutersFacet__removeRouterLiquidityFor_notOwner();
-
+    // Caller must be the router owner, if defined, else the router
```
jakekidd, line 480: this is misleading: judging by the `initializeRouter` method, the owner will always be defined as the `msg.sender` in that fn or w/e was passed in.

If `owner == address(0)` here then the router is uninitialized: I think we could revert here in that case?

packages/deployments/contracts/contracts/core/connext/facets/RoutersFacet.sol
```diff
@@ -525,6 +502,48 @@ contract RoutersFacet is BaseConnextFacet {
...
+    // Check owner is changing
+    if (_previous == _updated) revert RoutersFacet__setRouterOwner_noChange();
+
+    // Set new owner
+    s.routerConfigs[_router].owner = _updated;
+
+    // Emit event
+    emit RouterOwnerAccepted(_router, _previous, _updated);
+  }
+
```
jakekidd, line 546: Need to sanity check that the router has approval in `_addLiquidityForRouter`! If the router is unapproved and liquidity is added, then that will stack against our liq caps w/o actually offering up liquidity.

packages/deployments/contracts/contracts/core/connext/facets/RoutersFacet.sol
```diff
@@ -501,9 +477,10 @@ contract RoutersFacet is BaseConnextFacet {
     address payable _to,
     address _router
   ) external nonReentrant whenNotPaused {
-    // Caller must be the router owner
-    if (msg.sender != getRouterOwner(_router)) revert RoutersFacet__removeRouterLiquidityFor_notOwner();
-
+    // Caller must be the router owner, if defined, else the router
+    address owner = s.routerConfigs[_router].owner;
+    address permissioned = owner == address(0) ? _router : owner;
```
jakekidd, line 482: wb `recipient` ? are we not using that?

packages/deployments/contracts/contracts/core/connext/facets/RoutersFacet.sol
gpersoon: The `owner` can be set to `0` via `proposeRouterOwner()` and `acceptProposedRouterOwner()`.
(not sure if this is desired behavior)

LayneHaber: that is desired behavior, and you should still *at least* be able to withdraw liquidity to the recipient (which should revert if owner is 0) using the EOA

the hard requirement being the ability to change the recipient can be removed

packages/deployments/contracts/contracts/core/connext/facets/RoutersFacet.sol
LayneHaber: that check exists (L550-552):

```js
// Sanity check: router is approved.
if (!_isRouterWhitelistRemoved() && !getRouterApproval(_router))
    revert RoutersFacet__addLiquidityForRouter_badRouter();
```

packages/deployments/contracts/contracts/core/connext/facets/RoutersFacet.sol
```diff
@@ -395,67 +342,96 @@ contract RoutersFacet is BaseConnextFacet {
...
 
     // Check timestamp has passed
-    if (block.timestamp - s.routerPermissionInfo.proposedRouterTimestamp[router] <= _delay)
+    if (block.timestamp - config.proposedTimestamp <= _delay)
       revert RoutersFacet__acceptProposedRouterOwner_notElapsed();
 
-    // Get current owner + proposed
-    address _proposed = s.routerPermissionInfo.proposedRouterOwners[router];
+    // Check the caller
+    address expected = config.proposed == address(0) ? config.owner : config.proposed;
```
jakekidd, line 388: So it seems that setting the owner to address(0) will effectively prevent the router owner from changing the recipient again in the future. If someone does this by mistake, and they have the recipient set to an undesirable address (e.g an EOA with a typo) they can potentially brick their router, as they will be unable to withdraw their funds to an EOA they control.

This doesn't need to get fixed, but I'm just emphasizing it here as one of the pitfalls of running a router worth documenting. We need to be sure router operators understand the implications of setting owner to address(0).

jakekidd (approved):

packages/deployments/contracts/contracts/core/connext/facets/RoutersFacet.sol
LayneHaber: yeah, added a comment in propose for this

packages/deployments/contracts/contracts/core/connext/facets/RoutersFacet.sol
LayneHaber: used same convention as portal

PR #2432, opened 2022-11-10
Spearbit 32

rhlsthrm:
## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes


jakekidd (approved):

packages/deployments/contracts/contracts_forge/messaging/WatcherClient.t.sol
```diff
@@ -0,0 +1,40 @@
...
+    vm.prank(owner);
+    watcherClient = new WatcherClient(watcherManager);
+    vm.mockCall(address(0), abi.encodeWithSelector(WatcherManager.isWatcher.selector, watcher), abi.encode(true));
+  }
+
+  function test_WatcherClient__setWatcherManager_failsIfNotOwner(address caller) public {
+    if (caller == owner) {
+      // fuzz test, return if owner
+      return;
+    }
```
LayneHaber, line 27:
```suggestion
    vm.assume(owner != caller);
```

packages/deployments/contracts/contracts_forge/messaging/WatcherClient.t.sol
```diff
@@ -0,0 +1,40 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity 0.8.15;
```
LayneHaber, line 2:
```suggestion
pragma solidity 0.8.17;
```

packages/deployments/contracts/contracts/messaging/WatcherClient.sol
```diff
@@ -16,10 +16,11 @@ contract WatcherClient is ProposedOwnable, Pausable {
   event WatcherManagerChanged(address watcherManager);
 
   // ============ Properties ============
-  WatcherManager watcherManager;
+  WatcherManager public watcherManager;
 
   // ============ Constructor ============
-  constructor(address _watcherManager) {
+  constructor(address _watcherManager) ProposedOwnable() {
+    _setOwner(msg.sender);
```
LayneHaber, line 23: should be set from the topmost class, imo. only risk is duplicate events on constructors, so not a huge risk

LayneHaber (changes requested):

packages/deployments/contracts/contracts/messaging/WatcherClient.sol
LayneHaber: see: https://github.com/spearbit-audits/connext-nxtp/issues/32#issuecomment-1311770726

gpersoon: Setting at the topmost class sounds good, that way you only have 1x `_setOwner()`.
So that means it should be removed here.

rhlsthrm: Fixed.

rhlsthrm:
Fixed and updated PR.

PR #2434, opened 2022-11-10
2433 fix spearbit pr 57

rhlsthrm:
## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes


packages/deployments/contracts/contracts/core/connext/helpers/StableSwap.sol
```diff
@@ -83,7 +83,8 @@ contract StableSwap is IStableSwap, OwnerPausableUpgradeable, ReentrancyGuardUpg
 
     uint256[] memory precisionMultipliers = new uint256[](decimals.length);
 
-    for (uint256 i = 0; i < _pooledTokens.length; i++) {
+    uint256 len = _pooledTokens.length;
+    for (uint256 i = 0; i < len; i++) {
```
jakekidd, line 87: idk if we're conforming with this across this whole file, but we should be using the ideal for loop conditional pattern:
```
for (uint256 i; i < len; ) {
  ...
  unchecked {
    ++i;
  }
}
```

jakekidd: This comment applies to multiple `for` loops in the section, but isn't really the responsibility of this specific PR. So I'm going to label this as a nit (feel free to fix all instances) and approve

jakekidd (approved):
looks good. got that one nit above ^ which is optional

packages/deployments/contracts/contracts/core/connext/libraries/LibDiamond.sol
```diff
@@ -112,7 +112,8 @@ library LibDiamond {
       uint256 time = ds.acceptanceTimes[keccak256(abi.encode(_diamondCut, _init, _calldata))];
       require(time != 0 && time <= block.timestamp, "LibDiamond: delay not elapsed");
     } // Otherwise, this is the first instance of deployment and it can be set automatically
-    for (uint256 facetIndex; facetIndex < _diamondCut.length; facetIndex++) {
+    uint256 len = _diamondCut.length;
+    for (uint256 facetIndex; facetIndex < len; facetIndex++) {
```
gpersoon, line 116: There is also a for loop in function `addFunctions()` that can be optimized

packages/deployments/contracts/contracts/core/connext/libraries/LibDiamond.sol
```diff
@@ -155,7 +156,8 @@ library LibDiamond {
     if (selectorPosition == 0) {
       addFacet(ds, _facetAddress);
     }
-    for (uint256 selectorIndex; selectorIndex < _functionSelectors.length; selectorIndex++) {
+    uint256 len = _functionSelectors.length;
```
gpersoon, line 159: If the `len` is calculated as the first line of the function then the following optimization can be done:
```solidity
require(len != 0, "LibDiamondCut: No selectors in facet to cut");
```

LayneHaber (changes requested):
should apply @jakekidd and @gpersoon changes to treat all loops across the files in a uniform manner

packages/deployments/contracts/contracts/core/connext/helpers/StableSwap.sol
rhlsthrm: Fixed this. Changed quite a few loops so this needs another look.

rhlsthrm:
Addressed noted issues.

LayneHaber (approved):

PR #2449, opened 2022-11-11
Spearbit 194

rhlsthrm:
## Description
Fix https://github.com/spearbit-audits/connext-nxtp/issues/194

## Type of change

- [x] Refactoring

WatcherManager has a private mapping watchers and a getter function isWatcher() to query that mapping. Since WatcherManager is not inherited by any other contract, it is safe to make it public to avoid the need of an explicit getter function.

## Related Issue(s)

Fixes

### DIFF
--- a/packages/deployments/contracts/contracts/messaging/WatcherManager.sol
+++ b/packages/deployments/contracts/contracts/messaging/WatcherManager.sol
@@ -15,7 +15,7 @@ contract WatcherManager is ProposedOwnable {
   event WatcherRemoved(address watcher);
 
   // ============ Properties ============
-  mapping(address => bool) private watchers;
+  mapping(address => bool) public isWatcher;
 
   // ============ Constructor ============
   constructor() ProposedOwnable() {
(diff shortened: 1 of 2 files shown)


packages/deployments/contracts/contracts/messaging/WatcherManager.sol
```diff
@@ -29,17 +29,17 @@ contract WatcherManager is ProposedOwnable {
    * @dev Owner can enroll a watcher (abilities are defined by inheriting contracts)
    */
   function addWatcher(address _watcher) external onlyOwner {
-    require(!watchers[_watcher], "already watcher");
-    watchers[_watcher] = true;
+    require(!isWatcher[_watcher], "already watcher");
+    isWatcher[_watcher] = true;
     emit WatcherAdded(_watcher);
```
liu-zhipeng, line 34: We don't need to check `address(0)` when adding new watcher?

LayneHaber: we could, but it would be impossible to make that the `msg.sender`. 

should still add the check

PR #2494, opened 2022-11-15
Spearbit 135

LayneHaber:
## Description

Fixes https://github.com/spearbit-audits/connext-nxtp/issues/135

- Uses modifiers to track deadlines on ownership

## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes


packages/deployments/contracts/contracts/core/connext/facets/ProposedOwnableFacet.sol
```diff
@@ -67,6 +67,23 @@ contract ProposedOwnableFacet is BaseConnextFacet, IProposedOwnable {
...
+    if ((block.timestamp - s._proposedOwnershipTimestamp) <= delay())
+      revert ProposedOwnableFacet__ownershipDelayElapsed_delayNotElapsed();
+    _;
+  }
+
+  modifier delayElapsed(uint256 start) {
+    // Ensure delay has elapsed
+    if ((block.timestamp - start) <= delay()) revert ProposedOwnableFacet__delayElapsed_delayNotElapsed();
+    _;
+  }
```
Saw-mon-and-Natalie, line 85: If the custom error type is not important in case of a not elasped delay, these `2` modifiers can be combined into one (we can just use `delayElapsed(uint256)`).

PR #2497, opened 2022-11-16
Spearbit 86

LayneHaber:
## Description

Fixes https://github.com/spearbit-audits/connext-nxtp/issues/86

## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes


packages/deployments/contracts/contracts/core/connext/facets/BaseConnextFacet.sol
```diff
@@ -4,14 +4,14 @@ pragma solidity 0.8.17;
 import {TransferInfo, AppStorage, TokenId, Role} from "../libraries/LibConnextStorage.sol";
 import {LibDiamond} from "../libraries/LibDiamond.sol";
 import {AssetLogic} from "../libraries/AssetLogic.sol";
+import {Constants} from "../libraries/Constants.sol";
```
Saw-mon-and-Natalie, line 7: `Constants` not used in directly this file.

liu-zhipeng (approved):

packages/deployments/contracts/contracts/core/connext/facets/BridgeFacet.sol
```diff
@@ -859,7 +860,7 @@ contract BridgeFacet is BaseConnextFacet {
         _params.to,
         gasleft() - 10_000,
```
Saw-mon-and-Natalie, line 861: Perhaps, we can define a constant for `10_000`.

packages/deployments/contracts/contracts/core/connext/facets/StableSwapFacet.sol
```diff
@@ -4,6 +4,7 @@ pragma solidity 0.8.17;
 import {IERC20} from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
 
 import {AmplificationUtils, SwapUtils} from "../libraries/AmplificationUtils.sol";
+import {Constants} from "../libraries/Constants.sol";
```
Saw-mon-and-Natalie, line 7: Not used directly in this contract.

packages/deployments/contracts/contracts/core/connext/helpers/StableSwap.sol
```diff
@@ -77,8 +78,8 @@ contract StableSwap is IStableSwap, OwnerPausableUpgradeable, ReentrancyGuardUpg
     __ReentrancyGuard_init();
 
     // Check _pooledTokens and precisions parameter
-    require(_pooledTokens.length > 1, "_pooledTokens.length <= 1");
-    require(_pooledTokens.length <= 32, "_pooledTokens.length > 32");
+    require(_pooledTokens.length > Constants.MINIMUM_POOLED_TOKENS - 1, "_pooledTokens.length <= 1");
+    require(_pooledTokens.length < Constants.MAXIMUM_POOLED_TOKENS + 1, "_pooledTokens.length > 32");
```
gpersoon, line 82: the error message specify exact amounts, which might be different than the constants.

packages/deployments/contracts/contracts/core/connext/libraries/Constants.sol
```diff
@@ -0,0 +1,121 @@
...
+
+  // ============= Swaps
+
+  /**
+   * @notice the precision all pools tokens will be converted to
+   * @dev stored here to keep easily in sync between `SwapUtils` and `SwapUtilsExternal`
+   */
+  uint256 public constant MINIMUM_POOLED_TOKENS = 2;
+  uint256 public constant MAXIMUM_POOLED_TOKENS = 32;
+
```
liu-zhipeng, line 76: We decided max pooled tokens should be 16

LayneHaber: it wasnt merged in yet so i didnt have the constants updated, but will fix

packages/deployments/contracts/contracts/core/connext/libraries/AssetLogic.sol
```diff
@@ -307,8 +316,7 @@ library AssetLogic {
       IStableSwap pool = s.adoptedToLocalExternalPools[_key];
       address poolAddress = address(pool);
 
-      // NOTE: We revert here if the external stableswap pool doesn't exist.
-      if (poolAddress == address(0)) revert AssetLogic__swapAsset_externalStableSwapPoolDoesNotExist();
+      // NOTE: This call will revert if the external stableswap pool doesn't exist.
```
csanuragjain, line 320: The zero address check is missing. Kindly correct

gpersoon: Calling non existing contracts is always tricky, but is solved by solidity: [solidity docs]:(https://docs.soliditylang.org/en/v0.8.17/control-structures.html#external-function-calls)
`
Due to the fact that the EVM considers a call to a non-existing contract to always succeed, Solidity uses the extcodesize opcode to check that the contract that is about to be called actually exists (it contains code) and causes an exception if it does not. This check is skipped if the return data will be decoded after the call and thus the ABI decoder will catch the case of a non-existing contract.`
Thus the 0 check isn't needed

PR #2502, opened 2022-11-16
Spearbit 106

LayneHaber:
## Description

Fixes https://github.com/spearbit-audits/connext-nxtp/issues/106

- Removes unused code and errors

## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes

### DIFF
--- a/packages/deployments/contracts/contracts/shared/libraries/TypedMemView.sol
+++ b/packages/deployments/contracts/contracts/shared/libraries/TypedMemView.sol
@@ -393,16 +393,6 @@ library TypedMemView {
   }
 
   /**
-   * @notice          Optimized type comparison. Checks that the 5-byte type flag is equal.
-   * @param left      The first view
-   * @param right     The second view
-   * @return          bool - True if the 5-byte type flag is equal
-   */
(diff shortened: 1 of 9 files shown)


packages/deployments/contracts/contracts/shared/libraries/TypedMemView.sol
```diff
@@ -392,16 +392,6 @@ library TypedMemView {
...
-  /**
-   * @notice          Optimized type comparison. Checks that the 5-byte type flag is equal.
-   * @param left      The first view
-   * @param right     The second view
-   * @return          bool - True if the 5-byte type flag is equal
-   */
-  function sameType(bytes29 left, bytes29 right) internal pure returns (bool) {
-    return (left ^ right) >> TWENTY_SEVEN_BYTES == 0;
-  }
-
```
liu-zhipeng, line 404: We will not use this function in the future?

LayneHaber: if we do use it in the future, we can always return it from the source.

for now, we are not using it so may as well drop the function

PR #2530, opened 2022-11-17
Spearbit 185

LayneHaber:
## Description

Fixes: https://github.com/spearbit-audits/connext-nxtp/issues/185

- Tracks `decimals` in storage
- Updates token storage to use a `TokenConfig` struct

NOTE: Tests still failing

## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes


packages/deployments/contracts/contracts/core/connext/libraries/AssetLogic.sol
```diff
@@ -159,30 +171,30 @@ library AssetLogic {
...
         _key,
         _asset,
-        adopted,
+        config.adopted,
         _amount,
         // NOTE: To get the slippage boundary here, you must take the slippage % off of the
         // normalized amount in (at 18 decimals by convention), then convert that amount
         // to the proper decimals of adopted.
-        calculateSlippageBoundary(uint8(18), IERC20Metadata(adopted).decimals(), _normalizedIn, _slippage)
+        calculateSlippageBoundary(uint8(18), IERC20Metadata(config.adopted).decimals(), _normalizedIn, _slippage)
```
gpersoon, line 197: I guess this could be `config.adoptedDecimals` now

packages/deployments/contracts/contracts/core/connext/facets/TokenFacet.sol
```diff
@@ -280,10 +295,13 @@ contract TokenFacet is BaseConnextFacet {
   // ============ Private Functions ============
 
   function _enrollAdoptedAndLocalAssets(
+    bool _onCanonical,
     address _adopted,
     address _local,
     address _stableSwapPool,
-    TokenId calldata _canonical
+    uint8 _decimals,
```
gpersoon, line 302: `_decimals` doesn't seem to be used

packages/deployments/contracts/contracts/core/connext/facets/TokenFacet.sol
```diff
@@ -297,31 +315,38 @@ contract TokenFacet is BaseConnextFacet {
...
-    if (s.approvedAssets[_key]) revert TokenFacet__addAssetId_alreadyAdded();
-
-    // Update approved assets mapping
-    s.approvedAssets[_key] = true;
+    if (s.tokenConfigs[_key].approval) revert TokenFacet__addAssetId_alreadyAdded();
+
+    // Generate Config
+    // NOTE: using address(0) for stable swap, then using `_addStableSwap`. slightly less
+    // efficient, but preserves events. same case for cap / custody
+    // NOTE: if on canoncail doain, `representation` must *always* be address(0)
```
gpersoon, line 323: typos: `canoncail doain`

packages/deployments/contracts/contracts/core/connext/facets/TokenFacet.sol
```diff
@@ -297,31 +315,38 @@ contract TokenFacet is BaseConnextFacet {
...
       s.representationToCanonical[_local].id = _canonical.id;
-
-      // Update the canonical <> local
-      s.canonicalToRepresentation[_key] = _local;
-      // Add the swap pool if specified
-      if (_stableSwapPool != address(0)) {
-        _addStableSwapPool(_canonical, _stableSwapPool, _key);
-      }
-    } // on the canonical domain, there is no representation (no pool either)
+      // Update cap (only on non-cannonical)
```
gpersoon, line 345: doublecheck: I though `cap` was on canonical domain

packages/deployments/contracts/contracts/core/connext/facets/BridgeFacet.sol
```diff
@@ -512,28 +512,31 @@ contract BridgeFacet is BaseConnextFacet {
       // hash(canonicalId, canonicalDomain), this is safe even when the address(0) asset is not
       // whitelisted.
       bytes32 key;
+      TokenConfig memory config;
```
jakekidd, line 515: this is... slightly problematic. We should probably switch to
```solidity
TokenConfig storage config;
```

So instead we're assigning a reference to a storage location, instead of copying ALL the values from the token config into memory (doing an SLOAD for all the tuple values). That way, we can pull values from storage ONLY as needed

packages/deployments/contracts/contracts/core/connext/libraries/AssetLogic.sol
```diff
@@ -22,12 +22,26 @@ library AssetLogic {
...
 
+  function getConfig(bytes32 _key) internal view returns (TokenConfig storage) {
+    AppStorage storage s = LibConnextStorage.connextStorage();
+    TokenConfig storage config = s.tokenConfigs[_key];
+
+    // Sanity check: not empty
+    // NOTE: adopted decimals will *always* be nonzero (or reflect what is onchain
+    // for the asset). The same is not true for the representation assets, which
+    // will always have 0 decimals on the canonical domain
+    if (config.adoptedDecimals < 1) {
```
jakekidd, line 38: would it be better to just check `config.approval` here?

packages/deployments/contracts/contracts_forge/core/connext/facets/BridgeFacet.t.sol
```diff
@@ -675,8 +675,8 @@ contract BridgeFacetTest is BridgeFacet, FacetHelper {
       }
 
       // if on canonical domain, should decrease
-      if (s.caps[utils_calculateCanonicalHash()] > 0) {
-        assertEq(s.custodied[_local], prevBalances.bridge - routerAmt);
+      if (s.tokenConfigs[utils_calculateCanonicalHash()].cap > 0) {
+        assertEq(s.tokenConfigs[utils_calculateCanonicalHash()].custodied, prevBalances.bridge - routerAmt);
```
jakekidd, line 679: need to double check this is being tested correctly...

packages/deployments/contracts/contracts/core/connext/libraries/AssetLogic.sol
LayneHaber: you may want the config for a non-approved asset

packages/deployments/contracts/contracts/core/connext/facets/BridgeFacet.sol
```diff
@@ -793,6 +786,13 @@ contract BridgeFacet is BaseConnextFacet {
       }
     }
 
+    // If it is the canonical domain, decrease custodied value
+    if (s.domain == _args.params.canonicalDomain && AssetLogic.getConfig(_key).cap > 0) {
+      // NOTE: safe to use the amount here instead of post-swap because there are no
```
jakekidd, line 791: nit: fix comment?

packages/deployments/contracts/contracts/core/connext/facets/InboxFacet.sol
```diff
@@ -235,15 +235,24 @@ contract InboxFacet is BaseConnextFacet {
...
     bytes32 _canonicalId = _tokenId.id();
     uint32 _canonicalDomain = _tokenId.domain();
+
+    // Load amount once.
+    uint256 _amount = _action.amnt();
+
+    // Check for the empty case -- if it is 0 value there is no strict requirement the
+    // canonical information be defined (i.e. you can supply address(0) to xcall). If this
+    // is the case, return _token as address(0)
+    if (_amount == 0 && _canonicalDomain == 0 && _canonicalId == bytes32(0)) {
```
gpersoon, line 245: Shouldn't this have an `emit Receive(...)`?  Because the other 0 case (line 256) also has an emit.

packages/deployments/contracts/contracts/core/connext/libraries/AssetLogic.sol
```diff
@@ -361,7 +373,8 @@ library AssetLogic {
     AppStorage storage s = LibConnextStorage.connextStorage();
 
     // If the adopted asset is the local asset, no need to swap.
-    address adopted = s.canonicalToAdopted[_key];
+    TokenConfig memory config = getConfig(_key);
```
gpersoon, line 376: Different pattern. Here `memory` is used.

Other places use `storage`:
```solidity
TokenConfig storage config = getConfig(_key);`
```

jakekidd: I agree that this should use storage

We might not need every variable from the config, so no reason to SLOAD all of it and copy to mem

packages/deployments/contracts/contracts/core/connext/facets/TokenFacet.sol
```diff
@@ -509,7 +523,7 @@ contract TokenFacet is BaseConnextFacet {
     delete s.tokenConfigs[_key].adoptedToLocalExternalPools;
     delete s.tokenConfigs[_key].approval;
     delete s.tokenConfigs[_key].cap;
-    delete s.tokenConfigs[_key].custodied;
+    // NOTE: custodied will always be 0 at this point
```
gpersoon, line 526: This isn't always true:
Suppose cap & custodied have a value (`!=0`), then you call `_setLiquidityCap()` to set the cap to `0`.
The function doesn't reset the `custodied` value so it will stay at `!=0`

Issue #2627, opened 2022-11-29
Create watcher adapter


rhlsthrm:
Watcher adapter will be an adapter package. The idea is that the adapter package can be used in a standalone watcher agent, or it can be implemented into any workload.

Definition of done:
- [ ] Package exists and is able to be imported into other packages.
- [ ] Function to get total locked amount for a given asset.
- [ ] Function to get total minted amount for a given asset.
- [ ] Wrapper function to validate all assets and make sure that locked >= minted at all times and return info indicating it needs to be paused.

jakekidd:
Opened PR draft https://github.com/connext/nxtp/pull/2638

PR #2638, opened 2022-11-29
Watcher adapter package; method to get total locked, total minted for given asset

jakekidd:
## Description

Fixes https://github.com/connext/nxtp/issues/2627

## Type of change

- [ ] Refactoring

Adds a barebones watcher adapter package with all the functionality needed for watcher agent to monitor and interact with on-chain components


rhlsthrm (commented):
Honestly in this case it might be easier to use a class. Classes would make the implementation cleaner since all the dependencies can be initialized.

packages/deployments/contracts/contracts_forge/core/connext/facets/TokenFacet.t.sol
```diff
@@ -410,7 +410,7 @@ contract TokenFacetTest is TokenFacet, FacetHelper {
 
     vm.prank(_owner);
     // no error message given bc shouldnt be able to find function
-    vm.expectRevert();
+    vm.expectRevert(TokenFacet.TokenFacet__addAssetId_badMint.selector);
```
jakekidd, line 413: @liu-zhipeng you did this right? should we remove above comment, since we're giving error message now?

liu-zhipeng: removed!

packages/adapters/watcher/src/adapter.ts
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  public async pause(_requestContext: RequestContext, reason: string, domains: string[]): Promise<boolean[]> {
+    // TODO: Check to make sure domains are subset of what was provided in VerifierContext in constructor...?
+    return await this.pauser.pause(reason, domains);
+  }
+
+  public async alert(_requestContext: RequestContext, event: ReportEventType, report: Report): Promise<void> {
+    // TODO: Log attempt to alert.
+    await alertViaDiscord(report);
```
rhlsthrm, line 37: Let's try-catch each of these so that we don't break on error. Also please fix this TODO by adding logging.

packages/adapters/watcher/src/alert/discord.ts
```diff
@@ -0,0 +1,78 @@
+import { axiosPost } from "@connext/nxtp-utils";
+
+import { Report } from "../types";
+
+// TODO should be in config
```
rhlsthrm, line 5: Fix todo.

packages/adapters/watcher/src/alert/pagerduty.ts
```diff
@@ -0,0 +1,74 @@
...
+    domains,
+    relevantTransactions,
+    rpcs,
+  });
+
+  if (ROUTING_KEY.length != 32) {
+    throw new Error("PagerDuty Routing Key is invalid");
+  }
+
+  try {
```
rhlsthrm, line 35: This try catch shouldn't be here. We should throw an error in this function and we should catch it at the top level `alert` function.

packages/adapters/watcher/src/alert/pagerduty.ts
```diff
@@ -0,0 +1,74 @@
...
+          {
+            href: "https://connext.network/",
+            text: "Connext Watcher",
+          },
+        ],
+        event_action: "trigger",
+      },
+    });
+  } catch (e: unknown) {
+    console.log(e);
```
rhlsthrm, line 72: Use logger class for errors.

rhlsthrm: But after fixing other issue, this should just throw an error not log it. The catch/logging should be in the calling function.

packages/adapters/watcher/src/pause/pauser.ts
```diff
@@ -0,0 +1,68 @@
...
+          chainId,
+          to: connext.address,
+          data: connextInterface.encodeFunctionData("paused", []),
+        });
+
+        // 2. If not paused, call pause tx
+        if (!paused) {
+          const pauseCalldata = connextInterface.encodeFunctionData("pause", []);
+
+          try {
```
rhlsthrm, line 45: Same as above, I would like to throw custom errors here instead of catching and logging at this level. All the try catching should be at the top level.

rhlsthrm (changes requested):

rhlsthrm:
Actually now that I think about it, the right way for this package to work is for all the errors to be bubbled up and thrown at the top level. The calling package (i.e. agent) should be in charge of handling errors.

packages/adapters/watcher/src/adapter.ts
```diff
@@ -0,0 +1,81 @@
...
+    // attempt to alert via telegram
+    try {
+      await alertViaTelegram(report, config);
+    } catch (e: unknown) {
+      logger.error("alert: failed to alert via telegram", requestContext, methodContext, jsonifyError(e as Error));
+      errors.push(e);
+    }
+
+    if (errors.length) {
+      throw errors;
```
rhlsthrm, line 78: Excellent!

packages/adapters/watcher/src/config.ts
```diff
@@ -0,0 +1,78 @@
+import { existsSync, readFileSync } from "fs";
+
+import { ajv } from "@connext/nxtp-utils";
+import { Static, Type } from "@sinclair/typebox";
+
+export const WatcherConfigSchema = Type.Object({
```
rhlsthrm, line 6: Actually all this code should be in the agent. The adapter should assume this gets passed into it and not deal with any of the actual env stuff. We can add this to the top-level watcher class and then pass it in through the agent config.

liu-zhipeng: Fixed in here
https://github.com/connext/nxtp/pull/2654

liu-zhipeng (approved):

rhlsthrm (approved):

PR #2650, opened 2022-12-02
initial lh refactor for lambda

carlomazzaferro:
## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes


carlomazzaferro:
@rhlsthrm kept the application logic changes to the absolute bare minimum here, a larger refactoring could be in scope but should be left to a separate PR

packages/agents/lighthouse/src/tasks/index.ts
```diff
@@ -0,0 +1,21 @@
+import tracer from "dd-trace";
+
+import { makeProver } from "./prover";
+import { makePropagate } from "./propagate";
+import { makeProcessFromRoot } from "./processFromRoot";
+
+tracer.init();
+
+export const makeLighthouse = async () => {
+  switch (process.env.LIGHTHOUSE_SERVICE) {
```
rhlsthrm, line 10: Would be nicer if we could get this from the config directly instead of requiring env var. That would require pulling up the config to this level instead. I have an idea of how to do this, so I will implement this now.

rhlsthrm (changes requested):

packages/agents/lighthouse/src/tasks/index.ts
carlomazzaferro: I certainly agree. That's how Carto does it in fact, and as much as I wanted to implement the refactor to allow for that, it ended up being larger than I wished.

Looking forward to your impl

rhlsthrm (approved):

PR #2685, opened 2022-12-09
2684 h1

### LINKED ISSUE #2684: [contracts] Macro H1 / Defi Wonderland
**Describe the bug**
See Macro document / Defi Wonderland chat.

LayneHaber:
## Description

Fixes #2684 

## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes #2684


packages/deployments/contracts/contracts/messaging/RootManager.sol
```diff
@@ -197,23 +197,31 @@ contract RootManager is ProposedOwnable, IRootManager, WatcherClient, DomainInde
...
+        // Roots are different, try to send via connector to spoke
+
+        // Try to send the message with appropriate encoded data and fees
+        // Continue on revert, but emit an event
+        try
+          IHubConnector(_connectors[i]).sendMessage{value: _fees[i]}(abi.encodePacked(_aggregateRoot), _encodedData[i])
+        {
+          // Update last sent root IFF successful
+          lastPropagatedRoot[domains[i]] = _aggregateRoot;
+
```
skeletor-spaceman, line 217: [Optional]
the `RootPropagated` is now a bit misleading since it  the root was propagated on ALL domains.
would it make sense to move and rename `emit RootPropagated(_aggregateRoot, _count, domainsHash);` here?
to: `emit DomainRootPropagated(_aggregateRoot, _count, domainsHash, domains[i], _connectors[i]);`
feel free to remove unnecessary params from the `DomainRootPropagated` event

I'm not sure how would this impact individual L2/Sidechain relayers that trigger on this event.

jakekidd: Emitting for each domain would add to the O(N) cost here.

Personally, I don't think it will mislead. If we fail to propagate to any 1 domain, then that 1 domain will never emit an `AggregateRootReceived` event:
https://github.com/connext/nxtp/blob/10e5df93b125b508bb53cb6393d6a836d5824391/packages/deployments/contracts/contracts/messaging/connectors/SpokeConnector.sol#L443

I would rather just treat this event as an indicator that "propagate was attempted for all domains"

And use the `PropagateFailed` event as an indicator that it failed for any particular domain. Subgraph mapping can use the failure event to track domains for which we failed to send root.

jakekidd: Worth saying this is a good note for the edge case in which ALL domains failed - in which case, we'd be emitting `RootPropagated` and N `PropagateFailed` events which would be kinda redundant/pointless

PR #2741, opened 2022-12-15
Deploy messaging only

LayneHaber:
## Description

Updates the init script to:
- optionally initialize the liquidity layer
- optionally send init transactions via multi send (in a single transaction)

Current status:
- Has utilities in place to upgrade using ^^
- Has command line options to specify the type of init

Remaining:
- [ ] Testing
- [ ] Update `setupMessaging`, top-level `init` to use multisend
- [ ] (optional) Implement only liquidity layer upgrading

## Type of change

- [x] Refactoring

## Related Issue(s)

Fixes


packages/deployments/contracts/src/init/init.ts
```diff
@@ -53,6 +57,21 @@ export const sanitizeAndInit = async () => {
...
+  if (typeof liquidity !== "boolean") {
+    throw new Error(`"liquidity" option must be a boolean, liquidity: ${liquidity}`);
+  }
+
+  if (!messaging) {
+    throw new Error(
+      `"messaging" option must be true, messaging: ${messaging}. (can be false in the future, but this script should be updated)`,
+    );
+  }
+  // TODO: when messaging
```
jakekidd, line 69: ?

packages/deployments/contracts/src/init/helpers/tx.ts
```diff
@@ -78,9 +101,90 @@ export const updateIfNeeded = async <T>(schema: CallSchema<T>): Promise<void> =>
...
+  return {
+    checkResult:
+      desired != undefined
+        ? {
+            method: readCall,
+            desired,
+          }
+        : undefined,
+    data: getMultisendTransaction(
+      "delegatecall",
```
jakekidd, line 163: shouldnt it be a call if we're updating values?

packages/deployments/contracts/src/init/init.ts
LayneHaber: honestly idk -- update when we want to init only liquidity layer is the guess?

LayneHaber:
closing

PR #2920, opened 2023-01-04
Enforce relayer fee in sequencer

wanglonghong:
## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes


packages/utils/src/peripherals/relayer.ts
```diff
@@ -0,0 +1,114 @@
...
+const relayerBufferPercentage = 20; // 20% bump on total estimated relayer fee
+
+export const calculateRelayerFee = async (
+  params: {
+    originDomain: string;
+    destinationDomain: string;
+    originNativeToken?: string;
+    destinationNativeToken?: string;
+    callDataGasAmount?: number;
+    isHighPriority?: boolean;
```
jakekidd, line 19: What does the high priority arg mean exactly?

Also it would be nice to make this a separate type (could go in this same file here I think?)

packages/agents/sdk/src/sdkBase.ts
```diff
@@ -203,91 +195,7 @@ export class NxtpSdkBase extends NxtpSdkShared {
...
-            .div(impactedOriginTokenPrice)
-            .div(BigNumber.from(10).pow(destinationTokenDecimals - originTokenDecimals));
-
-    this.logger.info("Method end", requestContext, methodContext, {
-      bumpedFee,
-      originTokenPrice,
-      destinationTokenPrice,
-      relayerFeeInOrginNativeAsset,
-    });
+    const relayerFeeInOrginNativeAsset = await calculateRelayerFee(params, this.chainData, this.logger);
```
jakekidd, line 198: nit: relayerFeeInOr**i**ginNativeAsset

packages/utils/src/peripherals/relayer.ts
wanglonghong: high priority is used to calculate fee using gelato api.

packages/agents/sdk/src/sdkBase.ts
wanglonghong: good catch!

packages/agents/sequencer/src/config.ts
```diff
@@ -95,6 +96,13 @@ export const getEnvConfig = (
       : configJson.relayers
       ? configJson.relayers
       : configFile.relayers,
+    relayerFeeTolerance: process.env.RELAYER_FEE_TOLERANCE
```
jakekidd, line 99: this looks good. we may wanna add an API endpoint for retrieving this number in the near future, so users can do an accurate calc on SDK-side as to whether the sequencer will reject their transfer

packages/agents/sequencer/src/lib/errors/relayer.ts
```diff
@@ -5,3 +5,9 @@ export class RelayerSendFailed extends NxtpError {
     super(`Relayer Send Failed`, context, RelayerSendFailed.name);
   }
 }
+
+export class NotEnoughRelayerFee extends NxtpError {
```
jakekidd, line 9: We should specify in the error what the relayer fee provided was, which transfer this is regarding, and how much the relayer fee *should be* in order to be sufficient

jakekidd: Alternatively we could just specify in the context object, that's fine

packages/utils/src/peripherals/relayer.ts
```diff
@@ -0,0 +1,114 @@
...
+
+  // TODO: Convert the estimatedRelayerFee to the originNativeToken
+  const [originTokenPrice, destinationTokenPrice, originTokenDecimals, destinationTokenDecimals] = await Promise.all([
+    getConversionRate(originChainId, undefined, undefined),
+    getConversionRate(destinationChainId, undefined, undefined),
+    getDecimalsForAsset(originNativeToken, originChainId, undefined, chainData),
+    getDecimalsForAsset(destinationNativeToken, destinationChainId, undefined, chainData),
+  ]);
+
+  if (originTokenPrice == 0 || destinationTokenPrice == 0) {
```
jakekidd, line 86: What if the origin token has a price of 0 but the destination has a price of > 0? I can't imagine this ever being the case but... idk, maybe we should throw an error? Because the destination is where we're ultimately paying for the tx...

Weird edge case, but worth accounting for. I think throw an error here if `originTokenPrice == 0 && destinationTokenPrice > 0`

if origin native token is 0, that would mean user would have no way to pay fees using a worthless native token... but I'm more worried about the edge case where the resource fails (i.e. `getConversionRate`'s pricing is incorrect)

Alternatively if `originTokenPrice > 0 && destinationTokenPrice == 0` we actually don't care about paying fees at all - it's free on the destination! So we should return `BigNumber.from(0)` like we're already doing in that case.

packages/agents/sequencer/src/lib/errors/relayer.ts
wanglonghong: yes it was actually being specified in the context object

packages/utils/src/peripherals/relayer.ts
```diff
@@ -0,0 +1,114 @@
...
+      destinationChainId,
+      executeGasAmount,
+      callDataGasAmount,
+    });
+  }
+
+  // add relayerFee bump to estimatedRelayerFee
+  const bumpedFee = estimatedRelayerFee.add(estimatedRelayerFee.mul(BigNumber.from(relayerBufferPercentage)).div(100));
+
+  // TODO: Convert the estimatedRelayerFee to the originNativeToken
```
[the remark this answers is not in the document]
wanglonghong: Lmao

packages/utils/src/peripherals/relayer.ts
wanglonghong: getConversionRate is the gelato utility function which was getting used in nxtp v1. This is a util package so I guess its better to return `0` and process it on agents rather than throwing an error.

packages/utils/src/peripherals/relayer.ts
```diff
@@ -0,0 +1,114 @@
...
+      originChainId,
+      destinationNativeToken,
+      destinationChainId,
+      executeGasAmount,
+      callDataGasAmount,
+    });
+  }
+
+  // add relayerFee bump to estimatedRelayerFee
+  const bumpedFee = estimatedRelayerFee.add(estimatedRelayerFee.mul(BigNumber.from(relayerBufferPercentage)).div(100));
```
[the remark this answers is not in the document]
wanglonghong: Its not tolerance stuff and this function is also used to calculate relayer fee needed which would be actually paid to gelato. bumpedFee is required to submit a tx successfully through gelato relayer.

PR #2922, opened 2023-01-04
2916 assets

sanchaymittal:
## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes


packages/agents/sdk/src/sdkShared.ts
```diff
@@ -92,6 +92,14 @@ export class NxtpSdkShared {
     return undefined;
   }
 
+  async getAssetsData(): Promise<any> {
```
rhlsthrm, line 95: Can we add a type for this? Since it's a SDK response it's public and it shouldn't be `any`.

rhlsthrm (changes requested):

packages/agents/sdk/src/sdkShared.ts
sanchaymittal: added it!

PR #2991, opened 2023-01-10
2986 pause

### LINKED ISSUE #2908: QA: Adjust txservice interface to make domain/chain mapping more intuitive
**Describe the bug**
This is more of a QA issue than a bug but the following method interface is really confusing:

```typescript
const encoded = await txservice.readTx({
          chainId: +domain,
          to: connext.address,
          data: connextInterface.encodeFunctionData("paused", []),
});
```

We input chainId as domain? Yes.
What happens if we input chainId? Well, it fails, because it's expecting domain.
Why do we input domain and not chain ID? Because that's how the mapping is keyed for providers/wallets/etc, i.e. all the stuff that a multi-chain txservice is concerned with.

This is a bit unexpected behavior and has wasted a engineers' time because they thought they found a bug (I mean, it looks like someone made a mistake in the above, inputting domain as the chain ID when clearly those are separate things).

**How Fix**
Just make it so all the input arg types for `readTx` and other methods use `domainId` instead of `chainId`. Technically the easiest fix by far! Just a naming thing and will work out perfectly.

Improve further on this by letting consumer specify `domainId:` **OR** `chainId:` in the input arg types (so `domainId: +domain` in the above example). This would mean we would need to change consumer usage across the board though, but would make things much more clear. Also, we would have to translate from input chain ID to domain ID in that case but that should be nbd.

Remember to do this for both `readTx` and `sendTx` as well as anything else that uses input `chainId`.

Example above SHOULD BE:
```typescript
const encoded = await txservice.readTx({
          domainId: +domain,
          to: connext.address,
          data: connextInterface.encodeFunctionData("paused", []),
});
```

### LINKED ISSUE #2986: Watcher should pause successfully
**Describe the bug**
Watcher failed to pause contracts during internal firedrill.

**To Reproduce**
1. Put watcher in pausable state
2. Watch logs for failure.

**Additional context**

LayneHaber:
## Description

Fixes #2986 
Fixes #2908 

## Type of change

- [x] Refactoring

- [x] pausing should happen concurrently across all domains not serially 
- [x] logs for failure should include domain in log
- [x] should send transaction at elevated gas not standard

## Related Issue(s)

Fixes

### DIFF
--- a/packages/adapters/relayer/src/connext/connext.ts
+++ b/packages/adapters/relayer/src/connext/connext.ts
@@ -51,8 +51,8 @@ export const connextRelayerSend = async (
     from: relayerAddress,
   });
 
-  const gas = await chainReader.getGasEstimateWithRevertCode(Number(domain), {
-    chainId,
+  const gas = await chainReader.getGasEstimateWithRevertCode({
+    domain: +domain,
     to: destinationAddress,
     data: encodedData,
     from: relayerAddress,
--- a/packages/adapters/relayer/src/gelato/gelato.ts
+++ b/packages/adapters/relayer/src/gelato/gelato.ts
@@ -216,8 +216,8 @@ export const send = async (
     from: relayerAddress,
   });
 
-  const gas = await chainReader.getGasEstimateWithRevertCode(Number(domain), {
-    chainId,
+  const gas = await chainReader.getGasEstimateWithRevertCode({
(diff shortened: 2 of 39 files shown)


wanglonghong (commented):
No much context on watcher stuff so I reviewed txservice and other agents stuff. Looks good.

PR #2992, opened 2023-01-10
Option to automatically handle unwrapping WETH on destination.

jakekidd:
## Description
Fixes https://github.com/connext/monorepo/issues/2946

## Type of change

- [ ] Refactoring


packages/deployments/contracts/contracts/shared/libraries/Unwrapper.sol
```diff
@@ -0,0 +1,91 @@
...
+    uint32,
+    bytes memory callData
+  ) external returns (bytes memory) {
+    // Get the target recipient, which should be in the callData.
+    address recipient = abi.decode(callData, (address));
+    // Sanity check: recipient is non-zero.
+    require(recipient != address(0), "unwrap: !recipient");
+
+    // Sanity check: asset we've received matches our target wrapper.
+    require(asset == address(WRAPPER), "unwrap: !asset");
```
LayneHaber, line 76: note: the `asset` will still be delivered here in the event of failure, and will be unretrievable

LayneHaber (changes requested):

packages/agents/sdk/src/interfaces/index.ts
```diff
@@ -29,3 +31,20 @@ export type AssetData = {
   key: string;
   id: string;
 };
+
+export const NxtpSdkXCallArgsSchema = Type.Object({
```
just-a-node, line 35: Change to SdkXCallArgsSchema cause I believe we are going to purge all "nxtp" references eventually.

packages/agents/sdk/src/sdkBase.ts
```diff
@@ -54,87 +48,162 @@ export class NxtpSdkBase extends NxtpSdkShared {
...
-    if (!chainId) {
-      chainId = await getChainIdFromDomain(origin, this.chainData);
+    // Sanitize arguments. Substitute default values as needed.
+    const amount = BigNumber.from(_amount);
+    let callData = args.callData ?? "0x";
+    const delegate = args.delegate ?? to; // Default to using the user's signer address as the delegate.
+    const relayerFee = BigNumber.from(_relayerFee ?? "0");
+    // TODO: Calculate estimate for relayer fee and include it in the call params.
+    // TODO: Should additionally make sure the asset is set to address(0) if the amount is 0.
+    // TODO: Use ajv validator.
```
just-a-node, line 105: We might want to sync our branches because I've added ajv validation on SdkBase methods and annotations for the methods in branch `sdk-validation`.

packages/agents/sdk/src/sdkBase.ts
```diff
@@ -54,87 +48,162 @@ export class NxtpSdkBase extends NxtpSdkShared {
...
+    if (unwrapNativeOnDestination) {
+      // Sanity check: We'll need to overwrite the callData with an unwrapping call. If a `callData` argument
+      // is also specified, we throw to ensure the user's callData isn't being overwritten unexpectedly.
+      // TODO: Implement an xReceive Multisend call to have the option to combine the unwrapping AND the user's
+      // original callData!
+      if (callData !== "0x") {
+        throw new CannotUnwrapOnDestination("`callData` argument must be empty!" + ` callData specified: ${callData}`);
+      }
+      // NOTE: We don't need to check on-chain to ensure destination Unwrapper supports unwrapping the
+      // target native token; if it didn't, it wouldn't be deployed and therefore wouldn't be configured.
```
just-a-node, line 116: What happens if someone calls this with:
```
asset: something other than ETH
wrapNativeOnOrigin: true
unwrapNativeOnDestination: true
```

Will probably revert on the call? Can we check that the asset is actually wrappable and throw a good error?

packages/deployments/contracts/contracts/shared/libraries/Unwrapper.sol
```diff
@@ -0,0 +1,91 @@
...
+interface Wrapper {
+  function withdraw(uint256 wad) external;
+
+  function transfer(address dst, uint256 wad) external returns (bool);
+}
+
+/**
+ * @notice A utility contract for unwrapping native tokens at the destination.
+ */
+contract Unwrapper is IXReceiver {
```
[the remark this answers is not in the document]
kryptoklob: specifically, `receive` and/or `fallback` methods (that are payable)

packages/deployments/contracts/contracts/shared/libraries/Unwrapper.sol
```diff
@@ -0,0 +1,91 @@
...
+    uint256 amount,
+    address asset,
+    address,
+    uint32,
+    bytes memory callData
+  ) external returns (bytes memory) {
+    // Get the target recipient, which should be in the callData.
+    address recipient = abi.decode(callData, (address));
+    // Sanity check: recipient is non-zero.
+    require(recipient != address(0), "unwrap: !recipient");
```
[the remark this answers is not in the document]
jakekidd: If there's no recipient, what should we do here? We could send to `originSender` as a sort of hail mary?

I'm going to add a mapping for `claimableWrappedEth` for users to later rescue their funds if they need to... we could set it so the `originSender` can rescue their funds... but ultimately this incurs risk of `originSender` being a random contract on origin.

packages/agents/sdk/src/sdkBase.ts
jakekidd: For `wrapNativeOnOrigin`:
At some point the consumer will call eth_estimateGas with the tx we've made and it will revert. But I've thought about this - it's kind of tough because we'd need to have a way to evaluate whether it *is* in fact *a* wrapper contract... we could of course have *our* set of "official" wrapper contracts (like as part of chainData or something) and just check if this is one of those assets... ultimately the only really sustainable solution here adds on to our responsibility of custodially managing "approved/acceptable" assets. I just feel like that responsibility should come from consumer.

Okay but one thing we could do I suppose is, if we have an injected provider, call eth_estimateGas ourselves on the "deposit" and "approve" method calls and make sure those pass, but that kind of feels like it's outside the scope of this method. After all, why aren't we calling eth_estimateGas on `xcall` itself inside this method? We just take for granted that the configured connext contract address in the SDK will be the official connext address.

jakekidd: For `unwrapNativeOnDestination`:
There's a bunch of diff failure paths here.

1. If the user sends a wrapped token that doesn't match the wrapper contract on destination (e.g. sending WETH to Polygon and then wanting it to unwrapNativeOnDestination won't work, because Unwrapper will point to the WMATIC contract), we'll need to have a fallback route for handling that in the contract itself of course. But can we prevent that from happening here in the SDK, at the outset? Again, we could keep a sort of custodial DB/mapping here that will tell us  what the sending asset *should* be if we want to unwrap on destination (i.e. "tell me what the wrapped polygon native token address is on origin - is it the token we're sending?").

2. We use the `to` address as the "intended recipient" argument for Unwrapper, since we have to replace `to` with the Unwrapper's address here. If the user sends xcall that's missing their `to` address... we effectively prevent that here in the SDK by checking to make sure that a valid `to` value is provided (as in, the recipient can't be address(0), or we throw an error - I just added a check for this!).

jakekidd: Actually for (1) I think we can check this. We should be able to do it this way:
1. get the canonicalId of the sending asset
2. get the adopted asset on destination
3. get the Unwrapper.WRAPPER property (there's a getter for that)
4. ensure Unwrapper.WRAPPER == adopted on destination. 

Right?

just-a-node: Looks right for (1).

For (2), how would someone get onto the `unwrapNativeOnDestination` flow without using SDK? They'd have to encode it in `callData` and at that point the `to` address *should* be the Unwrapper.

just-a-node: From internal chat:

for `wrapNativeOnOrigin`, isn't there only one supported WETH contract per chain? feels like we would be able to assert that `asset` is the WETH contract for this origin domain when you set `wrapNativeOnOrigin=true`

packages/deployments/contracts/contracts/shared/libraries/Unwrapper.sol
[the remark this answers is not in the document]
LayneHaber: hm well we could have a `claim` function where the `_originSender` is allowed to call `claim` for their balance.. 

really that should be the the agent though, but unfortunately thats not exposed. it could be include in the extra data though

packages/agents/sdk/src/sdkBase.ts
LayneHaber: re (2) see [this comment](https://github.com/connext/monorepo/pull/2992#discussion_r1067461552) i just added. we already have the concept of "oh shit i need to make adjustments on the destination chain" eoa in the agent, so i think we can lean on that if it becomes irretrievable for some reason

for 3 -- it seems like it should be a wrapper <> receiver pairing. each receiver contract has a designated wrapper (weth8, weth9, etc.) and the sdk determines which one to use based on the version it wrapped?

LayneHaber: actually an additional thought on (3) -- why does it matter? can we just try to send a `withdraw` via `.call` then if it fails we can assume the `asset` isn't weth? if it does, great. there's the obvious footguns of `xcall`ing a bad asset, but i dont know if it should matter what weth version it is

packages/agents/sdk/src/interfaces/index.ts
rhlsthrm: Agreed, no new code should ever use NXTP.

packages/deployments/contracts/contracts/shared/libraries/Unwrapper.sol
```diff
@@ -0,0 +1,159 @@
...
+    address asset,
+    address originSender,
+    uint32,
+    bytes memory callData
+  ) external onlyConnext returns (bytes memory) {
+    // Sanity check: amount is non-zero (otherwise, what are we unwrapping?).
+    require(amount != 0, "unwrap: !amount");
+
+    // Get the target recipient, which should be in the callData.
+    // TODO: What if decode results in missing address? i.e. empty callData or poorly formatted callData?
```
rhlsthrm, line 112: This is fine, we can't fully control for this. However it does become problematic because we will try to execute it forever. This is where the exponential backoff comes into play.

packages/deployments/contracts/deploy/04_deployUtils.ts
```diff
@@ -22,29 +40,22 @@ const func: DeployFunction = async (hre: HardhatRuntimeEnvironment): Promise<voi
...
+  // Multisend utility contract is used by the SDK to conveniently wrap ETH => WETH before
+  // making xcalls transferring WETH tokens.
+  deployContract({ hre, deployer, contractName: "MultiSend" });
+
+  /// MARK - Unwrapper
+  // NOTE: Unwrapper can be shared between staging and production environments; we do not
+  // deploy 1 for each.
+  // Unwrapper utility contract is used by the SDK to conveniently unwrap WETH => ETH on the
+  // transfer's destination chain after an xcall transferring WETH tokens.
+  deployContract({ hre, deployer, contractName: "Unwrapper" });
```
rhlsthrm, line 58: We'll need constructor args here right?

rhlsthrm (approved):
Looks good so far.

packages/deployments/contracts/contracts/shared/libraries/Unwrapper.sol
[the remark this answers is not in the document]
just-a-node: Keeping this simple, assets will be lost if recipient is zero or set incorrectly.

packages/deployments/contracts/contracts_forge/core/xreceivers/Unwrapper.t.sol
```diff
@@ -0,0 +1,239 @@
...
+      // vm.expectCall(_recipient, amount, abi.encode(""));
+
+      vm.expectEmit(true, true, true, true);
+      emit FundsDelivered(_recipient, address(0), amount);
+    } else {
+      vm.expectCall(asset, abi.encodeWithSelector(IERC20.transfer.selector, _recipient, amount));
+
+      vm.expectEmit(true, true, true, true);
+      emit FundsDelivered(_recipient, asset, amount);
+    }
```
LayneHaber, line 80:
```suggestion
    if (asset == address(0) || asset == MOCK_WRAPPER) {
      // Fund the target amount of ETH; our fake Wrapper contract certainly won't return any when we `withdraw`.
      vm.deal(address(unwrapper), amount);
      asset = address(0);
    } else {
      vm.expectCall(asset, abi.encodeWithSelector(IERC20.transfer.selector, _recipient, amount));
    }
    vm.expectEmit(true, true, true, true);
    emit FundsDelivered(_recipient, asset, amount);
```

packages/deployments/contracts/contracts_forge/core/xreceivers/Unwrapper.t.sol
```diff
@@ -0,0 +1,239 @@
...
+  function utils_recordInitialBalances(address _recipient) internal {
+    initialRecipientBalance = _recipient.balance;
+    initialUnwrapperBalance = address(unwrapper).balance;
+  }
+
+  function utils_assertSentEthToRecipient(address _recipient, uint256 amount) internal {
+    assertEq(
+      _recipient.balance,
+      initialRecipientBalance + amount,
+      "recipient didn't receive ETH! did we record initial balances?"
```
LayneHaber, line 95: error message huge

packages/deployments/contracts/contracts_forge/core/xreceivers/Unwrapper.t.sol
```diff
@@ -0,0 +1,239 @@
...
+    // as if it did not. Assets will always be sweepable at a later point.
+    utils_setUpMockErc20(false);
+
+    utils_expectRecipientToReceive(recipient, asset, amount);
+
+    vm.prank(MOCK_CONNEXT);
+    unwrapper.xReceive(transferId, amount, asset, originSender, origin, callDataWithRecipient);
+  }
+
+  function test_Unwrapper__xReceive_emitsEventIfNonWrapperAssetAndERC20TransferFails() public {
```
LayneHaber, line 177: we should have some assertions the asset remained on the wrapper and the recipient didn't get anything

packages/deployments/contracts/contracts_forge/core/xreceivers/Unwrapper.t.sol
```diff
@@ -0,0 +1,239 @@
...
+
+    vm.prank(MOCK_CONNEXT);
+    unwrapper.xReceive(transferId, amount, asset, originSender, origin, callDataWithRecipient);
+  }
+
+  // ============ Unwrapper.sweep ============
+  function test_Unwrapper__sweep_shouldSendERC20ToRecipient(uint256 amount) public {
+    if (amount == 0) {
+      return; // Skip 0 amount case.
+    }
```
LayneHaber, line 192:
```suggestion
    vm.assume(amount > 0);
```

packages/deployments/contracts/contracts_forge/core/xreceivers/Unwrapper.t.sol
```diff
@@ -0,0 +1,239 @@
...
+  function utils_expectRecipientToReceive(
+    address _recipient,
+    address asset,
+    uint256 amount
+  ) internal {
+    if (asset == address(0) || asset == MOCK_WRAPPER) {
+      // Fund the target amount of ETH; our fake Wrapper contract certainly won't return any when we `withdraw`.
+      vm.deal(address(unwrapper), amount);
+
+      // vm.expectCall(_recipient, amount, abi.encode(""));
```
LayneHaber, line 71: should expect a call to the `WRAPPER` to unwrap things

packages/deployments/contracts/contracts_forge/core/xreceivers/Unwrapper.t.sol
LayneHaber: erm wait this test isn't testing what it should, it should test that the function will revert

LayneHaber (approved):

PR #3129, opened 2023-01-19
Better cache cleanup on sequencer-publisher

wanglonghong:
## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes

### DIFF
--- a/packages/agents/sequencer/src/lib/operations/execute/fastpath.ts
+++ b/packages/agents/sequencer/src/lib/operations/execute/fastpath.ts
@@ -397,6 +397,10 @@ export const executeFastPathData = async (
       combinations: combinedBidsForRound,
     });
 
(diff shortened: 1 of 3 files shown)


rhlsthrm (approved):
Definitely need this.

rhlsthrm:
However this would not cause out of mem errors since the cache is a separate process.

wanglonghong:
> However this would not cause out of mem errors since the cache is a separate process.

maybe, by the way, would you post the possible causes for out of mem errors?

rhlsthrm:
> > However this would not cause out of mem errors since the cache is a separate process.
> 
> maybe, by the way, would you post the possible causes for out of mem errors?

We don't know, this is still under investigation. But it is something within the app, not the external cache.

PR #3148, opened 2023-01-20
3144 watcher pause

### LINKED ISSUE #3144: Watcher Mistaken Firing
## Background

- Watcher fired `pause` with the following error:

```
Connext Watcher Alert - totalMintedAssets (619480292707) is less than or equal to totalLockedAssets (32) for 0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48
```

despite >700K in USDC on mainnet chains.

### Tasks

- [x] to have consistent request context for the entirety of an interval
- [x] to have the same debug logs around direct data used and retrieved for totalMintedAssets
- [x] to have the same debug logs around direct data used and retrieved for totalLockedAssets
- [x] enforce minimum providers on watcher
- [x] log and pull the latest transaction hashes from the previous unalerting interval
- [x] allow txservice config to specify a fallback provider quorum, increase this to >1 in the watcher

### Linked Issues & Documentation

- BetterUptime incident: https://betteruptime.com/team/40435/incidents/330137557

LayneHaber:
## Description

Fixes #3144 

## Type of change

- [x] Refactoring

## Related Issue(s)

Fixes


packages/agents/watcher/src/config.ts
```diff
@@ -52,15 +52,27 @@ export const getEnvConfig = (): WatcherConfig => {
...
   }
   // return configFile;
+  const parsedChains: object = process.env.WATCHER_CHAIN_CONFIG
+    ? JSON.parse(process.env.WATCHER_CHAIN_CONFIG)
+    : configJson.chains
+    ? configJson.chains
+    : configFile.chains;
+
+  const chains: any = {};
+  Object.entries(parsedChains).map(([key, values]) => {
```
jakekidd, line 62: You made a sep. sanitized `chains` object but didn't end up using it? It's using `parsedChains` below

jakekidd: i fix

packages/adapters/watcher/src/adapter.ts
```diff
@@ -151,4 +170,51 @@ export class WatcherAdapter {
...
+  private beginRecordingTransactions(): void {
+    // clear out all the existing transactions. also ensures an array will always be present
+    this.clearTransactions();
+    // start listeners for each transaction
+    this.context.domains.forEach((_domain) => {
+      const domain = +_domain;
+      const chain = domainToChainId(domain);
+      const provider = this.context.txservice.getProvider(domain);
+      // FIXME: the only way to access events is by using the fallback provider, which
+      // will soon be deprecated
```
jakekidd, line 210: alternatively you can have an ethers Contract instance that has attached listeners
https://docs.ethers.org/v5/api/contract/contract/#Contract-listenerCount
Will technically be inefficient as we'll have a redundant provider nested in the instance, but at least it won't be a temporary solution / obsoleted in the future

packages/adapters/watcher/src/adapter.ts
```diff
@@ -8,30 +9,48 @@ import {
...
 export class WatcherAdapter {
   private readonly verifiers: Verifier[];
   private readonly pauser: Pauser;
+  private readonly context: VerifierContext;
+
+  // transactions sent to connext within the polling interval. cleared after
+  // each time `checkInvariants` is called
+  private readonly latestTransactions: Record<string, string[]> = {};
+
+  private push = false;
```
jakekidd, line 27: what's this used for?

packages/adapters/watcher/src/adapter.ts
```diff
@@ -8,30 +9,48 @@ import {
...
+  // paused. will hold all transactions between checks that have gone to
+  // connext
+  private transactionSnapshot: Record<string, string[]> | undefined;
 
   constructor(context: VerifierContext, assets: AssetInfo[]) {
+    this.context = context;
     // Should set all applicable verifiers here!
     this.verifiers = [new AssetVerifier(context, assets)];
     this.pauser = new Pauser(context);
+    this.beginRecordingTransactions();
```
jakekidd, line 39: Feels slightly dangerous, as we'll basically be piling on txs indefinitely until the consumer calls `checkInvariants`...

Right now, we know the consumer will call `checkInvariants` frequently enough that this should cause no concern... anyway, just making a mental note for myself here about this. We could make sure we're fixing this problem by capping the number of txs really high so we have a built-in failsafe in case we start consuming a lot of memory

packages/agents/watcher/src/operations/validateAndPause.ts
```diff
@@ -1,31 +1,33 @@
...
     adapters: { watcher },
     logger,
     config,
   } = getContext();
-  const methodContext = createMethodContext("pauseAndAlert");
+  const methodContext = createMethodContext(pauseAndAlert.name);
 
   const domains = Object.keys(config.chains);
-  logger.warn("Pausing contracts!!!", requestContext, methodContext, { reason });
+  logger.warn("Pausing contracts!!!", requestContext, methodContext, { reason, transactions, domains });
```
jakekidd, line 30: Good first step here; I think ideally we should have some way to view these transactions in the alert itself, but obviously it's going to be a really long list.

jakekidd: Would be ideal to have a table of txs we can view somewhere that shows amounts being transferred / delivered of diff tokens (even more ideally, ordered by amount).

What we want to provide here is a list of "suspects": at some point, we violated the invariant because some subset of these transactions are minting more than should be possible.

Technically, you can present these transactions chronologically and 1 of the txs will be the one that pushes us over the maximum amount we'd expect to be bridged/minted, BUT that tx isn't necessarily the perpetrator. It could be that the attacker pushed us right UP to the limit with fraudulent mints, and a subsequent tx is what actually did it. It could also be a series of txs that does this.

There's no telling just from amounts alone which txs are fraudulent... or is there? Important to ask: why do we want to know which fraudulent tx did the dirty? We can track down EOA, etc. but really it doesn't change the fact that we should be pausing.

Basically, I'm realizing that the critical info here we want to know is which of the txs' mints have a corresponding deposit on the sending chain!

If we check every tx's `transferId` and then check on the tx's `origin` whether an appropriate `xcall` exists that involves a deposit for the expected amount, we can quickly deduce whether there actually ARE fraudulent txs and which txs are fraudulent and which are not!

packages/adapters/watcher/src/adapter.ts
LayneHaber: oh we can delete this, looks like i forgot to do that

packages/adapters/watcher/src/adapter.ts
LayneHaber: this establishes the listeners, but the transactions are cleared between non-alerting intervals. 

can change to between all intervals as there are the snapshots

LayneHaber: but keep in mind the memory buildup would only happen after there had been an alert, so wouldn't really be top of mind

packages/agents/watcher/src/operations/validateAndPause.ts
LayneHaber: cleared after each interval

packages/adapters/watcher/src/adapter.ts
LayneHaber: cleared after each interval

packages/agents/watcher/src/operations/validateAndPause.ts
LayneHaber: created an issue to track separately: https://github.com/connext/monorepo/issues/3196

PR #3174, opened 2023-01-22
Add config for quorum to RpcProviderAggregator; require quorum is met if specified

jakekidd:
## Description

Now we can specify a quorum (consensus) must be met in order to validate returned values from providers.

When a quorum is specified, we go from being stringent/economical with our RPC requests to a mode where we consult all available RPC providers and require that at least Q providers agree on a return value for the request, where Q is the configured quorum.

An unconfigured quorum or a quorum value of 1 means we continue using the provider shuffling and only consult 1 provider at a time until we get a response.

Having a quorum is valuable for agents that need maximum accuracy to operate: namely watcher.

## Type of change

- [ ] Refactoring

### DIFF
--- a/packages/adapters/txservice/src/aggregator.ts
+++ b/packages/adapters/txservice/src/aggregator.ts
@@ -1,5 +1,6 @@
 import {
   createLoggingContext,
+  createRequestContext,
   delay,
   ERC20Abi,
   jsonifyError,
@@ -730,6 +731,20 @@ export class RpcProviderAggregator {
         }
         // Technically it could be multiple top responses...
         const topResponses = Array.from(counts.keys()).filter((k) => counts.get(k)! === maxCount);
+        if (topResponses.length > 0) {
+          // Did we get multiple conflicting top responses? Worth logging.
+          this.logger.info(
+            "Received conflicting top responses from RPC providers.",
+            createRequestContext(this.execute.name),
+            undefined,
+            {
+              topResponses,
+              providersCount: this.providers.length,
+              responsesCount: filteredResults.length,
+              requiredQuorum: quorum,
+            },
+          );
+        }
         return topResponses[0];
       }
     } else {
--- a/packages/adapters/txservice/src/aggregator.ts
+++ b/packages/adapters/txservice/src/aggregator.ts
@@ -26,6 +26,7 @@ import {
   OnchainTransaction,
   StallTimeout,
   WriteTransaction,
+  QuorumNotMet,
 } from "./shared";
 import { axiosGet } from "./mockable";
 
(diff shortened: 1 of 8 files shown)


packages/adapters/txservice/src/aggregator.ts
```diff
@@ -678,25 +679,71 @@ export class RpcProviderAggregator {
...
+      } else {
+        // e.g. a TransactionReverted, TransactionReplaced, etc.
+        throw error;
+      }
+    };
+
+    const quorum = this.config.quorum ?? 1;
+    if (quorum > 1) {
+      // Consult ALL providers.
+      const results: (T | undefined)[] = await Promise.all(
```
LayneHaber, line 701: this will actually be problematic if the providers hang -- we should have a cutoff so these promises dont just wait forever

packages/adapters/txservice/src/aggregator.ts
```diff
@@ -678,25 +679,71 @@ export class RpcProviderAggregator {
...
+        }),
+      );
+      // Filter out undefined results.
+      // NOTE: If there aren't any defined results, we'll proceed out of this code block and throw the
+      // RpcError at the end of this method.
+      const filteredResults: T[] = results.filter((item) => item !== undefined) as T[];
+      if (filteredResults.length > 0) {
+        // Pick the most common answer.
+        let counts: Map<T, number> = new Map();
+        counts = filteredResults.reduce((counts, item) => {
```
LayneHaber, line 718: `item` may not be a string since its part of the returned provider value

LayneHaber: and im not sure if this keying will work to identify the same returned object

packages/adapters/txservice/src/aggregator.ts
```diff
@@ -678,25 +679,71 @@ export class RpcProviderAggregator {
...
+        const maxCount = Math.max(...Array.from(counts.values()));
+        if (maxCount < quorum) {
+          // Quorum is not met: we should toss this response as it could be unreliable.
+          throw new QuorumNotMet(maxCount, quorum, {
+            errors,
+            providersCount: this.providers.length,
+            responsesCount: filteredResults.length,
+          });
+        }
+        // Technically it could be multiple top responses...
```
LayneHaber, line 731: we should log if there are multiple conflicting top responses at least

packages/adapters/txservice/src/aggregator.ts
jakekidd: That's what syncprovider is for!

It does this internally. Check out SyncProvider.send in packages/adapters/txservice/src/shared/provider.ts

packages/adapters/txservice/src/aggregator.ts
jakekidd: Sure, agreed :)

packages/adapters/txservice/src/aggregator.ts
jakekidd: With `Map` you can key using a generic type afaik... wouldn't I be getting compile issues if this were true? When I highlight `item` at any point, it seems to think the generic type `T` is maintained

jakekidd: TBH I'm pretty sure it's `string` in all cases though

LayneHaber: provider responses? i dont think those are always strings. but im saying im not sure if its going to be indexed as the same value if its an object, not that it isn't technically supported. not sure though

packages/adapters/txservice/src/aggregator.ts
LayneHaber: oh okay awesome!

LayneHaber (approved):

PR #3207, opened 2023-01-24
3106  stableswap subgraph cartographer

liu-zhipeng:
## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes


packages/adapters/subgraph/src/lib/operations/queries.ts
```diff
@@ -829,3 +868,62 @@ export const getRootManagerMetaQuery = (domain: string) => {
...
+      ${queryString}
+    }
+  `;
+};
+
+const swapExchangeQueryString = (
+  prefix: string,
+  fromTimestamp: number,
+  maxBlockNumber?: number,
+  orderDirection: "asc" | "desc" = "desc",
```
preethamr, line 893: Should this be ascending by default ?

packages/agents/cartographer/poller/src/lib/operations/stableswap.ts
```diff
@@ -0,0 +1,89 @@
...
+    domains.map(async (domain) => {
+      const pools = await subgraph.getStableSwapPools(domain);
+      logger.info("Retrieved stable swap pools", requestContext, methodContext, {
+        domain,
+        pools: pools,
+        count: pools.length,
+      });
+
+      if (pools.length) {
+        await database.saveStableSwapPool(pools);
```
preethamr, line 30: Is there no need to checkpoint ?

packages/adapters/subgraph/src/lib/operations/queries.ts
```diff
@@ -829,3 +868,62 @@ export const getRootManagerMetaQuery = (domain: string) => {
     }
   `;
 };
+
+export const getStableSwapPoolsQuery = (domain: string) => {
+  const { config } = getContext();
+  const prefix = config.sources[domain].prefix;
+  const queryString = `
+  ${prefix}_stableSwaps ( 
+    first: 100
```
preethamr, line 877: In current state it would retrieve only the first hundred every time.
Is there a scenario where there are more than 100 ?

packages/adapters/database/src/client.ts
```diff
@@ -676,3 +716,25 @@ export const saveReceivedAggregateRoot = async (
...
+export const saveStableSwapExchange = async (
+  _swapExchanges: StableSwapExchange[],
+  _pool?: Pool | db.TxnClientForRepeatableRead,
+): Promise<void> => {
+  const poolToUse = _pool ?? pool;
+  const exchanges: s.stableswap_exchanges.Insertable[] = _swapExchanges
+    .map((m) => convertToDbStableSwapExchange(m))
+    .map(sanitizeNull);
+
+  await db.upsert("stableswap_exchanges", exchanges, ["id"]).run(poolToUse);
```
preethamr, line 739: Primary per the schema is (domain, id). Here it upserts based only on id.

preethamr (commented):
We need TF config to create lambda for the new stableswap service

preethamr:
@liu-zhipeng We also need to add betteruptime healthcheck entpoints, and corresponding config in TF

packages/adapters/subgraph/src/lib/operations/queries.ts
liu-zhipeng: Correct!

packages/agents/cartographer/poller/src/lib/operations/stableswap.ts
liu-zhipeng: Yes, no need. because we get 100 pools for each domain. I believe that max pool count of domain is less than 100.

packages/adapters/subgraph/src/lib/operations/queries.ts
liu-zhipeng: `first: 1000` will be pretty enough!

PR #3450, opened 2023-02-08
Set max limit on backoff

preethamr:
## Description

Set max limit on backoff in sequencer to a week.

## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes


just-a-node (approved):
24 hours? Maybe pull this out into a const in case we need to update it later.

PR #3500, opened 2023-02-13
[DNM] 3497 processfromroot should only process latest root message

rhlsthrm:
## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes


liu-zhipeng (approved):

packages/agents/lighthouse/src/tasks/processFromRoot/operations/processFromRoot.ts
```diff
@@ -76,7 +76,27 @@ export const processFromRoot = async () => {
...
+  const toMarkProcessed: RootMessage[] = [];
+
+  const latestUnprocessed: RootMessage[] = [];
+  Object.keys(byDomain).forEach((domain) => {
+    byDomain[domain].sort((a, b) => a.timestamp - b.timestamp);
+    latestUnprocessed.push(byDomain[domain].pop()!);
+    toMarkProcessed.push(...byDomain[domain]);
+  });
+
+  await database.markRootMessagesProcessed(toMarkProcessed);
```
preethamr, line 97: This could open a rolling window, and a new type for processing gap.

Consider the scenario where `sendWithRelayerWithBackup` keeps failing, while new roots for domain come in. Old roots (even ones that relayer did not execute) will be marked `processed` since there is a new root that supersedes the old ones. And this could potentially go on indefinitely, and cause a long chain of roots to be marked processed, when on-chain they are not.

preethamr (changes requested):
LG for normal operating conditions.

Posted about a rare but possible failure mode, that prob needs special handling.

packages/agents/lighthouse/src/tasks/processFromRoot/operations/processFromRoot.ts
rhlsthrm: This is a good point. Especially in the case of Arbitrum where the process will fail for a while until it succeeds. What's the best solution for this? I think maybe if it fails then we can try the next one in the list and so on?

preethamr: one way is we don't set processed here, while we pick and process the latest root as in this PR.
Instead, we set `processed` as additional steps around here:
https://github.com/connext/monorepo/blob/5a2c880ecdea540cdfe323a2592858469dbb4762/packages/agents/cartographer/poller/src/lib/operations/messages.ts#L144

For everything that is before the root, and still unprocessed.

rhlsthrm: I like your idea of saving it in the db, but I don't quite understand how to resolve the issue if the newest root that comes in is always failing. It seems like we would still need to try-catch and run through them all?

preethamr (approved):
LGTM

PR #3572, opened 2023-02-22
Update transfers with relayer fee updates

rhlsthrm:
## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes

### DIFF
--- a/packages/agents/cartographer/poller/src/lib/operations/transfers.ts
+++ b/packages/agents/cartographer/poller/src/lib/operations/transfers.ts
@@ -240,9 +240,13 @@ export const updateBackoffs = async (): Promise<void> => {
     increases.map((increase) => increase.transferId).concat(updates.map((update) => update.transferId)),
   );
 
+  // filter by domain
+  const increasesByDomain: Record<string, string[]> = {};
+
   const increaseCheckpoints = domains
     .map((domain) => {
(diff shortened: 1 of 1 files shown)


packages/agents/cartographer/poller/src/lib/operations/transfers.ts
```diff
@@ -252,6 +256,13 @@ export const updateBackoffs = async (): Promise<void> => {
     })
     .filter((x) => !!x) as { domain: string; checkpoint: number }[];
 
+  // update transfers with relayer fee bumps
+  // query subgraphs to avoid recalculating relayer fee based on increases
```
preethamr, line 260: Should we do the same for slippage updates too ?

preethamr (approved):
LGTM
One minor comment

packages/agents/cartographer/poller/src/lib/operations/transfers.ts
rhlsthrm: We do, above.

rhlsthrm: https://github.com/connext/monorepo/pull/3572/files#diff-f23a559cc9ba71d6ad70ac0518924cac925aee6f3357b5482223ce1184ddb655R237

We have to handle this differently because the event does not emit the total relayer fee, just the increase.

PR #3597, opened 2023-02-23
Revert "Fix receiveLocal on carto transfers table"

preethamr:
Reverts connext/monorepo#3586


preethamr:
This fix would not work @rhlsthrm

PR #3609, opened 2023-02-27
Execute simulation/ Debugging tooling for execute at explorer

### LINKED ISSUE #3563: debugging and simulation tooling for execute transfer
## Problem
_What problem are we solving and why does it matter, ser?_ 

Debugging failing transfers at the destination is hard and had been reported multiple times by developers

## Proposed Solution
_Thoughts on solution design. We should do <insert idea> to solve this problem._
- [ ] Add simulation data to DB
- [ ] Add tenderly simulation

## Acceptance Criteria
_From an end tester and/or user perspective, the below criteria must be met to consider this done:_
- [ ] X happens

sanchaymittal:
## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes #3563

### DIFF
--- a/packages/agents/sequencer/src/lib/operations/relayer/fastpath.ts
+++ b/packages/agents/sequencer/src/lib/operations/relayer/fastpath.ts
@@ -1,4 +1,4 @@
-import { RequestContext, createLoggingContext, Bid, OriginTransfer } from "@connext/nxtp-utils";
+import { RequestContext, createLoggingContext, Bid, OriginTransfer, GELATO_RELAYER_ADDRESS } from "@connext/nxtp-utils";
 
 import { sendWithRelayerWithBackup } from "../../../mockable";
 import { getContext } from "../../../sequencer";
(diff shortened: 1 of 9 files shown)


packages/agents/sequencer/src/lib/operations/relayer/fastpath.ts
```diff
@@ -30,6 +30,17 @@ export const sendExecuteFastToRelayer = async (
 
   const executeEncodedData = await encodeExecuteFromBids(round, bids, transfer, requestContext);
 
+  // Simulation data for Execute transfer
+  const relayerFrom = await relayers[0].instance.getRelayerAddress(destinationChainId);
```
rhlsthrm, line 34: Not ideal lol. We could just use the hardcoded Gelato address here. It will always work on every chain.

rhlsthrm (approved):
Looks good, minor fix requested.

packages/agents/sequencer/src/lib/operations/relayer/fastpath.ts
sanchaymittal: we won't have gelato for every chain and ideally, this shouldn't be an issue with latency too. We only need a from address which is whitelisted.

rhlsthrm: Yes we don't have gelato but we still whitelist their relayer address.

sanchaymittal: resolved

PR #3631, opened 2023-03-01
Reset exponential backoff after a bump

wanglonghong:
## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes


rhlsthrm (commented):
https://github.com/connext/monorepo/blob/main/packages/agents/cartographer/poller/src/lib/operations/transfers.ts#L239

We actually already do this here for resetting backoff. This is only relevant for slow path FYI.

PR #3653, opened 2023-03-02
Fix to add error properly

rhlsthrm:
## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes

### DIFF
--- a/packages/agents/sequencer/src/lib/operations/execute/fastpath.ts
+++ b/packages/agents/sequencer/src/lib/operations/execute/fastpath.ts
@@ -252,6 +252,8 @@ export const executeFastPathData = async (
       transferId,
     });
 
+    await database.updateErrorStatus(transferId, XTransferErrorStatus.NoBidsReceived);
+
     return { taskId };
   }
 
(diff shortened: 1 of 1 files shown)


packages/agents/sequencer/src/lib/operations/execute/fastpath.ts
```diff
@@ -247,12 +247,8 @@ export const executeFastPathData = async (
   const bidsRoundMap = getBidsRoundMap(bids, config.auctionRoundDepth);
   const availableRoundIds = [...Object.keys(bidsRoundMap)].sort((a, b) => Number(a) - Number(b));
   if ([...Object.keys(bidsRoundMap)].length < 1) {
-    logger.warn("No rounds available for this transferId", requestContext, methodContext, {
-      bidsRoundMap,
-      transferId,
-    });
```
wanglonghong, line 253: IMO, it was useful to check the error status on DD. If we remove logging from here, I guess it would be difficult to figure out the error on DD because we don't have even transferId included in `NoBidsSent`

wanglonghong (changes requested):

packages/agents/sequencer/src/lib/operations/execute/fastpath.ts
rhlsthrm: But throwing the error should log it because the caller should log the error. The transferId should be part of requestContext anyways.

rhlsthrm: https://github.com/connext/monorepo/blob/3652-transfer-not-showing-up-as-nobidsreceived-even-though-no-bids-received/packages/agents/sequencer/src/sequencer.ts#L140 right here it will log. This is actually better, the practice should be throw errors in internal functions and try catch at the top level.

preethamr (approved):

sanchaymittal (approved):

PR #3809, opened 2023-03-20
Keep3r scripts

rhlsthrm:
## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes


packages/deployments/contracts/contracts/core/connext/helpers/RelayerProxyHub.sol
```diff
@@ -81,13 +253,68 @@ contract RelayerProxyHub is RelayerProxy {
...
+        data._message,
+        data._proof
+      );
+    }
+
+    if (fromChain == 137 || fromChain == 80001) {
+      IPolygonHubConnector(hubConnectors[fromChain]).receiveMessage(encodedData);
+    }
+
+    processedRootMessages[fromChain][l2Hash] = true;
```
0xng, line 318: I'd probably move this all the way up after the checks, just to follow the cei pattern. Also, is it correct to assume none of the functions called in this contract fail silently? As in, they return instead of reverting?

packages/deployments/contracts/contracts/core/connext/helpers/RelayerProxy.sol
```diff
@@ -288,6 +271,19 @@ contract RelayerProxy is ProposedOwnable, ReentrancyGuard, GelatoRelayFeeCollect
...
+   * For example, if priority is 3, then Autonolas will be able to work on blocks 0, 1, 2, and 3.
+   * @param sender The address of the caller
+   */
+  function isWorkableBySender(address sender) public view returns (bool) {
+    if (sender == autonolas) {
+      return block.number % 10 <= autonolasPriority;
+    } else {
+      return block.number % 10 > autonolasPriority;
+    }
+  }
```
0xGorilla, line 285: What do you think about making it a modifier, which contains the error as well?
One more thing, the `sender` arg should be named `_sender`

0xGorilla (approved):

PR #3831, opened 2023-03-22
3829 router backwards compatibility

preethamr:
## Description

Add backwards compatibility to routers to support relayer fee upgrade.

- [ ] Refactoring
- [x] Requires changes in customer code

## Related Issue(s)

Fixes


packages/adapters/subgraph/src/lib/helpers/parse.ts
```diff
@@ -73,6 +73,11 @@ export const originTransfer = (entity: any, asset: Record<string, AssetId>): Ori
     for (const relayerFee of entity.relayerFees) {
       relayerFees[relayerFee.asset] = relayerFee.fee;
     }
+  } else {
+    // TODO: Remove after relayer fee upgrade is completed on mainnet
```
LayneHaber, line 77: really this is after all the routers upgrade

packages/adapters/subgraph/src/lib/helpers/parse.ts
```diff
@@ -73,6 +73,11 @@ export const originTransfer = (entity: any, asset: Record<string, AssetId>): Ori
...
+    // INFO: https://github.com/connext/monorepo/issues/3811
+    // Handle entity from previous subgraph for backwards compatibility
+    relayerFees[transactingAsset] = entity?.relayerFee ?? "0";
```
LayneHaber, line 80: im confused here -- what exactly is the assumption this is embedding?

all transfers on non-updated routers will register the `entity?.relayerFee` with the `transactingAsset` as the asset? most will have native asset as the relayer fee asset, so I'm not sure how those will be accounted for in routers that aren't up to date

packages/adapters/subgraph/src/lib/helpers/parse.ts
preethamr: True. Updated the TODO.

packages/adapters/subgraph/src/lib/helpers/parse.ts
preethamr: Changed it to address zero, will be the same as how the DB migration defaults to it; for older data
Since routers don't actually use the relayer fee for anything, and the sequencer pulls it up on its own without relying on info passed by router, the relayer fee asset info stored in the router would not take effect.

LayneHaber (approved):

packages/adapters/subgraph/src/reader.ts
```diff
@@ -369,7 +376,12 @@ export class SubgraphReader {
   public async getOriginTransfers(agents: Map<string, SubgraphQueryMetaParams>): Promise<XTransfer[]> {
     const { execute, parser } = getHelpers();
     const { config } = getContext();
-    const xcalledXQuery = getOriginTransfersQuery(agents);
+    let xcalledXQuery = undefined;
+    try {
+      xcalledXQuery = getOriginTransfersQuery(agents);
+    } catch (err: any) {
```
rhlsthrm, line 382: Let's add a warning log here.

rhlsthrm (approved):

packages/adapters/subgraph/src/reader.ts
preethamr: Done

PR #3862, opened 2023-03-24
Fast path

### LINKED ISSUE #3847: UI checks liquidity with new isFastPathAvailable function
## Problem
Currently we say that fast liquidity is always available as a default 

## Proposed Solution
Add a way to estimate if there is enough liquidity as it's own function in the SDK

> the last optimization for calculation we did fetches the total of router liquidity available for an asset at provided destination network.
> And then compares to it's 70%
> But cause routerLiquidity call to carto was adding time to calculatRelayerFee calculation time, we removed it from the code and made true as default

just-a-node:
## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes #3847


rhlsthrm:
We might need to make a separate PR for this into prod.

rhlsthrm (approved):

PR #3887, opened 2023-03-28
Workflow fixes

rhlsthrm:
## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes

### DIFF
--- a/.github/workflows/build-test-deploy.yml
+++ b/.github/workflows/build-test-deploy.yml
@@ -540,7 +540,7 @@ jobs:
           images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
           tags: |
             type=sha,format=short
-            type=raw,value=${{ steps.build-and-push-router-watcher-image.outputs.docker_tag }}
+            type=raw,value=${{ steps.build-and-push-watcher-image.outputs.docker_tag }}
 
       - name: Build and push Docker image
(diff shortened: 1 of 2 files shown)


.github/workflows/build-test-deploy.yml
```diff
@@ -477,14 +523,24 @@ jobs:
...
+
       - name: Extract metadata (tags, labels) for Docker
         id: meta
         uses: docker/metadata-action@98669ae865ea3cffbcbaa878cf57c20bbf1c6c38
         with:
           images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
           tags: |
             type=sha,format=short
-            type=semver,pattern={{raw}}
+            type=raw,value=${{ steps.build-and-push-router-watcher-image.outputs.docker_tag }}
```
preethamr, line 543: This should instead be:
`type=raw,value=${{ steps.build-and-push-watcher-image.outputs.docker_tag }}`

preethamr (changes requested):

carlomazzaferro (approved):

.github/workflows/build-test-deploy.yml
rhlsthrm: Nice catch! Fixed

rhlsthrm:
Looks okay now, I think this will need to be tested in prod by doing a tag. @just-a-node @preethamr @sanchaymittal pls make sure this is in before doing the next SDK release so we can test the process.

PR #3969, opened 2023-04-12
feat: optimistic roots smart contracts

gotzenx:
## Description
Reduce gas consumption on the message layer to a fraction of the current cost by adding to the system the ability to operate in an optimistic mode when desired.
System will have two modes: `Optimistic Mode` and `Slow Mode` (current one).
Optimistic mode is based on an external agent fetching the `inboundRoots` and calculating the new `aggregateRoot` off-chain to propose on the RootManager contract.
The propose will have a dispute time where other off-chain agents will have the chance to verify that the proposed data is correct and dispute it if something is not right.
The way we propagate the aggregateRoots will not change.

## Type of change

- [ ] Refactoring

- Messages will reach L1 (RootManager) faster and without the need of going through AMBs
-

### DIFF
--- a/packages/deployments/contracts/contracts/messaging/RootManager.sol
+++ b/packages/deployments/contracts/contracts/messaging/RootManager.sol
@@ -33,13 +33,64 @@ contract RootManager is ProposedOwnable, IRootManager, WatcherClient, DomainInde
 
   event RootDiscarded(bytes32 fraudulentRoot);
 
+  event ProposerAdded(address proposer);
+
+  event ProposerRemoved(address proposer);
+
   event ConnectorAdded(uint32 domain, address connector, uint32[] domains, address[] connectors);
 
   event ConnectorRemoved(uint32 domain, address connector, uint32[] domains, address[] connectors, address caller);
 
   event PropagateFailed(uint32 domain, address connector);
 
+  event SlowModeActivated();
+
+  event OptimisticModeActivated();
+
+  event OptimisticRootPropagated(bytes32 aggregateRoot, bytes32 domainsHash);
+
+  event AggregateRootProposed(
+    uint256 snapshotId,
+    uint256 endOfDispute,
+    bytes32 aggregateRoot,
+    bytes32 baseRoot,
+    bytes32[] snapshotsRoots,
+    uint32[] domains
+  );
+
+  event ProposedRootFinalized(bytes32 aggregateRoot);
+
+  // ============ Errors ============
+
+  error RootManager_InvalidSnapshotId(uint256 snapshotId);
+
+  error RootManager_InvalidDomains();
+
+  error RootManager_InvalidAggregateRoot();
+
+  error RootManager_SlowModeOn();
+
+  error RootManager_OptimisticModeOn();
+
+  error RootManager_ProposeInProgress();
+
+  error RootManager_NotWhitelistedProposer(address caller);
+
+  error RootManager_EmptyFinalizedOptimisticRoot();
+
+  error RootManager_OldAggregateRoot();
+
   // ============ Properties ============
+
+  /**
+   * @notice The time watchers have to detect and invalidate the proposed root.
+   */
+  uint256 public constant DISPUTE_TIME = 30 minutes;
+
+  /**
+   * @notice Duration of the snapshot
+   */
+  uint256 public constant SNAPSHOT_DURATION = 30 minutes;
 
   /**
    * @notice Maximum number of values to dequeue from the queue in one sitting (one call of `propagate`
@@ -54,6 +105,35 @@ contract RootManager is ProposedOwnable, IRootManager, WatcherClient, DomainInde
   uint256 public delayBlocks;
 
   /**
+   * @notice The amount of inserted leaves prior to switching to optimistic mode
+   * @dev Used to prevent the propagation of an old aggregate root right after switching to Slow mode.
+   * After switching back to slow mode, if the count didnt change from previous slow mode, we consider the root as an old one since
+   * probably lot of optimistic roots have been propagated.
+   * Example: Propagation in slow mode happens, switch to Op Mode, multiple propagations happen, switch back to Slow mode.
+   * If at this point someone calls propagate and the queue is empty or non of elemenets are ready, _slowPropagate will
+   * call the dequeue function which will return the old root and count before Op mode was activated. And since multiple
+   * propagations happened while in optimistic mode, the lastPropagatedRoot will be different than the old but current MERKLE.root
+   * which will lead to propagating a deprecated root.
+   */
+  uint256 public lastCountBeforeOpMode;
+
+  /**
+   * @notice True if the system is working in optimistic mode. Otherwise is working in slow mode
+   */
+  bool public optimisticMode;
+
+  /**
+   * @notice The last aggregate root we propagated to spoke chains. Used to prevent sending redundant
+   * aggregate roots in `propagate`.
+   */
+  bytes32 public lastPropagatedRoot;
+
+  /**
+  @notice The last finalized optimistic aggregate root.
+   */
+  bytes32 public finalizedOptimisticAggregateRoot;
+
+  /**
    * @notice Queue used for management of verification for inbound roots from spoke chains. Once
    * the verification period elapses, the inbound messages can be aggregated into the merkle tree
    * for propagation to spoke chains.
@@ -63,10 +143,9 @@ contract RootManager is ProposedOwnable, IRootManager, WatcherClient, DomainInde
   QueueLib.Queue public pendingInboundRoots;
 
   /**
-   * @notice The last aggregate root we propagated to spoke chains. Used to prevent sending redundant
-   * aggregate roots in `propagate`.
+  @notice The proposed aggregate root.
    */
-  bytes32 public lastPropagatedRoot;
+  ProposedData public proposedAggregateRoot;
 
   /**
    * @notice MerkleTreeManager contract instance. Will hold the active tree of aggregated inbound roots.
@@ -74,10 +153,43 @@ contract RootManager is ProposedOwnable, IRootManager, WatcherClient, DomainInde
    */
   MerkleTreeManager public immutable MERKLE;
 
+  /**
+   * @notice This is used for the `onlyProposers` modifier, which gates who
+   * can propose new roots using `proposeAggregateRoot`.
+   */
+  mapping(address => bool) public allowlistedProposers;
+
+  /**
+   * @notice Struct to store the proposed data
+   * @dev The list of the snapshots roots and the domains must be in the
+   * same order as the roots insertions on the tree.
+   * @param endOfDispute The timestamp when the dispute period is over
+   * @param aggregateRoot The new aggregate root
+   */
+  struct ProposedData {
+    uint256 endOfDispute;
+    bytes32 aggregateRoot;
+  }
+
   // ============ Modifiers ============
 
   modifier onlyConnector(uint32 _domain) {
     require(getConnectorForDomain(_domain) == msg.sender, "!connector");
+    _;
+  }
+
+  modifier checkDomains(uint32[] calldata _domains) {
+    if (keccak256(abi.encode(_domains)) != domainsHash) revert RootManager_InvalidDomains();
+    _;
+  }
+
+  modifier onlyOptimisticMode() {
+    if (!optimisticMode) revert RootManager_SlowModeOn();
+    _;
+  }
+
+  modifier onlyProposer() {
+    if (!allowlistedProposers[msg.sender]) revert RootManager_NotWhitelistedProposer(msg.sender);
     _;
   }
 
@@ -100,6 +212,7 @@ contract RootManager is ProposedOwnable, IRootManager, WatcherClient, DomainInde
     MERKLE = MerkleTreeManager(_merkle);
 
     delayBlocks = _delayBlocks;
+    optimisticMode = true;
 
     // Initialize pending inbound root queue.
     pendingInboundRoots.initialize();
@@ -112,6 +225,24 @@ contract RootManager is ProposedOwnable, IRootManager, WatcherClient, DomainInde
   }
 
   // ============ Admin Functions ============
+
+  /**
+   * @notice Adds a proposer to the allowlist.
+   * @dev Only allowlisted proposers can call `proposeAggregateRoot`.
+   */
+  function addProposer(address _proposer) external onlyOwner {
+    allowlistedProposers[_proposer] = true;
+    emit ProposerAdded(_proposer);
+  }
+
+  /**
+   * @notice Removes a proposer from the allowlist.
+   * @dev Only allowlisted proposers can call `proposeAggregateRoot`.
+   */
+  function removeProposer(address _proposer) external onlyOwner {
+    delete allowlistedProposers[_proposer];
+    emit ProposerRemoved(_proposer);
+  }
 
   /**
    * @notice Set the `delayBlocks`, the period in blocks over which an incoming message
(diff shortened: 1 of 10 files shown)


packages/deployments/contracts/contracts/messaging/RootManager.sol
```diff
@@ -265,4 +538,32 @@ contract RootManager is ProposedOwnable, IRootManager, WatcherClient, DomainInde
...
+
+  /**
+   * @notice Watcher can set the system in slow mode.
+   * @dev Clear the proposed aggregate root.
+   */
+  function activateSlowMode() external onlyWatcher onlyOptimisticMode {
+    optimisticMode = false;
+
+    delete proposedAggregateRoot;
+    emit SlowModeActivated();
```
LayneHaber, line 550: we should make sure to emit the watcher that called this in the event

packages/deployments/contracts/contracts/messaging/RootManager.sol
```diff
@@ -235,6 +507,7 @@ contract RootManager is ProposedOwnable, IRootManager, WatcherClient, DomainInde
    * @param _inbound The inbound root coming from the given domain.
    */
   function aggregate(uint32 _domain, bytes32 _inbound) external whenNotPaused onlyConnector(_domain) {
+    if (optimisticMode) revert RootManager_OptimisticModeOn();
```
LayneHaber, line 510: @preethamr we should make sure that on the off-chain side we are halting the `processFromRoot` tasks appropriately

pretty sure we already talked about that though

packages/deployments/contracts/contracts/messaging/connectors/SpokeConnector.sol
```diff
@@ -54,6 +54,8 @@ abstract contract SpokeConnector is Connector, ConnectorManager, WatcherClient,
 
   event DelayBlocksUpdated(uint256 indexed updated, address caller);
 
+  event SnapshotRootSaved(uint256 snapshotId, bytes32 root, uint256 count);
```
LayneHaber, line 57: we should index the ids and roots (maybe count but eh)

packages/deployments/contracts/contracts/messaging/connectors/SpokeConnector.sol
```diff
@@ -90,6 +92,11 @@ abstract contract SpokeConnector is Connector, ConnectorManager, WatcherClient,
 
   // ============ Public Storage ============
 
+  /**
+   * @notice Duration of the snapshot
+   */
+  uint256 public constant SNAPSHOT_DURATION = 30 minutes;
```
LayneHaber, line 98: see above comments re: time (one thing that makes this annoying is it would have to be configured per-connector, so instantiated with the constructors)

packages/deployments/contracts/contracts/messaging/connectors/SpokeConnector.sol
```diff
@@ -591,4 +624,14 @@ abstract contract SpokeConnector is Connector, ConnectorManager, WatcherClient,
     // emit process results
     emit Process(_messageHash, _success, _returnData);
   }
+
+  /**
+   * @notice This function calculates the last completed snapshot id
+   * @return _lastCompletedSnapshotId The last completed snapshot id
+   */
+  function _getLastCompletedSnapshotId() internal view returns (uint256 _lastCompletedSnapshotId) {
```
LayneHaber, line 632: should we put this function in a library or something that both the root manager and spoke connector can use?

packages/deployments/contracts/contracts/messaging/RootManager.sol
```diff
@@ -33,14 +33,65 @@ contract RootManager is ProposedOwnable, IRootManager, WatcherClient, DomainInde
...
 
   event PropagateFailed(uint32 domain, address connector);
 
+  event SlowModeActivated();
+
+  event OptimisticModeActivated();
+
+  event OptimisticRootPropagated(bytes32 aggregateRoot, bytes32 domainsHash);
+
+  event AggregateRootProposed(
```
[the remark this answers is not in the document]
gotzenx: What you you mean by: `can we rearrange the params in this event to support that?` ?

packages/deployments/contracts/contracts/messaging/RootManager.sol
```diff
@@ -187,22 +386,97 @@ contract RootManager is ProposedOwnable, IRootManager, WatcherClient, DomainInde
     address[] calldata _connectors,
     uint256[] calldata _fees,
     bytes[] memory _encodedData
-  ) external payable whenNotPaused {
+  ) public payable whenNotPaused {
     validateConnectors(_connectors);
 
     uint256 _numDomains = _connectors.length;
     // Sanity check: fees and encodedData lengths matches connectors length.
     require(_fees.length == _numDomains && _encodedData.length == _numDomains, "invalid lengths");
```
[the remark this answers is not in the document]
gotzenx: Yeah def, i would suggest to first work on all the required changes about the code we added, and once thats approved, we can mark that as done and make the refactor a separated task

packages/deployments/contracts/contracts/messaging/connectors/SpokeConnector.sol
gotzenx: Sorry what comment are you referring to?

packages/deployments/contracts/contracts/messaging/connectors/SpokeConnector.sol
gotzenx: Will do

packages/deployments/contracts/contracts/messaging/RootManager.sol
```diff
@@ -33,14 +33,65 @@ contract RootManager is ProposedOwnable, IRootManager, WatcherClient, DomainInde
...
+  /**
+   * @notice The time watchers have to detect and invalidate the proposed root.
+   */
+  uint256 public constant DISPUTE_TIME = 30 minutes;
+
+  /**
+   * @notice Duration of the snapshot
+   */
+  uint256 public constant SNAPSHOT_DURATION = 30 minutes;
+
```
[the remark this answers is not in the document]
gotzenx: We talked with @skeletor-spaceman about this... We consider that leaving the SNAPSHOT_DURATION as timestamp is more intuitive and does not represents any danger to the system. Also easier and less risky to be set on every domain since will use the same value on all of them.

About changing the DISPUTE_TIME to be in blocks, we consider that is a good idea so we will implement that.

packages/deployments/contracts/contracts/messaging/RootManager.sol
```diff
@@ -173,10 +307,75 @@ contract RootManager is ProposedOwnable, IRootManager, WatcherClient, DomainInde
...
+   * @param _domains Array of all the domains
+   */
+  function proposeAggregateRoot(
+    uint256 _snapshotId,
+    bytes32 _aggregateRoot,
+    bytes32[] calldata _snapshotsRoots,
+    uint32[] calldata _domains
+  ) external onlyProposer onlyOptimisticMode checkDomains(_domains) {
+    if (_snapshotId != block.timestamp / SNAPSHOT_DURATION) revert RootManager_InvalidSnapshotId(_snapshotId);
+    if (proposedAggregateRoot.aggregateRoot > 0) revert RootManager_ProposeInProgress();
```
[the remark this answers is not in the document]
gotzenx: When comparing or manipulating `bytes32` you can use zero without the need of casting it to bytes32 previously. This saves little bit of gas and reduces the code needed.
It is ok if we do it that way?

packages/deployments/contracts/contracts/messaging/RootManager.sol
```diff
@@ -33,14 +33,65 @@ contract RootManager is ProposedOwnable, IRootManager, WatcherClient, DomainInde
...
+
   // ============ Properties ============
 
+  /**
+   * @notice The time watchers have to detect and invalidate the proposed root.
+   */
+  uint256 public constant DISPUTE_TIME = 30 minutes;
+
+  /**
+   * @notice Duration of the snapshot
```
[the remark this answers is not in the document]
gotzenx: What do you mean `with effective snapshot length`?

packages/deployments/contracts/contracts/messaging/RootManager.sol
```diff
@@ -53,6 +104,35 @@ contract RootManager is ProposedOwnable, IRootManager, WatcherClient, DomainInde
...
+  /**
+   * @notice The last aggregate root we propagated to spoke chains. Used to prevent sending redundant
+   * aggregate roots in `propagate`.
+   */
+  bytes32 public lastPropagatedRoot;
+
+  /**
+  @notice The last finalized optimistic aggregate root.
+   */
+  bytes32 public finalizedOptimisticAggregateRoot;
```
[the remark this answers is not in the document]
gotzenx: Yes its to be able to move a proposedRoot to the finalized status independently of propagating it or not.
This way if the system is not able to propagate for any reason we can still keep the flow of propose + validation going without the need of propagation.

packages/deployments/contracts/contracts/messaging/RootManager.sol
```diff
@@ -33,14 +33,65 @@ contract RootManager is ProposedOwnable, IRootManager, WatcherClient, DomainInde
 
   event RootDiscarded(bytes32 fraudulentRoot);
 
+  event ProposerAdded(address proposer);
+
+  event ProposerRemoved(address proposer);
```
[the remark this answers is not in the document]
ashitakah: 👀

packages/deployments/contracts/contracts/messaging/RootManager.sol
```diff
@@ -33,14 +33,65 @@ contract RootManager is ProposedOwnable, IRootManager, WatcherClient, DomainInde
...
 
   event ConnectorRemoved(uint32 domain, address connector, uint32[] domains, address[] connectors, address caller);
 
   event PropagateFailed(uint32 domain, address connector);
 
+  event SlowModeActivated();
+
+  event OptimisticModeActivated();
+
+  event OptimisticRootPropagated(bytes32 aggregateRoot, bytes32 domainsHash);
```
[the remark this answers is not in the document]
ashitakah: 👀

packages/deployments/contracts/contracts/messaging/RootManager.sol
ashitakah: 👀

packages/deployments/contracts/contracts/messaging/connectors/SpokeConnector.sol
ashitakah: 👀

packages/deployments/contracts/contracts/messaging/RootManager.sol
```diff
@@ -63,24 +143,56 @@ contract RootManager is ProposedOwnable, IRootManager, WatcherClient, DomainInde
...
+   * @notice Struct to store the proposed data
+   * @dev The list of the snapshots roots and the domains must be in the
+   * same order as the roots insertions on the tree.
+   * @param endOfDispute The timestamp when the dispute period is over
+   * @param aggregateRoot The new aggregate root
+   */
+  struct ProposedData {
+    uint256 endOfDispute;
+    bytes32 aggregateRoot;
+  }
```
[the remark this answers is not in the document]
gotzenx: We are currently analyzing a possible solution for this that involves reducing gas a lot by keeping the slot warm and still be able to save both variables for data availability on-chain.

packages/deployments/contracts/contracts/messaging/RootManager.sol
gotzenx: Should it be indexed too @LayneHaber ?

packages/deployments/contracts/contracts/messaging/connectors/SpokeConnector.sol
LayneHaber: woah - odd. looks like I deleted the wrong comment

the question is what about the duration being in blocks -- doesn't actually work with the snapshot duration since that's used to generate the ID, but we generally use blocks for dispute timers. we have seen chains halt, which is problematic if you are measuring dispute windows in time since the dispute window will elapse without you ever actually having the opportunity to submit.

packages/deployments/contracts/contracts/messaging/RootManager.sol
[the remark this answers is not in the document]
LayneHaber: only the first 3 params emitted in an event can be indexed, so to index those roots / ids we would have to change the event signature slightly

packages/deployments/contracts/contracts/messaging/RootManager.sol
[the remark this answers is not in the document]
LayneHaber: the snapshot length that is perceived. so if I skip every other snapshot, the effective duration is one hour (so this is the smallest unit of snapshot duration, not just the only option)

packages/deployments/contracts/contracts/messaging/RootManager.sol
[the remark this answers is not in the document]
LayneHaber: lol I think this is the comment that I thought was "above"

agree with the assessment of both!

packages/deployments/contracts/contracts/messaging/RootManager.sol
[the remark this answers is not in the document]
LayneHaber: interesting, what's the high-level idea?

packages/deployments/contracts/contracts/messaging/RootManager.sol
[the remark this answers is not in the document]
LayneHaber: yep!

packages/deployments/contracts/contracts/messaging/RootManager.sol
LayneHaber: yeah you know I love me an opinionated event topic lets do it

packages/deployments/contracts/contracts/messaging/RootManager.sol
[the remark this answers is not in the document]
gotzenx: Indexed params dont need to be the first 3 of the event. There's a max of 3 indexed params per event, but the order should not matter, and indexed or not it should not modify the event signature.

packages/deployments/contracts/contracts/messaging/connectors/SpokeConnector.sol
gotzenx: This is related to the comment about changing the DISPUTE_TIME to blocks so its already addressed.

packages/deployments/contracts/contracts/messaging/libraries/SnapshotId.sol
```diff
@@ -0,0 +1,33 @@
...
+    }
+  }
+
+  /**
+   * @notice Get the snapshot duration
+   * @return _snapshotDuration The snapshot duration
+   */
+  function getSnapshotDuration() internal pure returns (uint256 _snapshotDuration) {
+    _snapshotDuration = SNAPSHOT_DURATION;
+  }
```
LayneHaber, line 32: do you need a getter if its a public constant?

packages/deployments/contracts/contracts/messaging/RootManager.sol
```diff
@@ -33,26 +34,146 @@ contract RootManager is ProposedOwnable, IRootManager, WatcherClient, DomainInde
...
+
+  error RootManager_finalize__InvalidInputHash();
+
   // ============ Properties ============
 
+  /**
+   * @notice The number of blocks watchers have to detect and invalidate the proposed root.
+   * @dev    150 blocks is equivalent to 30 minutes in the current Ethereum Proof of Stake.
+   */
+  uint256 public constant DISPUTE_BLOCKS = 150;
```
LayneHaber, line 129: since we will likely want additional hub chains in the future, this should be set via the constructor

LayneHaber (commented):
outside of those two small comments it lgtm!

packages/deployments/contracts/contracts/messaging/RootManager.sol
gotzenx: done.  We also added a setter and introduced a new variables for minDisputeBlock to guard fat-finger problems

LayneHaber (approved):

PR #4061, opened 2023-04-25
Relayer proxy new functions

rhlsthrm:
## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes


packages/deployments/contracts/contracts/core/connext/helpers/RelayerProxyHub.sol
```diff
@@ -139,21 +154,59 @@ contract RelayerProxyHub is RelayerProxy {
...
+  /**
+   * @notice Indicates if the job is workable by the sender. Takes into account the Autonolas priority.
+   * For example, if priority is 3, then sender will not be able to work on blocks 0, 1, 2, unless they are Autonolas.
+   * Priority 0 disables Autonolas priority completely."
+   * @param _sender The address of the caller
+   */
+  modifier isWorkableBySender(AutonolasPriorityFunction _function, address _sender) {
+    if (
+      _sender != autonolas && autonolasPriority[_function] != 0 && block.number % 10 <= autonolasPriority[_function] - 1
+    ) {
```
LayneHaber, line 204: I feel like the `_sender` condition is wrong here -- the rest of the conditions are relevant IFF its autonolas

LayneHaber: but im not sure I understand why some of these functions are restricted

LayneHaber (commented):
no tests :(

packages/deployments/contracts/contracts/core/connext/helpers/RelayerProxy.sol
```diff
@@ -46,12 +46,8 @@ contract RelayerProxy is ProposedOwnable, ReentrancyGuard, GelatoRelayFeeCollect
   ISpokeConnector public spokeConnector;
   // special consideration for Autonolas keeper
   address public autonolas;
```
LayneHaber, line 48: do we want to keep this when everything else is removed?

packages/deployments/contracts/contracts/core/connext/helpers/RelayerProxy.sol
```diff
@@ -162,13 +145,6 @@ contract RelayerProxy is ProposedOwnable, ReentrancyGuard, GelatoRelayFeeCollect
    */
   event AutonolasChanged(address updated, address previous);
```
LayneHaber, line 146: do we want to include this?

packages/deployments/contracts/contracts/core/connext/helpers/RelayerProxy.sol
```diff
@@ -202,7 +178,6 @@ contract RelayerProxy is ProposedOwnable, ReentrancyGuard, GelatoRelayFeeCollect
     _setFeeCollector(_feeCollector);
     _setKeep3r(_keep3r);
     _setAutonolas(_autonolas);
-    _setAutonolasPriority(_autonolasPriority);
```
LayneHaber, line 205: remove `_autonolasPriority` from constructor

packages/deployments/contracts/contracts/core/connext/helpers/RelayerProxy.sol
LayneHaber: if we are keeping autonolas on the `RelayerProxy` so it is usable on spokes, wouldn't we want to keep all of the helpers in the same contract? im definitely missing some context on what this is and why its being moved around right now

LayneHaber (changes requested):

LayneHaber:
added a signature and removed references to autonolas. If autonolas is desired in the future, we can return by reverting [55f3670](https://github.com/connext/monorepo/pull/4061/commits/55f36706053fc8b31ff9b93df3b2ba218cd96755)

LayneHaber (approved):

Issue #4074, opened 2023-04-27
[Title] What would you like us to build or improve? 


prathmeshkhandelwal1:
## Problem
Adding readme for chain-abstraction package

## Impact
_Why does this matter? Who does it impact? How much does it impact them? What is the urgency?_

## Proposed Solution
_[OPTIONAL] Thoughts on solution design. We could do <insert idea> to solve this problem._

## Acceptance Criteria
_From an end user perspective, the below criteria must be met to consider this done:_
- [ ]

PR #4075, opened 2023-04-27
feat: added readme for chain-abstraction

prathmeshkhandelwal1:
#4074 Fixed


sanchaymittal:
I think we should directly add it to docs, instead of ReadMe. Better to maintain and keep docs to latest.
cc: @rhlsthrm

rhlsthrm:
Agree with @sanchaymittal , the readme should contain instructions for running code locally, testing, etc. And then point to the docs for usage examples.

prathmeshkhandelwal1:
> Agree with @sanchaymittal , the readme should contain instructions for running code locally, testing, etc. And then point to the docs for usage examples.

Sure, Changing it accordingly

prathmeshkhandelwal1:
Suggested changes added @rhlsthrm 
cc: @alexwhte

rhlsthrm (approved):

PR #4079, opened 2023-04-28
Watcher monitor

gotzenx:
## Description
Watcher is able to monitor, verify the optimistic proposal and put the system in Slow Mode when needed.
- Add the ability to have different invariant checks with different intervals
- Monitor optimistic proposal
- Verify if proposal is valid or not generating a virtual `aggregateRoot` to compare with proposed one
- Switch the system to Slow Mode in case the proposal has an invalid `aggregateRoot`

## Type of change

- [x] Refactoring


packages/adapters/watcher/src/switch/switcher.ts
```diff
@@ -0,0 +1,117 @@
...
+        gasPrice: tx.gasPrice!.toString(),
+        price: gasPrice.toString(),
+      });
+
+      // 5. Send tx
+      receipt = await txservice.sendTx(tx, requestContext);
+      return {
+        domain: hubDomain,
+        switched: !!receipt.status,
+        error: !receipt.status ? "Transaction reverted" : "",
```
LayneHaber, line 74: Can we decode a better error showing why the transaction reverted here?

packages/adapters/watcher/src/switch/switcher.ts
```diff
@@ -0,0 +1,117 @@
...
+        relevantTransaction: receipt.transactionHash,
```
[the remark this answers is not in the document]
LayneHaber: in this case it would be the most recent proposal

packages/adapters/watcher/src/types.ts
```diff
@@ -41,8 +53,51 @@ export abstract class Verifier {
...
+      throw new Error(`SpokeConnector deployment not found for chain ${chainId}!`);
+    }
+    return spokeConnector;
+  }
 }
 
+export type SwitchResponse = {
+  domain: number;
+  switched: boolean;
+  error: any;
```
LayneHaber, line 97: the error should have a stronger typing here -- seems that its being used as `error: string | Error`

packages/adapters/watcher/src/types.ts
```diff
@@ -41,8 +53,51 @@ export abstract class Verifier {
...
+  relevantTransaction: ethers.providers.TransactionResponse | string;
```
LayneHaber, line 98: should be a transaction receipt right? if sent using the `chainservice`, it would be a receipt as the bumping / monitoring is handled internally to that class

packages/adapters/watcher/src/verifiers/proposal.ts
```diff
@@ -0,0 +1,249 @@
...
+
+import { ProposedData, Verifier, VerifierContext, VerifyResponse } from "../types";
+import { virtualTreeBytecode } from "../bytecodes";
+import { EMPTY_ROOT } from "../helpers/constants";
+
+export class ProposedRootVerifier extends Verifier {
+  constructor(
+    context: VerifierContext,
+    private readonly hubDomain: number,
+    private readonly hubProvider: providers.JsonRpcProvider,
```
LayneHaber, line 12: why not use the `chainreader` here?

packages/adapters/watcher/src/verifiers/proposal.ts
```diff
@@ -0,0 +1,249 @@
...
+    const { aggregateRoot: proposedRoot } = proposedData;
+
+    // stop check execution if the proposed root dispute period has elapsed
+    const isDisputeTimeOver = await this.checkDisputeTimeFinalization(+proposedData.endOfDispute);
+    if (isDisputeTimeOver) {
+      return {
+        needsAction: false,
+        reason: `Dispute time for root ${proposedRoot} is already over`,
+      };
+    }
```
LayneHaber, line 41: if the dispute time elapses, and there is a valid reason to error on the proposal, we should likely make sure that is handled with at least some explicit logging if not alerts

packages/adapters/watcher/src/verifiers/proposal.ts
```diff
@@ -0,0 +1,249 @@
...
+  ): Promise<ProposedData> {
+    const { methodContext } = createLoggingContext(this.getProposedData.name);
+
+    // It's important to ensure the provider doesn't have a block-limit to fetch block from, otherwise we must add a contraint
+    // in the toBlock and fromBlock optional parameters of rootManagerInstance.queryFilter
+
+    try {
+      // create a filter to get the latest proposed root event
+      const eventFilter = rootManager.filters.AggregateRootProposed();
+      const events = await rootManager.queryFilter(eventFilter);
```
LayneHaber, line 82: this will need some block boundaries otherwise some providers will error (and most all will be slow)

packages/adapters/watcher/src/verifiers/proposal.ts
```diff
@@ -0,0 +1,249 @@
...
+
+      // sort events by block number to get the latest proposed root
+      const [latestProposedRootEvent] = events.sort((a, b) => b.blockNumber - a.blockNumber);
+
+      if (!latestProposedRootEvent?.args) {
+        throw new Error(`No valid events found for root proposal`);
+      }
```
LayneHaber, line 89: when would this happen? I suppose if `events.length == 0`

LayneHaber: is that the only case? should the error be more indicative of that?

packages/adapters/watcher/src/verifiers/proposal.ts
```diff
@@ -0,0 +1,249 @@
...
+        const spokeConnectorInterface = new utils.Interface(spokeConnector.abi as string[]);
+        let snapshotRoot = await this.getSnapshotRootForDomain(
+          proposedData.snapshotId,
+          spokeConnector.address,
+          spokeConnectorInterface,
+          domain,
+        );
+
+        // If the snapshot root is empty, we need to get the root from `getOutboundRoot` method
+        if (snapshotRoot == EMPTY_ROOT) {
```
LayneHaber, line 132: `snapshotRoot == constants.HashZero` ?

packages/adapters/watcher/src/verifiers/proposal.ts
```diff
@@ -0,0 +1,249 @@
...
+  }
+
+  /**
+   * @notice Gets the snapshot root for a given snapshotId in a given domain
+   * @param snapshotId - The id of the snapshot root we are going to fetch
+   * @param spokeConnectorAddress - The address of the spoke connector in the given domain
+   * @param spokeConnectorInterface - The interface of the spokeConnector
+   * @return snapshotRoot The snapshot root for the given snapshotId
+   */
+  private async getSnapshotRootForDomain(
```
LayneHaber, line 187: should we handle the `if empty, fetch from merkle` edgecase here?

packages/adapters/watcher/src/verifiers/proposal.ts
```diff
@@ -0,0 +1,249 @@
...
+
+    return snapshotRoots;
+  }
+
+  /**
+   * @notice Uses the batching method to get the latest aggregate root from the MerkleTreeManager and inserts each snapshot root of each domain to get the correct virtual aggregate root
+   * @param leaves - Array of the snapshoot roots of each domain. Its order must follow the order of the domains emitted in the AggregateRootProposed event.
+   * @return virtualRoot The virtual aggregate root computed after the insertion of the leaves
+   */
+  private async insertVirtualLeaves(leaves: string[]): Promise<string> {
```
LayneHaber, line 149: is there anyway this could get out of sync with what is maintained off-chain since the root merkle tree doesn't get updated each time a new set of leaves is proposed?

packages/agents/watcher/src/operations/validateAndSwitch.ts
```diff
@@ -1,49 +1,44 @@
...
     adapters: { monitor },
     logger,
     config,
   } = getContext();
   const methodContext = createMethodContext(switchAndAlert.name);
 
-  const domains = Object.keys(config.chains);
-  logger.warn("SWITCHING TO SLOW MODE!!!", requestContext, methodContext, { reason, transactions });
+  const hubDomain = config.hubDomain;
+  logger.warn("SWITCHING TO SLOW MODE!!!", requestContext, methodContext, { reason, transactions: undefined });
```
LayneHaber, line 27: the `transactions` field will just be dropped here when its translated to logging services, and its not providing any particularly valuable information, so we should just drop it

packages/agents/watcher/src/operations/validateAndSwitch.ts
```diff
@@ -1,49 +1,44 @@
...
-      // reason,
-      // timestamp: Date.now(),
-      // event: ReportEventType.Switch,
-      // logger,
-      // relevantTransactions: result.map((p) => p.relevantTransaction),
-      // requestContext,
-      // rpcs: Object.entries(config.chains)
-      //   .map((chain) => chain[1].providers)
-      //   .flat(),
+      //TODO: Check with connext if alerter approach is correct, and modify to only pass hubDomain instead of domains
```
LayneHaber, line 32: yes this is correct -- this is used to log the domains action was taken on in the report

LayneHaber (commented):
we likely will not merge this into the main branch and restructure how the watcher is built and distributed

preethamr:
Not valid anymore.

PR #4102, opened 2023-05-04
feat: do not cache the sdks 

0xksure:
## Description

As of now it's not possible to initialize two different clients inside one runtime. This is due to the way the sdks are initialized. Instead of caching the sdk instances it's possible to create a new one each time. This would incur some overhead for projects initializing the client inside their project. However, projects should be responsible for caching sdk through standard patterns like the usage of context.   

The changes are mainly removing the `this._instance` check in the sdk initialization and writing a test that checks that it's possible to create two different clients in the same runtime/context. 

## Type of change

- [ ] Refactoring

Users might incur a loss in loading times if they initializes the sdk many places in their project. However, initializing an sdk multiple times can be viewed as an anti pattern. Instead, it might make more sense to pass a context around that contains the initialized sdk.   

## Related Issue(s)

We at LIFI wants to present mainnet and testnet transfers in the same context. This is not possible ATM due to the caching of sdks.


sanchaymittal (approved):
looking good!!

just-a-node (approved):

PR #4131, opened 2023-05-07
Db read replica

just-a-node:
## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes


rhlsthrm (approved):

ops/mainnet/prod/backend/main.tf
```diff
@@ -1,7 +1,7 @@
 terraform {
   backend "s3" {
     bucket = "nxtp-terraform-mainnet-prod-backend"
-    key    = "state/"
```
carlomazzaferro, line 4: This needs to be reverted immediately, otherwise the state will be migrated

ops/mainnet/prod/core/main.tf
```diff
@@ -1,7 +1,7 @@
 terraform {
   backend "s3" {
     bucket = "nxtp-terraform-mainnet-prod-core"
-    key    = "state/"
```
carlomazzaferro, line 4: same

ops/infra/main.tf
```diff
@@ -1,7 +1,7 @@
 terraform {
   backend "s3" {
     bucket = "nxtp-terraform-infra"
-    key    = "state/"
```
carlomazzaferro, line 4: same

carlomazzaferro:
Any reason why we're splitting the modules between `db` and `db-replica`? For sake of DRY, the best option would likely be to just pass a variable `replicate_db` as a `true` or `false` and then the replica would be created based on this variable.

ops/mainnet/prod/backend/main.tf
just-a-node: Was getting an error "key must not end with '/'" running `terraform init`. Reverting as this will cause the migrate.

PR #4199, opened 2023-05-11
4145 pg cron

### LINKED ISSUE #4145: Terraform setup for db read replica
## Problem

Setup terraform for DB read replica + Route53.
Add `pg_cron` job to refresh the new mat view.

just-a-node:
## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes #4145

### DIFF
--- a/.github/workflows/build-test-deploy.yml
+++ b/.github/workflows/build-test-deploy.yml
@@ -30,7 +30,7 @@ jobs:
       # Label used to access the service container
       postgres:
         # Docker Hub image
-        image: postgres
+        image: ghcr.io/connext/database:sha-fa66dca
         # Provide the password for postgres
         env:
           POSTGRES_PASSWORD: qwerty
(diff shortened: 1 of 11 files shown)


.github/workflows/build-test-deploy.yml
```diff
@@ -519,9 +519,8 @@ jobs:
     if: github.ref == 'refs/heads/main' || github.ref == 'refs/heads/staging' || github.ref == 'refs/heads/testnet-prod' || github.ref == 'refs/heads/prod'
     runs-on: ubuntu-latest
     env:
-      REGISTRY: 679752396206.dkr.ecr.us-east-1.amazonaws.com
-      IMAGE_TAG: ${{ github.sha }}
-      REPOSITORY: nxtp-cartographer
+      REGISTRY: ghcr.io
+      IMAGE_NAME: connext/sequencer-subscriber
```
carlomazzaferro, line 523: I fail to understand these changes here

ops/mainnet/prod/backend/main.tf
```diff
@@ -43,7 +43,7 @@ module "cartographer_db" {
     Domain      = var.domain
   }
 
-  parameter_group_name = "default.postgres14"
+  parameter_group_name = "rds-postgres"
```
carlomazzaferro, line 46: Hmmm, this needs to depend on the other module, otherwise if one gets created before the other, this might fail.

What you can do is get the output of the `aws_db_parameter_group` and then use it in `main.tf`

Something like
```
# ops/modules/db/outputs.tf

output "rds_parameter_group" {
   vault = aws_db_parameter_group.rds_postgres.name
}

# ops/mainnet/prod/main.tf

parameter_group_name = module.db.parameter_group_name
```

carlomazzaferro:
How is the DB docker image actually run? I fail to understand how / when it'd run. A high level overview would be helpful

ops/mainnet/prod/backend/main.tf
just-a-node: Roger that

.github/workflows/build-test-deploy.yml
just-a-node: @rhlsthrm Was this a typo? What is sequencer image doing under carto?

PR #4281, opened 2023-05-21
Scale LH Prover

preethamr:
## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes


packages/agents/lighthouse/src/config.ts
```diff
@@ -160,6 +162,8 @@ export const getEnvConfig = (
       configJson.relayerWaitTime ||
       configFile.relayerWaitTime ||
       DEFAULT_RELAYER_WAIT_TIME,
+    concurrency:
+      process.env.NXTP_PROVER_CONCURRENCY || configJson.concurrency || configFile.concurrency || DEFAULT_CONCURRENCY,
```
wanglonghong, line 166: In general, the data type of env variable is string but concurrency is number so need a fix here

wanglonghong (commented):
I am still skeptical about this solution how it could improve the messaging layer processing because as we know nodejs is a single threaded application. 
The one way I am thinking is to make lh prover pub/sub as we do in sequencer. but need others input here

rhlsthrm:
> I am still skeptical about this solution how it could improve the messaging layer processing because as we know nodejs is natively a single threaded application. The one way I am thinking is to make lh prover pub/sub as we do in sequencer. but need others input here

This would still work because even though Node is single threaded it will work on all the processes at the same time. So when it's waiting on one, it will process the other.

packages/agents/lighthouse/src/config.ts
rhlsthrm: I think this is correct.

preethamr: True. Pushed one way check and convert, perhaps there is a more elegant way to do this.

Interesting that we have this pattern at a few different places like  for `relayerWaitTime` , `NXTP_CARTOGRAPHER_POLL_INTERVAL`. 
We should update this in all places right @rhlsthrm @wanglonghong  ?

wanglonghong: Yes right. we have similar things in other places.

rhlsthrm: Correct I think it's working because we are using the defaults, so we need this if we override it.

wanglonghong (approved):

PR #4363, opened 2023-05-30
feat: 4326 scale sequencer http server

preethamr:
## Description
4326 scale sequencer http server

highlights:
- http server is a separate entity, just validate data and queue, nothing else
- a new queue `http` is introduced, as task queue for the "http subscriber"
- the current workload of `publisher` is split between `server` and `publisher`, with this queue in between
- tests
- terraform
- docker
- docker compose
- setup/integrations scripts
- seq MQ config changes, increase max limit to 100K
- fix bugs from changing client lib from `foofoomq` to `amqplib` (possible there are other regression bugs)

## Type of change

- [x] Refactoring

## Related Issue(s)

Fixes


rhlsthrm (dismissed):

sanchaymittal (approved):
reviewed on high level, lg!

packages/agents/sequencer/src/bindings/server/index.ts
```diff
@@ -111,14 +112,29 @@ export const bindServer = async (): Promise<FastifyInstance> => {
       try {
         const bid = request.body;
         requestContext.transferId = bid.transferId;
-        await storeFastPathData(bid, requestContext);
+        // Validate Input schema
+        const validateInput = ajv.compile(BidSchema);
```
rhlsthrm, line 116: There's actually an easier way to do this schema validation directly through Fastify: https://www.fastify.io/docs/latest/Reference/Validation-and-Serialization/

We are doing this in other HTTP handlers you can look at.

packages/agents/sequencer/src/bindings/server/index.ts
```diff
@@ -111,14 +112,29 @@ export const bindServer = async (): Promise<FastifyInstance> => {
...
+        const validInput = validateInput(bid);
+        if (!validInput) {
+          const msg = validateInput.errors?.map((err: any) => `${err.instancePath} - ${err.message}`).join(",");
+          throw new DataInvalid({
```
rhlsthrm, line 120: Schema validation errors should not throw 500, I think using the validation strategy above will take care of that.

rhlsthrm (changes requested):
Looks great! Minor change requested.

packages/agents/sequencer/src/bindings/server/index.ts
preethamr: Makes sense. Done

PR #4398, opened 2023-06-01
Fee estimate server side cache

### LINKED ISSUE #4223: Host sdk-server on AWS
## Problem
_What problem are we solving?_ 

Our current SDK is heavy and is not possible to use in some front-end applications. If we deploy a hosted SDK server that can generate the appropriate transactions data then we can have a highly optimized front end SDK that simply queries this server.

## Impact
_Why does this matter? Who does it impact? How much does it impact them? What is the urgency?_

We have anecdotal evidence that users are churning due to our SDK bundle size.

## Proposed Solution
_[OPTIONAL] Thoughts on solution design. We could do <insert idea> to solve this problem._

### App Layer
* Containerize [sdk-server package](https://github.com/connext/monorepo/tree/main/packages/examples/sdk-server)
* Host on ECS to expose HTTP endpoints
* Set up autoscaling groups in multiple regions
* [Optional] Configure multiple availability zones per region

### Cache Layer
* Set up ElastiCache (Redis) clusters in multiple regions
* [Optional] Enable cross-region replication (probably not necessary)
* [Optional] Configure multiple availability zones per region
* Implement middleware in `sdk-server` to cache requests in Redis

### Load Balancer
* Set up ALB to distribute traffic across ECS instances in each region
* One ALB per region

### DNS
* Geolocation routing of requests to different regional ALBs with Route 53

### [Optional] API Gateway
* First-class support for OpenAPI/Swagger specification
* Generate SDKs for multiple platforms using the OpenAPI spec (these are just client SDKs that help with hitting the API)
* Anti-DDoS vectors using rate limiting, throttling, API keys

### [Optional] Analytics
* This depends on what we want to capture. If we set up API Gateway, we can get metrics like API call count, latency, cache hits/misses. This should be possible to collect from logs without API Gateway as well but slightly more difficult.

### Other Notes
* Use Terraform to define all the infrastructure pieces
* Replicate for each environment (`testnet-staging`, `testnet-prod`, `mainnet`)

## Acceptance Criteria
_From an end user perspective, the below criteria must be met to consider this done:_
- [ ] SDK Server is hosted and publicly accessible
- [ ] API requests have low latency from multiple regions
- [ ] Can handle scaling requirements (TODO)

just-a-node:
## Description

Minimal changes to enable optional caching on the `estimateRelayerFee` route. This can be extended to other routes if necessary. Opted not to intercept `axiosGet` requests in the Gelato calls as this requires changes in the `utils` package and would make removing the server dependency more complex later. This should suffice for reducing latency since most client calls will not use the optional params in `SdkEstimateRelayerFeeParams` and will hit the cache for subsequent requests to the same `originDomain`/`destinationDomain`.

## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes #4223


sanchaymittal:
i think we should do that cache at the sdk code only, instead of adding it to server. Server should be purely a wrapper around the sdk.
@just-a-node

sanchaymittal (changes requested):
best to keep cache at core-sdk, and keeping server wrapper to main functions

just-a-node:
@sanchaymittal Wait, `sdk-server` is what's being hosted. We're going to have a separate wrapper package that mimics the client SDK interface and calls these server routes. So this cache is at the API side

sanchaymittal:
From what I can recall from the last discussion, Is that we want to release under package @connext/sdk only with different versioning.  So ideally we want to have cache being optional at core-sdk itself and wrapper at Server. And sdk turned into wrapper of api calls of server.

just-a-node:
I think we have a similar understanding, but see structure in https://github.com/connext/monorepo/issues/4228. 

What I was assuming is:
- core SDK stays the same
- sdk server stays basically the same, now it's hosted
- new package wraps api calls to the server and it's published to `@connext/sdk`

just-a-node:
The cache implemented here is the middleware described in https://github.com/connext/monorepo/issues/4223. This is caching at the API layer, not the service level of core sdk

sanchaymittal (approved):

rhlsthrm (dismissed):
Nice! I guess we'd need ops setup for this.

just-a-node:
Yes for redis and the right configs to feed server

rhlsthrm (approved):

sanchaymittal (approved):

PR #4402, opened 2023-06-03
Wrapper sdk

### LINKED ISSUE #4228: Wrapper SDK-Server with new versioning
## Problem

SDK pains:
- Large bundle size
- Compatibility issues with UI frameworks (CRA)
- Additional package downloading issues (what are some examples?)

## Impact

We have a few integrations which are blocked by the above and future integrations will experience friction without changes.

### SDK Wrapper Package 

*MUST:*
- Be published as a new version to `@connext/sdk`
- Match *exactly* the current function signatures and be backwards-compatible

*SHOULD:*
- Call the hosted `sdk-server` routes
- Include a minimal subset of core functions (`xcall`, `estimateRelayerFee`, `bumpTransfer`, `calculateAmountReceived`, `updateSlippage`)

### SDK Test Harness

*MUST:*
- Include unit tests with coverage abiding by the [global monorepo thresholds](https://github.com/connext/monorepo/blob/8ad4c909df849bec4b1122b442ec1512f6d8298a/.nycrc.json) such that this package's tests can be included in all CI runs

*SHOULD:*
- Include integration tests with a local `sdk-server` that validates correctness of routes and error response codes (note: might be combined with unit tests since this wrapper SDK will be a thin layer on top of the server routes)

### Package Structure

*SHOULD:*

 Involve the following:
    - `packages/agents/sdk`: existing sdk core classes
    - `packages/agents/sdk-wrapper`: package with wrapper classes exposed for clients to use, published to `@connext/sdk` under a new version
    - `packages/examples/server`: sdk-server routes that will be hosted on AWS

### Transition Strategy

*MUST:*
- Be seamless for clients during the transition to the wrapped SDK version
- Be seamless for clients during the transition from the wrapped SDK version to the original architecture of client-side-only operations
- Be able to eliminate dependency on the hosted server at some point

### Hosted SDK Server
- Tasking tracked separately in #4223
- Exposes API routes for all core SDK functions
- Will include rate limiting and analytics

## Acceptance Criteria
- [ ] Clients installing `@connext/sdk@<new_version>` will not need to change any existing implementations
- [ ] Latency for calls should fall under 1000ms (this will depend on server caching, geo, and scaling)

just-a-node:
## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes #4228


packages/agents/sdk-wrapper/src/sdkUtils.ts
```diff
@@ -0,0 +1,59 @@
...
+  async getTransfers(params?: {
+    userAddress?: string;
+    routerAddress?: string;
+    status?: XTransferStatus;
+    errorStatus?: XTransferErrorStatus;
+    transferId?: string;
+    transactionHash?: string;
+    xcallCaller?: string;
+    range?: { limit?: number; offset?: number };
+  }): Promise<any> {
```
rhlsthrm, line 45: Don't use Promise<any>. Just merged this PR which fixes it in main: https://github.com/connext/monorepo/pull/4401

packages/examples/bridge-reference/src/components/Header.tsx
```diff
@@ -1,6 +1,6 @@
 import React, { useEffect } from "react";
-import { create, SdkConfig } from "@connext/sdk";
-import { ChainConfig } from "@connext/sdk/dist/config";
+import { create, SdkConfig } from "@connext/sdk-core";
+import { ChainConfig } from "@connext/sdk-core/dist/config";
```
rhlsthrm, line 3: Why is this imported from the dist/? We should make sure this is a proper export if we need it imported in other places.

rhlsthrm (changes requested):
Looks good! Just a couple fixes needed.

packages/agents/sdk-wrapper/src/sdkUtils.ts
just-a-node: Pulling the type into `interfaces/index.ts` as all the other types are defined there.

rhlsthrm (approved):

sanchaymittal (approved):
approving, didn't really go through all the files. Recommeded to test rigorously locally first before releasing.

rhlsthrm (changes requested):
All the tests need to have proper assertions to assert we are calling the HTTP function with correct params. Otherwise it's useless.

rhlsthrm (dismissed):

rhlsthrm (approved):

preethamr (approved):

PR #4452, opened 2023-06-09
3437 checkpoints for carto at scale

liu-zhipeng:
## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes


packages/agents/cartographer/poller/src/lib/operations/transfers.ts
```diff
@@ -12,23 +12,23 @@ import {
 import { getContext } from "../../shared";
 
 const getMaxNonce = (transfers: DestinationTransfer[] | XTransfer[]): number => {
-  return transfers.length == 0 ? 0 : Math.max(...transfers.map((transfer) => transfer.xparams.nonce ?? 0));
+  return transfers.length == 0 ? 0 : Math.max(...transfers.map((transfer) => transfer.origin?.xcall.txNonce ?? 0));
```
rhlsthrm, line 15: Since xcall already has a nonce I don't think we need this, right?

liu-zhipeng: Yes, right. but I was going to process them consistently

rhlsthrm: Right. But having nonce already makes this unnecessary, so I think it's better to just remove it.

rhlsthrm (changes requested):
Remove the xcall processing.

rhlsthrm (dismissed):

preethamr (dismissed):

wanglonghong (approved):

rhlsthrm (approved):

PR #4493, opened 2023-06-13
fix: no need to increase offset because the query uses both startIndex and endIndex

wanglonghong:
## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes


wanglonghong:
someone's review is definitely needed here

rhlsthrm:
So this will get all the messages starting from the `cachedNonce + 1`. Seems reasonable if we aren't paginating.

preethamr (approved):

PR #4548, opened 2023-06-19
feat: 4538 axelar connector

liu-zhipeng:
## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes


rhlsthrm (approved):

rhlsthrm:
Add unit tests. Also build failing.

rhlsthrm (changes requested):
Need unit tests.

packages/deployments/contracts/deploy/01_deployMessaging.ts
```diff
@@ -286,8 +287,9 @@ const handleDeploySpoke = async (
   const merkleContract = await hre.ethers.getContractAt("MerkleTreeManager", merkleTreeManager.address, deployer);
   console.log("merkleContract: ", merkleContract.address);
 
-  console.log("await merkleContract.arborist(): ", await merkleContract.arborist());
-  if (!(await merkleContract.arborist())) {
+  const arborist = await merkleContract.arborist();
+  console.log("arborst for spoke merkle ", arborist);
+  if (arborist == constants.AddressZero) {
```
LayneHaber, line 292: we should remove this to ensure configuration only happens in the `initialize` scripts and not the `deploy` ones

packages/deployments/contracts/deploy/01_deployMessaging.ts
```diff
@@ -219,6 +219,7 @@ const handleDeploySpoke = async (
       !contract.includes("PolygonZk") &&
       !contract.includes("ZkSync") &&
       !contract.includes("Consensys") &&
+      !contract.includes("Axelar") &&
       !contract.includes("Multichain")) ||
```
LayneHaber, line 223: we can remove? or are we leaving the implementation in place.

imo we should delete

packages/deployments/contracts/contracts/messaging/connectors/axelar/BaseAxelar.sol
```diff
@@ -0,0 +1,75 @@
...
+import {IAxelarGateway} from "@axelar-network/axelar-gmp-sdk-solidity/contracts/interfaces/IAxelarGateway.sol";
+import {IAxelarGasService} from "@axelar-network/axelar-gmp-sdk-solidity/contracts/interfaces/IAxelarGasService.sol";
+import {StringToAddress, AddressToString} from "@axelar-network/axelar-gmp-sdk-solidity/contracts/utils/AddressString.sol";
+
+abstract contract BaseAxelar is AxelarExecutable {
+  using StringToAddress for string;
+  using AddressToString for address;
+
+  // ============ Internal Storage ============
+  IAxelarGasService public immutable gasService;
```
LayneHaber, line 14: will the address ever change?

packages/deployments/contracts/contracts/messaging/connectors/axelar/BaseAxelar.sol
```diff
@@ -0,0 +1,75 @@
...
+
+  // Mirror chain id
+  string public MIRROR_CHAIN_ID;
```
LayneHaber, line 17: this should be immutable

packages/deployments/contracts/contracts/messaging/connectors/axelar/BaseAxelar.sol
```diff
@@ -0,0 +1,75 @@
...
+
+    // set immutable propertioes
+    gasService = IAxelarGasService(_gasReceiver);
+    MIRROR_CHAIN_ID = _mirrorChainId;
+  }
+
+  // ============ Public Fns ============
+  function _execute(
+    string calldata _sourceChain,
+    string calldata _sourceAddress,
```
LayneHaber, line 33: should `_verifySender` be implemented at this layer if its the same to check the crosschain sender on both hub and spoke?

packages/deployments/contracts/contracts/messaging/connectors/axelar/BaseAxelar.sol
```diff
@@ -0,0 +1,75 @@
...
+
+    string memory mirrorConnectorStr = _mirrorConnector.toString();
+
+    // Get the min fees
+    gasService.payNativeGasForContractCall{value: msg.value}(
+      address(this),
+      MIRROR_CHAIN_ID,
+      mirrorConnectorStr,
+      _data,
+      msg.sender
```
LayneHaber, line 66: the refund address here will be the `RootManager`, which has no way to withdraw funds

LayneHaber: there's also no way to bump gas here, which may be needed?

also:
- this should only happen if `msg.value > 0`
- its not clear who excess fees should be refunded to to me here

packages/deployments/contracts/tasks/connector/sendAxelar.ts
```diff
@@ -0,0 +1,60 @@
...
+} from "../../src/utils";
+import { calculateAxelarBridgeFee, chainIdToDomain } from "@connext/nxtp-utils";
+
+type TaskArgs = {
+  data?: string;
+  env?: Env;
+  networkType?: ProtocolNetwork;
+};
+
+export default task("connector-send-axelar", "Call `Connector.send()` to distribute outbound root via axelar connector")
```
LayneHaber, line 20: this task should exist as a special case of `send-connector` based on chain, should not exist as a standalone task

packages/utils/src/axelar/bridge.ts
```diff
@@ -0,0 +1,52 @@
+import { AxelarQueryAPI, CHAINS, Environment } from "@axelar-network/axelarjs-sdk";
+
+import { domainToChainId, mainnetDomains } from "../helpers";
+
+export const chainIdToAxelarName: Map<number, string> = new Map([
```
LayneHaber, line 5: :(

packages/utils/src/helpers/domain.ts
```diff
@@ -41,6 +41,29 @@ export const chainIdToDomainMapping: Map<number, number> = new Map([
   [13338, 13338],
 ]);
 
+export const mainnetDomains: number[] = [
```
LayneHaber, line 44: I feel like this list has existed in a few forms -- can we make sure everything is using this? and since its a `const` it should really be `MAINNET_DOMAINS`

LayneHaber (changes requested):
missing lots of connector tests + subgraph changes.

in the future with these types of PRs where there was associated onchain testing, can we include a summary of that testing with associated links / methodology?

packages/deployments/contracts/deploy/01_deployMessaging.ts
liu-zhipeng: removed

packages/deployments/contracts/contracts/messaging/connectors/axelar/BaseAxelar.sol
liu-zhipeng: https://etherscan.io/address/0x2d5d7d31F671F86C782533cc367F14109a082712#code

GasService contract implemented using proxy, so I don't think it will not change

packages/deployments/contracts/contracts/messaging/connectors/axelar/BaseAxelar.sol
liu-zhipeng: string variable can't be immutable

packages/deployments/contracts/contracts/messaging/connectors/axelar/BaseAxelar.sol
liu-zhipeng: We should bum gas fee on their explorer manually, there is no way to do that onchain

packages/utils/src/axelar/bridge.ts
liu-zhipeng: couldn't find a better way to get `chainId - name` mapping from axelar SDK.

packages/deployments/contracts/contracts/messaging/connectors/axelar/BaseAxelar.sol
rhlsthrm: Should this be a string? I think we usually do unit32 for chain ID.

LayneHaber: > string variable can't be immutable

was afraid of that

> Should this be a string? I think we usually do unit32 for chain ID.

its not actually the chain id its the chain name

packages/deployments/contracts/contracts/messaging/connectors/axelar/BaseAxelar.sol
LayneHaber: isn't that passed into the function?

packages/deployments/contracts/contracts/messaging/connectors/axelar/BaseAxelar.sol
LayneHaber: you [can do it onchain](https://docs.axelar.dev/dev/general-message-passing/gas-services/increase-gas#addnativegas), but we can default to the explorer

LayneHaber: it will just be more annoying to track down

LayneHaber:
lint is running out of memory lol

packages/deployments/contracts/contracts/messaging/connectors/axelar/BaseAxelar.sol
rhlsthrm: I don't think we will ever be bumping gas because we will just wait for the next root to be sent in the next cycle. So I'm not sure this should be part of the functionality of this contract if we can also use the explorer.

For the refund address it should be passed in, it can be our safe, since we are the ones paying the fees.

rhlsthrm:
> lint is running out of memory lol

Yikes this is not good. I think it might be because of the types we added?

LayneHaber:
we will go wormhole for the BNB replacement, I think that means we can close this PR for now?

not sure if its easier to close or repurpose tbh

LayneHaber (changes requested):
see above comment re: wormhole

rhlsthrm:
> we will go wormhole for the BNB replacement, I think that means we can close this PR for now?
> 
> not sure if its easier to close or repurpose tbh

Let's close and open a new one for Wormhole.

PR #4557, opened 2023-06-20
fix: reset cache on verification fail

preethamr:
## Description
Reset cache on verification fail

## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes

### DIFF
--- a/packages/agents/lighthouse/src/tasks/prover/operations/consumer.ts
+++ b/packages/agents/lighthouse/src/tasks/prover/operations/consumer.ts
@@ -41,7 +41,7 @@ export const consume = async () => {
           channel.ack(message);
         } catch (err: unknown) {
           logger.error("Processing messaages failed", requestContext, methodContext, undefined, { err });
-          channel.reject(message);
+          channel.reject(message, false);
         }
       }
     },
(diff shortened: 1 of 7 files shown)


packages/agents/lighthouse/src/tasks/prover/operations/consumer.ts
```diff
@@ -41,7 +41,7 @@ export const consume = async () => {
           channel.ack(message);
         } catch (err: unknown) {
           logger.error("Processing messaages failed", requestContext, methodContext, undefined, { err });
-          channel.reject(message);
```
wanglonghong, line 44: I am just curious why `channel.reject(message, false)` has been removed before.

wanglonghong (approved):

wanglonghong (approved):

packages/agents/lighthouse/src/tasks/prover/operations/consumer.ts
preethamr: It was added.
Would it not be a better design to drop the message if execute failed ?
It would get retried again, later right ?

wanglonghong: Its a better design to drop the message if execute failed. If we do nothing in a failure case, it will reenqueued and the backlog get bigger.  It has been originally added here when I am doing refactors with amqplib so I was asking someone tried to remove it but why???  lol

PR #4562, opened 2023-06-21
feat: add confirmation blocks msg poller

preethamr:
## Description
Add confirmation blocks msg poller

## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes


ops/testnet/prod/core/config.tf
```diff
@@ -279,7 +279,8 @@ locals {
         providers = ["https://eth-goerli.g.alchemy.com/v2/${var.goerli_alchemy_key_0}"]
       }
       "9991" = {
-        providers = ["https://polygon-mumbai.g.alchemy.com/v2/${var.mumbai_alchemy_key_0}"]
+        providers     = ["https://polygon-mumbai.g.alchemy.com/v2/${var.mumbai_alchemy_key_0}"],
+        confirmations = 200
```
rhlsthrm, line 283: We should add these for all domains. We should use 1 confirmation for rollups, 200 is good for mumbai, and 10 for Goerli. Let's not rely on the default for any domain. Also should add these for mainnet proactively so we don't miss this.

rhlsthrm (changes requested):

ops/testnet/prod/core/config.tf
rhlsthrm: This needs to be in the cartographer config also.

rhlsthrm (changes requested):

sanchaymittal (approved):

PR #4569, opened 2023-06-21
feat: admin connector wip

rhlsthrm:
## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes


packages/deployments/contracts/contracts/messaging/connectors/admin/AdminHubConnector.sol
```diff
@@ -0,0 +1,40 @@
...
+
+    // Emit event
+    emit MessageProcessed(abi.encode(_data), msg.sender);
+  }
+
+  // ============ Private fns ============
+
+  function _verifySender(address _expected) internal view override returns (bool) {}
+
+  function _sendMessage(bytes memory _data, bytes memory _encodedData) internal override {}
```
LayneHaber, line 39: should include comments to indicate why its a no-op vs. reverting

packages/deployments/contracts/contracts/messaging/connectors/admin/AdminHubConnector.sol
```diff
@@ -0,0 +1,40 @@
...
+    uint32 _domain,
+    uint32 _mirrorDomain,
+    address _rootManager
+  ) ProposedOwnable() HubConnector(_domain, _mirrorDomain, address(0), _rootManager, address(0)) {
+    _setOwner(msg.sender);
+  }
+
+  // ============ Admin fns ============
+  function addSpokeRootToAggregate(bytes32 _data) external onlyOwner {
+    IRootManager(ROOT_MANAGER).aggregate(MIRROR_DOMAIN, bytes32(_data));
```
LayneHaber, line 29: Sanity check: root is not empty

LayneHaber: also nit: no need to double cast

packages/deployments/contracts/contracts/messaging/connectors/admin/AdminHubConnector.sol
```diff
@@ -0,0 +1,40 @@
...
+import {TypedMemView} from "../../../shared/libraries/TypedMemView.sol";
+
+import {IRootManager} from "../../interfaces/IRootManager.sol";
+import {IArbitrumInbox} from "../../interfaces/ambs/arbitrum/IArbitrumInbox.sol";
+import {IArbitrumOutbox} from "../../interfaces/ambs/arbitrum/IArbitrumOutbox.sol";
+import {IArbitrumRollup, Node} from "../../interfaces/ambs/arbitrum/IArbitrumRollup.sol";
+
+import {HubConnector} from "../HubConnector.sol";
+import {Connector} from "../Connector.sol";
+import {ProposedOwnable} from "../../../shared/ProposedOwnable.sol";
```
LayneHaber, line 15: many of these are not being used, should remove unused imports

packages/deployments/contracts/contracts/messaging/connectors/admin/AdminHubConnector.sol
```diff
@@ -0,0 +1,40 @@
...
+
+contract AdminHubConnector is ProposedOwnable, HubConnector {
```
LayneHaber, line 17: need to have some context into why and how this connector should used

packages/deployments/contracts/contracts_forge/messaging/connectors/admin/AdminHubConnector.t.sol
```diff
@@ -0,0 +1,171 @@
...
+import {AdminHubConnector} from "../../../../contracts/messaging/connectors/admin/AdminHubConnector.sol";
+import {RootManager} from "../../../../contracts/messaging/RootManager.sol";
+import {WatcherManager} from "../../../../contracts/messaging/WatcherManager.sol";
+import {MerkleTreeManager} from "../../../../contracts/messaging/MerkleTreeManager.sol";
+import {Message} from "../../../../contracts/messaging/libraries/Message.sol";
+import {RateLimited} from "../../../../contracts/messaging/libraries/RateLimited.sol";
+
+import "../../../utils/ForgeHelper.sol";
+
+contract AdminHubConnectorTest is ForgeHelper {
```
LayneHaber, line 15: if this is a fork test, you should put it in the `upgrade` directory

LayneHaber: or otherwise somehow omit from standard unit testing

packages/deployments/contracts/contracts_forge/messaging/connectors/admin/AdminHubConnector.t.sol
```diff
@@ -0,0 +1,171 @@
...
+      "https://eth-mainnet.g.alchemy.com/v2/CAtB_bTRwPneOAh_xLABggG3yUOR2wsV",
+      // 17528353
+      17464115
+    );
+    adminHubConnector = new AdminHubConnector(ETH_DOMAIN, BNB_DOMAIN, address(rootManager));
+  }
+
+  // ============ utils ============
+
+  function test_AdminHubConnector__connectAndProve_works() public {
```
LayneHaber, line 43: this needs to be an integration level test, not just on forge.

even if the message is processed twice, the `InboxFacet` prevents the funds from being minted twice. however, the `proveAndProcess` function would succeed, and im concerned about a drain on relayer fees if old messages are reattempted.

because this is a system-level test, we need to test on testnet. but because it is majorly a concern if *spoke connectors* are replaced, this testing is likely not incredibly impactful

LayneHaber: but we should have unit tests

packages/deployments/contracts/contracts/messaging/connectors/admin/AdminHubConnector.sol
```diff
@@ -0,0 +1,40 @@
...
+  constructor(
+    uint32 _domain,
+    uint32 _mirrorDomain,
+    address _rootManager
+  ) ProposedOwnable() HubConnector(_domain, _mirrorDomain, address(0), _rootManager, address(0)) {
+    _setOwner(msg.sender);
+  }
+
+  // ============ Admin fns ============
+  function addSpokeRootToAggregate(bytes32 _data) external onlyOwner {
```
LayneHaber, line 28: comments

LayneHaber (changes requested):

packages/deployments/contracts/contracts_forge/messaging/connectors/admin/AdminHubConnector.t.sol
rhlsthrm: Not sure why, this should be able to be run with normal tests.

packages/deployments/contracts/contracts/messaging/connectors/admin/AdminHubConnector.sol
```diff
@@ -0,0 +1,45 @@
...
+    IRootManager(ROOT_MANAGER).aggregate(MIRROR_DOMAIN, _data);
+
+    // Emit event
+    emit MessageProcessed(abi.encode(_data), msg.sender);
+  }
+
+  // ============ Private fns ============
+
+  // function included to match interface but is not used internally, therefore is a no op.
+  function _verifySender(address _expected) internal view override returns (bool) {}
```
LayneHaber, line 41: if its not used it should revert

packages/deployments/contracts/contracts/messaging/connectors/admin/AdminHubConnector.sol
```diff
@@ -0,0 +1,45 @@
...
+
+  // function included to match interface but is not used internally, therefore is a no op.
+  function _sendMessage(bytes memory _data, bytes memory _encodedData) internal override {}
```
LayneHaber, line 44: this is a no-op to ensure the root manager calling `sendMessage` doesn't revert

packages/deployments/contracts/contracts_forge/messaging/connectors/admin/AdminHubConnector.t.sol
```diff
@@ -0,0 +1,123 @@
...
+  address owner = 0x4d50a469fc788a3c0CdC8Fd67868877dCb246625;
+  SpokeConnector ETH_SPOKE_CONNECTOR = SpokeConnector(payable(0xF7c4d7dcEc2c09A15f2Db5831d6d25eAEf0a296c));
+  RootManager ROOT_MANAGER = RootManager(0xd5d61E9dfb6680Cba8353988Ba0337802811C2e1);
+
+  // ============ config
+  AdminHubConnector adminHubConnector;
+
+  // ============ Setup ============
+  function setUp() public {
+    vm.createSelectFork("https://eth.llamarpc.com", 17528353);
```
LayneHaber, line 35: should use what's in the `foundry.toml` or the utilities in `MotherForker`

packages/deployments/contracts/contracts_forge/messaging/connectors/admin/AdminHubConnector.t.sol
```diff
@@ -0,0 +1,123 @@
...
+
+    encodedData = new bytes[](6);
+    encodedData[0] = hex"";
+    encodedData[1] = hex"";
+    encodedData[2] = hex"";
+    encodedData[3] = hex"";
+    encodedData[4] = hex"";
+    encodedData[5] = hex"";
+
+    ROOT_MANAGER.propagate(connectors, fees, encodedData);
```
LayneHaber, line 116: no assertions here

LayneHaber (changes requested):
need to change the deploy script as well

LayneHaber:
Adding these types of fork tests into the default tests will also make CI more flakey, as forge testing will fail on any RPC failure as well

LayneHaber:
okay, I added a separate deploy script to allow admin hub connector deployments by tag only.

ideally, this should not be run with any other part of the deployment (this means we should only ever run with deployment tags, as I cannot enforce these are run separately).

there are also several places in our `init` script where there's an embedded assumption that 2 connectors exist for each domain. instead, I added a separate hardhat task to log the transactions needed to replace a connector with an admin one:

```sh
ywc hardhat enroll-admin-connector --deployment "ChapelHubConnector" --network "goerli"
```

Please note that the above is a trashy fix, and I'm not sure where that leaves the init script for BNB or while we have the admin connector enrolled. I think it will be okay because both sides are deployed, but I'm not entirely sure.

packages/deployments/contracts/contracts/messaging/connectors/admin/AdminHubConnector.sol
```diff
@@ -37,9 +41,18 @@ contract AdminHubConnector is ProposedOwnable, HubConnector {
...
+  function _verifySender(address) internal pure override returns (bool) {
+    require(false, "!supported");
+  }
 
-  // function included to match interface but is not used internally, therefore is a no op.
+  /**
+   * @notice This function is a no-op as this connector does NOT send messages across
+   * chains.
+   * @dev Is a no-op over a revert so `RootManager.sendMessage` does not revert.
+   */
```
rhlsthrm, line 56: I was going to say we should make this revert because the try-catch will log it as an error which is maybe expected. But we are changing this system so it's better to make it a no-op.

packages/deployments/contracts/package.json
```diff
@@ -3,8 +3,9 @@
   "version": "2.0.0-alpha.1",
   "description": "",
   "scripts": {
-    "test": "yarn forge test --no-match-path '*/upgrade/**.sol'",
+    "test": "yarn forge test --no-match-path '*/fork/**.sol'",
     "test:upgrade": "ts-node ./src/cli/upgrade/main.ts",
+    "test:fork": "yarn forge test --match-path '*/fork/**.sol'",
```
rhlsthrm, line 8: So do we not actually run fork tests as part of CI?

LayneHaber: it will be finicky if the provider fails, depends on your tolerance for bad CI

packages/deployments/contracts/contracts/messaging/connectors/admin/AdminHubConnector.sol
```diff
@@ -29,6 +29,10 @@ contract AdminHubConnector is ProposedOwnable, HubConnector {
    * @param _data The data to be added to the merkle tree
    */
   function addSpokeRootToAggregate(bytes32 _data) external onlyOwner {
+    // Sanity check: ensure the _data is nonzero
+    require(_data != bytes32(0), "empty");
```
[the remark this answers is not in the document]
LayneHaber: fair, but it would be more of a silly thing to do here

LayneHaber (dismissed):

liu-zhipeng (approved):

LayneHaber (approved):

PR #4578, opened 2023-06-23
feat: retry mechanism in lh publisher

wanglonghong:
## Type of change

- [x] Refactoring

## Related Issue(s)

Fixes

### DIFF
--- a/packages/adapters/cache/src/lib/caches/messages.ts
+++ b/packages/adapters/cache/src/lib/caches/messages.ts
@@ -1,33 +1,126 @@
+import { XMessage, getNtpTimeSeconds } from "@connext/nxtp-utils";
 import { Cache } from "./cache";
 
 /**
  * Redis Store Details:
  * Message Nonce:
- *   key: messages:nonce:$originDomain | value: number;
+ *   key: messages:nonce:$domain | value: number;
  */
 export class MessagesCache extends Cache {
   private readonly prefix = "messages";
 
   /// MARK - Message Nonce
   /**
-   * Gets the latest nonce for the given domain pair.
-   * @param originDomain - The origin domain.
-   * @param destinationDomain - The destination domain.
-   * @returns The latest nonce of the given pair.
+   * Gets the latest nonce for the given domain.
+   * @param domain - The domain ID.
+   * @returns The latest nonce of the given domain.
    */
-  public async getNonce(originDomain: string, destinationDomain: string): Promise<number> {
-    const res = await this.data.hget(`${this.prefix}:nonce`, `${originDomain}-${destinationDomain}`);
+  public async getNonce(domain: string): Promise<number> {
+    const res = await this.data.hget(`${this.prefix}:nonce`, domain);
     return res ? +res : 0;
   }
 
   /**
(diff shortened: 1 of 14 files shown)


packages/adapters/cache/src/lib/caches/messages.ts
```diff
@@ -1,33 +1,126 @@
...
+  /**
+   * Stores a message with `attempt` value
+   * @param message - The target message to store
+   * @param attempt - The retry count
+   */
+  private async storeMessage(message: XMessage, attempt?: number): Promise<number> {
+    await this.addPending(message.originDomain, message.destinationDomain, message.leaf);
+    return await this.data.hset(
+      `${this.prefix}:data`,
+      message.leaf,
```
preethamr, line 64: Can we be sure that message leafs will be unique across all domains ?

packages/agents/lighthouse/src/tasks/prover/operations/publisher.ts
```diff
@@ -22,11 +23,93 @@ import { DEFAULT_PROVER_BATCH_SIZE, DEFAULT_PROVER_PUB_MAX } from "../../../conf
...
+  let offset = 0;
+  const limit = 1000;
+  let end = false;
+  while (!end) {
+    const leaves = await cache.messages.getPending(originDomain, destinationDomain, offset, limit);
+    for (const leaf of leaves) {
+      const message = await cache.messages.getMessage(leaf);
+      if (
+        message &&
+        getNtpTimeSeconds() - message.timestamp > waitTime * 2 ** message.attempt &&
```
preethamr, line 95: Consider this scenario:
There is a long backlog in the queue > `waitTime * 2 ** message.attempt`. Since only consumer will bump the attempt count, after consuming, publisher will include the message in the batch, in every iteration, and create duplicates.

Solution could be, consumer check if it is still pending, and if so does the same timestamp check as publisher, before consuming the message. So that the exponential back off is applied

packages/agents/lighthouse/src/tasks/prover/operations/process.ts
```diff
@@ -207,7 +207,17 @@ export const processMessages = async (brokerMessage: BrokerMessage, _requestCont
       requestContext,
     );
     logger.info("Proved and processed message sent to relayer", requestContext, methodContext, { taskId });
+    if (taskId) {
+      await cache.messages.removePending(
+        originDomain,
+        destinationDomain,
+        messages.map((it) => it.leaf),
```
preethamr, line 214: should only remove the message from pending, if a proof was included in this batch.

preethamr: Fixed

packages/agents/lighthouse/src/tasks/prover/operations/publisher.ts
preethamr: Fixed

packages/adapters/cache/src/lib/caches/messages.ts
preethamr: Resolving it as safe assumption.

Issue #4538, opened 2023-06-19
Onchain task for BNB AMB


liu-zhipeng:
## Problem
Replace BNB multichain connector

## Context
- https://github.com/connext/monorepo/issues/4517
https://connext.notion.site/Pre-Spec-for-BSC-Amb-3578aea3d7a4485eb63498ece0b22fbd

## Action Items
- [x] HubConnector for Wormhole 
- [x] SpokeConnector for Wormhole
- [x] Forge Tests: Unit test covering all fuzz cases
- [x] E2E testing 
- [x] HH tasks for deployment and test
- [x] Init script
- [x] deployment at Testnet for initial testing for the Wormhole contracts
- [x] Deploy Connector for Mainnet
- [x] Deploy Connector for spoke

## Acceptance criteria

- [ ] E2E working at `BNB`, `ARBITRUM`, and `ETHEREUM` (Fork Mainnet).
- [ ] Succesful xcall() via HH tasks
- [ ] Succesful sendMessage() via HH tasks
- [ ] Succesful propagate() via HH tasks
- [ ] Succesful proveAndProcess() via HH tasks
- [ ] Does not double-reconcile transfers ( `MessageStatus` checks are overridden on new spoke deploy)
- [ ] Does not double-process transfers (should not cause many transfers to be `proveAndProcess`ed again. this will require offchain verification against newly deployed subgraphs)

Fixes #4623

rhlsthrm:
Verified that old message cannot be processed by new connector, so double spend is not an issue.

rhlsthrm:
Created forge test, requested review, R4R.

rhlsthrm:
Contract is done, need to test on testnet, but Wormhole does not support Goerli. Need to either test on other testnets or mainnet.

rhlsthrm:
Verified messaging from BNB to ETH, still verifying ETH to BNB.

LayneHaber:
txs + costs:

```md
ETH -> BSC 
 - 0.000254019196160768 ETH ($0.48) paid on ETH 
 - 0.000713894285678591 BNB ($0.17) refund
 - total: $0.31 paid
- txs:
https://etherscan.io/tx/0x580a3dcec415417b949e947e1335b58b6eeb31f128136b48a3b0c35e0570cbfa
https://bscscan.com/tx/0xc5478c6cacb5fffc2ced78201dd1d6e5d7f6e2b3ab41f89c5b30cc4788e51790

BSC -> ETH
[...]

PR #4623, opened 2023-06-30
feat: 4538 wormhole connector

liu-zhipeng:
## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes


packages/deployments/contracts/contracts/messaging/connectors/wormhole/BaseWormhole.sol
```diff
@@ -0,0 +1,116 @@
...
+import {IWormholeReceiver} from "../../interfaces/ambs/wormhole/IWormholeReceiver.sol";
+
+import {GasCap} from "../GasCap.sol";
+import "forge-std/console.sol";
+
+abstract contract BaseWormhole is GasCap, IWormholeReceiver {
+  // ============ Storage ============
+  address public immutable relayerAddress;
+  uint16 public immutable MIRROR_CHAIN_ID;
+  mapping(bytes32 => bool) public processedWhMessages;
```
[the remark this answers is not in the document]
LayneHaber: if we are supposed to protect against wormhole replay, we have to keep in mind that these connectors will be dropped in the future and we could lose that state.

if we don't plan on managing that, replay protection should be at a different level since its only protected against during the lifetime of the connector

packages/deployments/contracts/contracts/messaging/connectors/wormhole/BaseWormhole.sol
```diff
@@ -0,0 +1,116 @@
...
+  function _verifySender(address _mirrorConnector, address _expected) internal pure returns (bool) {
+    return _mirrorConnector == _expected;
+  }
+
+  // DO NOT override _processMessage, should revert from `Connector` class. All messages must use the _processMessageFrom function
+  /**
+   * @notice This function is called to handle incoming messages. Should store the latest
+   * root generated on the l2 domain.
+   */
+  function _processMessageFrom(address sender, bytes memory _data) internal virtual;
```
LayneHaber, line 69:
```suggestion
  function _processMessageFrom(address _sender, bytes memory _data) internal virtual;
```

packages/deployments/contracts/contracts/messaging/connectors/wormhole/BaseWormhole.sol
```diff
@@ -0,0 +1,116 @@
...
+    processedWhMessages[deliveryHash] = true;
+
+    _processMessageFrom(fromWormholeFormat(sourceAddress), payload);
+  }
+
+  /**
+   * @dev calculcate gas to call `receiveWormholeMessages` on target chain
+   * https://github.com/wormhole-foundation/wormhole/blob/main/ethereum/contracts/relayer/deliveryProvider/DeliveryProvider.sol
+   */
+  function quoteEVMDeliveryPrice(uint256 gasLimit) public view returns (uint256 cost) {
```
LayneHaber, line 52:
```suggestion
  function quoteEVMDeliveryPrice(uint256 _gasLimit) public view returns (uint256 cost) {
```

packages/deployments/contracts/contracts/messaging/connectors/wormhole/BaseWormhole.sol
```diff
@@ -0,0 +1,116 @@
...
+   * @notice This function is called to receive messages through the wormhole relayer module
+   * https://book.wormhole.com/technical/evm/relayer.html
+   */
+  function receiveWormholeMessages(
+    bytes memory payload,
+    bytes[] memory additionalVaas,
+    bytes32 sourceAddress,
+    uint16 sourceChain,
+    bytes32 deliveryHash
+  ) public payable override {
```
LayneHaber, line 33:
```suggestion
  function receiveWormholeMessages(
    bytes memory _payload,
    bytes[] memory _additionalVaas,
    bytes32 _sourceAddress,
    uint16 _sourceChain,
    bytes32 _deliveryHash
  ) public payable override {
```

packages/deployments/contracts/contracts/messaging/connectors/wormhole/BaseWormhole.sol
```diff
@@ -0,0 +1,116 @@
...
+      _mirrorConnector,
+      _data,
+      0,
+      gasLimit,
+      MIRROR_CHAIN_ID, // refundChain
+      _refund // refundAddress
+    );
+  }
+
+  function fromWormholeFormat(bytes32 whFormatAddress) internal pure returns (address) {
```
LayneHaber, line 101:
```suggestion
  function _fromWormholeFormat(bytes32 _whFormatAddress) internal pure returns (address) {
```

packages/deployments/contracts/contracts/messaging/connectors/wormhole/BaseWormhole.sol
```diff
@@ -0,0 +1,116 @@
...
+    require(uint256(whFormatAddress) >> 160 == 0, "!evm address");
+    return address(uint160(uint256(whFormatAddress)));
+  }
+
+  /**
+   * @notice Using Wormhole relayer (AMB), the gas is provided to `sendMessage` as an encoded uint
+   */
+  function _getGasFromEncoded(bytes memory _encodedData) internal view returns (uint256 _gas) {
+    // Should include gssas info in specialized calldata
+    require(_encodedData.length == 32, "!encoded data length");
```
LayneHaber, line 111: in other connectors, this input validation is done at the top level within `_sendMessage` I believe. we should make sure this is consistent with existing code

packages/deployments/contracts/contracts/messaging/connectors/wormhole/WormholeHubConnector.sol
```diff
@@ -0,0 +1,50 @@
...
+    require(_data.length == 32, "!length");
+
+    // set the outbound root for BSC domain
+    IRootManager(ROOT_MANAGER).aggregate(MIRROR_DOMAIN, bytes32(_data));
+
+    emit MessageProcessed(_data, msg.sender);
+  }
+
+  function _sendMessage(bytes memory _data, bytes memory _encodedData) internal override {
+    _sendMessage(mirrorConnector, owner(), _data, _encodedData);
```
LayneHaber, line 48: which chain do you get the refund on?

packages/deployments/contracts/contracts/messaging/connectors/wormhole/BaseWormhole.sol
```diff
@@ -0,0 +1,116 @@
...
+   */
+  function receiveWormholeMessages(
+    bytes memory payload,
+    bytes[] memory additionalVaas,
+    bytes32 sourceAddress,
+    uint16 sourceChain,
+    bytes32 deliveryHash
+  ) public payable override {
+    require(sourceChain == MIRROR_CHAIN_ID, "!source chain");
+    require(msg.sender == relayerAddress, "!relayer");
```
LayneHaber, line 35: so, when we call `processMessageFromRoot` on the backend the call stack will be:

1. Call some fn on the wormhole relayer
2. That calls `_receiveWormholeMessages`
3. `_processMessageFrom` processes the aggregate or spoke root, depending on direction

is the relayer different than the AMB?

packages/deployments/contracts/contracts_forge/messaging/connectors/wormhole/WormholeSpokeConnector.t.sol
```diff
@@ -0,0 +1,270 @@
...
+      )
+    );
+  }
+
+  // ============ Utils ============
+
+  // ============ send ============
+
+  // Happy path L1
+  function test_WormholeSpokeConnector_send_sendAndEmitEvent() public {
```
LayneHaber, line 60: fuzz data? not sure why it has to be an actual merkle root from a tree vs random bytes32

LayneHaber (changes requested):

packages/deployments/contracts/contracts/messaging/connectors/wormhole/WormholeHubConnector.sol
LayneHaber: ah I see its the current change. instead we should still have the ability to set a refund address as the `owner()` may not always have the option to withdraw funds

packages/deployments/contracts/contracts/messaging/connectors/wormhole/BaseWormhole.sol
LayneHaber: its not, I refactored to remove `relayerAddress` to make that clear.

packages/deployments/contracts/contracts/messaging/connectors/wormhole/BaseWormhole.sol
LayneHaber: ahh it's the same as in other contracts already, resolving

packages/deployments/contracts/deployConfig/shared.ts
```diff
@@ -407,27 +407,27 @@ export const MESSAGING_PROTOCOL_CONFIGS: Record<string, MessagingProtocolConfig>
...
+          spoke: "0x27428DD2d3DD32A4D7f7C497eAaa23130d894911",
         },
         delayBlocks: DEFAULT_DELAY_BLOCKS,
         processGas: DEFAULT_PROCESS_GAS,
         reserveGas: DEFAULT_RESERVE_GAS,
         custom: {
           hub: {
-            mirrorChainId: "56",
-            gasCap: "520000000000000", // calcSrcFee: 516960000000000
+            gasCap: "300000", // gas limit for receiveWormholeMessages on bnb
```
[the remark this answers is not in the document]
liu-zhipeng: As i tested, that was enough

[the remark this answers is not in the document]
liu-zhipeng: image

image

0.00510868 ETH refunded!

packages/deployments/contracts/deployConfig/shared.ts
```diff
@@ -407,27 +407,27 @@ export const MESSAGING_PROTOCOL_CONFIGS: Record<string, MessagingProtocolConfig>
...
+            mirrorChainId: "4", // bsc wormhole chainId: 4
           },
           spoke: {
-            mirrorChainId: "1",
-            gasCap: "150000000000000000", // calcSrcFee: 140112000000000000
+            gasCap: "400000", // gas limit for receiveWormholeMessages on mainnet
+            mirrorChainId: "2", // mainnet wormhole chainid: 2
```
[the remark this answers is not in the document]
liu-zhipeng: Same

packages/deployments/contracts/contracts/messaging/connectors/wormhole/BaseWormhole.sol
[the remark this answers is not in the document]
liu-zhipeng: Obviously we can't protect replaying stuff just with `processedWhMessages`, in case of dropping and replacing spoke connector. Maybe we need this mapping on other layer, like `MerkleTreeManager`.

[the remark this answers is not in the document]
LayneHaber: to effectively protect against replays, yes. however, it doesn't make sense to put different mappings there. other types of protection we could add:

1. Enforce no duplicate leaves are able to be added -- this would have to be on both the `RootManager`'s `MerkleTree` as well as the `SpokeConnector`
2. Ensuring duplicate messages are not processable twice -- this check is pending for the larger messaging upgrade, and exist son the execution layer.

both (1) and (2) **significantly** increase the scope and risk of this upgrade. what happens if there are duplicate leaves in the `RootManager`'s tree? specifically with the pricing utilities

LayneHaber:
where are we at with the network-wide tests mentioned in #4538 ?

liu-zhipeng:
> 

https://github.com/connext/monorepo/tree/4538-wormhole-connector-test

liu-zhipeng:
https://github.com/connext/monorepo/pull/4659
Offchain tasks ^^^

packages/deployments/contracts/contracts/messaging/connectors/wormhole/BaseWormhole.sol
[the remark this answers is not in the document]
LayneHaber: okay chatted with Rahul -- there should be no problem if we lose this check and get transport replays post connector upgrades. we can leave it in to prevent excess fees

packages/agents/lighthouse/src/tasks/sendOutboundRoot/helpers/bnb.ts
```diff
@@ -33,14 +33,20 @@ export const getSendOutboundRootParams = async (l2domain: string): Promise<Extra
...
-  let encodedData = contracts.spokeConnector.encodeFunctionData("AMB");
-  let encoded = await chainreader.readTx({ data: encodedData, domain: Number(l2domain), to: l2SpokeConnector.address });
-  const [ambAddress] = contracts.spokeConnector.decodeFunctionResult("AMB", encoded);
+  const l2Provider = getJsonRpcProvider(l2RpcUrl);
+  const l2SpokeConnectorContract = getContract(
+    l2SpokeConnector.address,
+    l2SpokeConnector.abi as ContractInterface,
+    l2Provider,
+  );
+  const ambAddress = await l2SpokeConnectorContract.AMB();
```
LayneHaber, line 42: why are we using the rpc directly instead of the chain reader?

packages/agents/lighthouse/src/tasks/sendOutboundRoot/helpers/bnb.ts
```diff
@@ -33,14 +33,20 @@ export const getSendOutboundRootParams = async (l2domain: string): Promise<Extra
...
 
-  const ambInterface = getInterface(ambs.bnb);
-  encodedData = ambInterface.encodeFunctionData("calcSrcFees", ["", hubChainId, 32]);
-  encoded = await chainreader.readTx({ data: encodedData, domain: Number(l2domain), to: ambAddress });
-  const [_fee] = ambInterface.decodeFunctionResult("calcSrcFees", encoded);
+  // gasLimit on hub side = 200_000
+  // actually it required about 100_000 gas on mainnet to call `receiveWormholeMessages`
+  // remain gas will be refunded on mainnetto refund address (default: deployer address)
+  const gasLimit = "200000";
+  const fee = await l2SpokeConnectorContract.quoteEVMDeliveryPrice(gasLimit, ambAddress);
```
LayneHaber, line 48: why are we using the rpc directly instead of the chain reader?

packages/deployments/subgraph/config/amarok-hub-v0.json
```diff
@@ -32,6 +32,11 @@
         "name": "ArbitrumHubConnector",
         "address": "0xd151C9ef49cE2d30B829a98A07767E3280F70961",
         "startBlock": 16233513
+      },
+      {
+        "name": "BnbHubConnector",
+        "address": "0x69009c6f567590d8B469dbF4C8808e8ee32b8a45",
+        "startBlock": 17640454
```
LayneHaber, line 39: do we need to add this to the hub subgraph now? why?

LayneHaber:
updated the init script so it works properly, and checked that against the submitted transactions: https://github.com/connext/monorepo/pull/4668

well need to update the spoke connector address on the relayer proxy as well. the contracts all look good to me, they didnt have any issues in the audit outside of some quality nits and weve tested it onchain.

the offchain stuff still have some questions about -- namely why we are using the provider instead of the chainservice. im going to try to think of some ways we can test this without exploding scope

LayneHaber:
deployment and testing updated: https://www.notion.so/connext/Spec-Replacing-Multichain-9e5aedcac7df488b8b38b7f72dcb8535?pvs=4#0d443c6f691448ddb65bbb7c236483af

packages/agents/lighthouse/src/tasks/sendOutboundRoot/helpers/bnb.ts
liu-zhipeng: We used `rpc` from `getBestProvider` function, so i thought that would be no problem

packages/deployments/subgraph/config/amarok-hub-v0.json
liu-zhipeng: @sanchaymittal  here ^^

rhlsthrm (approved):

LayneHaber (approved):

PR #4636, opened 2023-07-05
ci: add mainnet ops

wanglonghong:
## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes

### DIFF
--- a/ops/mainnet/prod/core/main.tf
+++ b/ops/mainnet/prod/core/main.tf
@@ -297,6 +297,25 @@ module "sequencer_web3signer" {
   container_env_vars       = local.sequencer_web3signer_env_vars
 }
 
+module "lighthouse_prover_cron" {
+  source              = "../../../modules/lambda"
+  ecr_repository_name = "nxtp-lighthouse"
+  docker_image_tag    = var.lighthouse_image_tag
+  container_family    = "lighthouse-prover"
+  environment         = var.environment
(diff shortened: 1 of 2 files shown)


ops/mainnet/prod/core/main.tf
```diff
@@ -297,6 +297,25 @@ module "sequencer_web3signer" {
...
+  source              = "../../../modules/lambda"
+  ecr_repository_name = "nxtp-lighthouse"
+  docker_image_tag    = var.lighthouse_image_tag
+  container_family    = "lighthouse-prover"
+  environment         = var.environment
+  stage               = var.stage
+  container_env_vars = merge(local.lighthouse_env_vars, {
+    LIGHTHOUSE_SERVICE = "prover-pub"
+  })
+  schedule_expression    = "rate(5 minutes)"
```
rhlsthrm, line 310: Rate needs to be > than timeout right?

rhlsthrm (changes requested):

ops/mainnet/prod/core/main.tf
wanglonghong: I don't think so. we have a mechanism to prevent multiple prover:pub running at the same time.

rhlsthrm (approved):

sanchaymittal (approved):

PR #4643, opened 2023-07-05
fix: add pagination on message status check

preethamr:
## Description
Add pagination on message status check

## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes

### DIFF
--- a/packages/agents/cartographer/poller/src/lib/operations/messagestatus.ts
+++ b/packages/agents/cartographer/poller/src/lib/operations/messagestatus.ts
@@ -12,18 +12,28 @@ export const updateMessageStatus = async () => {
 
   for (const domain of domains) {
     const limit = 100;
-    logger.debug("Updating message status", requestContext, methodContext, { domain, limit });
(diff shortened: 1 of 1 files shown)


packages/agents/cartographer/poller/src/lib/operations/messagestatus.ts
```diff
@@ -12,18 +12,28 @@ export const updateMessageStatus = async () => {
 
   for (const domain of domains) {
     const limit = 100;
```
rhlsthrm, line 14: We should be able to pull more than 100 from the DB, right? What would be a good value to use here that is okay on mem usage but still pulls in enough records? 1000?

preethamr: Yes, we could use larges pages. Updated it to 1000.

LayneHaber (approved):

LayneHaber (approved):

wanglonghong (approved):

Issue #4730, opened 2023-07-25
[sdk-server] /approveIfNeeded taking a min to respond


sanchaymittal:
**Describe the bug**

- Initial investigation suggests the issue is at network latency by RPCs
- Though we do require testing server status as well

[NEW ISSUE] Response is coming gateway error.

Issue reported by **Alchemix**

**To Reproduce**
2. 
3. 

**Additional context**

just-a-node:
Caused by use of default public providers in sdk-server. Users should be passing in a provider to the `options` param:

```
POST {{baseUrl}}/calculateSwap
Content-Type: application/json

{
  "domainId": "1869640809",
  "tokenAddress": "0x94b008aA00579c1307B0EF2c499aD98a8ce58e58",
  "tokenIndexFrom": "0",
  "tokenIndexTo": "1",
  "amount": "100000000000000",
  "options": {
    "originProviderUrl": "https://optimism-mainnet.infura.io/v3/<my_api_key>"
  }
}
```

This is a list of functions that users should pass `options` for:
```
approveIfNeeded: 180203ms
calculateSwap: 120274ms
calculateTokenAmount: 120305ms
calculateRemoveSwapLiquidity: 180567ms
calculateRemoveSwapLiquidityOneToken: 138842ms
calculateAddLiquidityPriceImpact: 120195ms
calculateAddLiquidityPriceImpact:  120336ms
calculateSwapPriceImpact: 120298ms
getTokenSupply: 180167ms
getTokenUserBalance: 180237ms
getPoolTokenBalance: 120187ms
getVirtualPrice: 130283ms 
removeLiquidityImbalance: 180176ms
swap: 180522ms
```

PR #4731, opened 2023-07-27
fix: provider passing in options

prathmeshkhandelwal1:
## Type of change

- [ ] Refactoring

## Related Issue(s)

#4730

### DIFF
--- a/packages/agents/sdk-wrapper/src/sdkPool.ts
+++ b/packages/agents/sdk-wrapper/src/sdkPool.ts
@@ -41,7 +41,6 @@ export class SdkPool extends SdkShared {
   ): Promise<BigNumber> {
     const _options = options ?? {
       chains: this.config.chains,
-      originProviderUrl: this.config.chains[domainId]?.providers?.[0],
       signerAddress: this.config.signerAddress,
     };
 
@@ -181,7 +180,6 @@ export class SdkPool extends SdkShared {
   ): Promise<BigNumber> {
     const _options = options ?? {
       chains: this.config.chains,
-      originProviderUrl: this.config.chains[domainId]?.providers?.[0],
       signerAddress: this.config.signerAddress,
     };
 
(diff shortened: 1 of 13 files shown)


packages/agents/sdk-wrapper/src/sdkPool.ts
```diff
@@ -39,10 +39,14 @@ export class SdkPool extends SdkShared {
     amount: BigNumberish,
     options?: Options,
   ): Promise<BigNumber> {
-    const _options = options ?? {
-      chains: this.config.chains,
-      signerAddress: this.config.signerAddress,
-    };
+    const originProviderUrl = this.config.chains[domainId]?.providers;
+
+    const _options = options ?? {};
```
just-a-node, line 44: We still need to pass chains and signerAddress from config

prathmeshkhandelwal1: So these specific functions dont really require chains and signerAddress internally.

just-a-node: swap() will need signerAddress, for example

packages/agents/sdk-wrapper/src/sdkPool.ts
```diff
@@ -41,8 +41,10 @@ export class SdkPool extends SdkShared {
   ): Promise<BigNumber> {
     const _options = options ?? {
       chains: this.config.chains,
+      originProviderUrl: this.config.chains[domainId]?.providers?.[0],
```
sanchaymittal, line 44: We need to use fallback provider here, can't rely on first provider implementation.
Ideally we should be using `chains` options for wrapper-sdk and fetch the provider from it under the core-sdk. single_url options for origin and destination were added for any direct request to sdk-server only.

prathmeshkhandelwal1:
Reverting this to @just-a-node Commit. If we are making changes according to this conversation https://github.com/connext/monorepo/pull/4731#discussion_r1277188024 . Then those features has been added to the PR attached here : #4736 
**We need to merge above mentioned PR before merging this.**

packages/agents/sdk-wrapper/src/sdkPool.ts
just-a-node: Now getting provider from `chains` config that is supplied by client during SDK init.

just-a-node (approved):

rhlsthrm (approved):

PR #4751, opened 2023-07-31
add P8 to monorepo

princepspolycap:
This PR adds the P8 token to the list of supported assets on the Gnosis Chain. The token's address, name, and other relevant details have been included in the configuration.

Type of change:
Adding a token


packages/deployments/contracts/src/cli/init/config/mainnet/production.ts
```diff
@@ -290,6 +290,21 @@ export const MAINNET_PRODUCTION_INIT_CONFIG: InitConfig = {
...
+    {
+      name: "P8",
+      canonical: {
+        domain: "100",
+        address: "0x772fCe4B8E88BD19e86dC92428d242704aC480a0",
+        decimals: 18,
+        cap: utils.parseUnits("10000000", 18).toString(),
+      },
+      representations: {
+        "100": {
```
rhlsthrm, line 302: Also canonical and representations should never be the same.

rhlsthrm (changes requested):
Need to use the correct domains.

prathmeshkhandelwal1 (changes requested):
Mainnet config and assets added for testnets. Home chain in this case is gnosis (which is not a testnet chain).

packages/deployments/contracts/src/cli/init/config/testnet/production.ts
```diff
@@ -132,6 +132,41 @@ export const TESTNET_PRODUCTION_INIT_CONFIG: InitConfig = {
...
+    {
+      name: "P8",
+      canonical: {
+        domain: "6778479",
+        address: "0x772fCe4B8E88BD19e86dC92428d242704aC480a0",
+        decimals: 18,
+      },
+      representations: {
+        // mainnet
+        "6648936": {
```
prathmeshkhandelwal1, line 144: If a token is not present for any chain, you can just not add that particular domain (chain) in the config. It will work fine, No need for zero address.

Reference Link: https://github.com/connext/monorepo/blob/79dae16d9f88f19cfd8a06d3a567899fc920ed92/packages/deployments/contracts/src/cli/init/config/testnet/production.ts#L136

packages/deployments/contracts/src/cli/init/config/testnet/production.ts
```diff
@@ -132,6 +132,41 @@ export const TESTNET_PRODUCTION_INIT_CONFIG: InitConfig = {
...
+          local: "0x0000000000000000000000000000000000000000",
+          adopted: "0x0000000000000000000000000000000000000000",
+        },
+        // polygon
+        "1886350457": {
```
prathmeshkhandelwal1, line 149: Similar here.

prathmeshkhandelwal1 (changes requested):

prathmeshkhandelwal1:
Hello @princepspolycap,

I've noticed that your token appears to be solely on Gnosis, and I can't find any representation of it other than the canonical version. In other words, it seems there's no token available on different chains to facilitate cross-chain transactions.

If possible, could you please provide me with your Telegram ID? That way, we can have a more in-depth conversation to clarify all the details.

prathmeshkhandelwal1 (changes requested):
No representation provided

PR #4785, opened 2023-08-07
feat: manage relayer tasks

wanglonghong:
## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes

### DIFF
--- a/packages/adapters/cache/src/lib/caches/messages.ts
+++ b/packages/adapters/cache/src/lib/caches/messages.ts
@@ -145,6 +145,83 @@ export class MessagesCache extends Cache {
   }
 
   /**
+   * Add pending task to the `tasks` list.
+   *
+   * @param taskId - The taskId.
+   * @param relayer - The relayer type which can be either Gelato or Connext.
+   * @param leaves - The leaves that the task includes.
+   */
+  public async addTaskPending(
+    taskId: string,
+    relayer: string,
+    originDomain: string,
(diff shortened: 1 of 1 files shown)


sanchaymittal (approved):

liu-zhipeng (approved):

sanchaymittal:
will wait for preetham's approval to for above

sanchaymittal:
@preethamr waiting for your review

preethamr (approved):
What would be the impact on performance ?

wanglonghong:
> What would be the impact on performance ?

Once the task is sent to gelato, it just removes messages from pending list and nonce also increases.
In worst case(out of funds), the task can fail on gelato's end but that batch messages will never picked up because it has been already removed and don't re-add it to pending list

preethamr:
> In worst case(out of funds), the task can fail on gelato's end but that batch messages will never picked up because it has been already removed and don't re-add it to pending list

This is a possible scenario right ?  How would we then retry ?

wanglonghong:
> How would we then retry ?

1. We don't remove the batch from the pending list but we update its status to `sent`.
2. In each run, it tries to get the task status thought gelato api. 
3. If Cancelled, it updates the statuses to `nonce` which means it can be picked up again by lighthouse-prover-pub.
4. If ExecSuccess, it removes the batch from pending,

PR #4951, opened 2023-10-02
Optimistic roots v1.1

gotzenx:
## Description
Optimistic Roots v1.1
With v1.0 we introduced the optimistic system to the Root Manager contract (L1) removing the need of:
Use AMBs to send the inbound roots from Spoke Connectors (L2s) to Root Manager (mainnet)
Calculate the aggregate root on-chain on L1.

With v1.1 we aim to bring the optimistic system on the Spoke Connectors to remove the need of propagating the aggregate root from Root Manager -> Spokes through AMBs.

## Type of change

- [X] Refactoring
- [X] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [X] Breaking change (fix or feature that would cause existing functionality to change)
- [X] Requires changes in customer code

## Related Issue(s)
Not sure if theres a current issue for it, but current v1 code still has the DOS attack on the `propagate` function. This PR fixes that attack vector.


packages/deployments/contracts/contracts/messaging/connectors/SpokeConnector.sol
```diff
@@ -265,6 +387,46 @@ abstract contract SpokeConnector is Connector, ConnectorManager, WatcherClient,
...
+  /**
+   * @notice Set the `disputeBlocks`, the duration, in blocks, of the dispute process for
+   * a given proposed root
+   */
+  function setMinDisputeBlocks(uint256 _minDisputeBlocks) public onlyOwner {
+    if (_minDisputeBlocks == minDisputeBlocks)
+      revert SpokeConnector_setMinDisputeBlocks__SameMinDisputeBlocksAsBefore();
+    emit MinDisputeBlocksUpdated(minDisputeBlocks, _minDisputeBlocks);
+    minDisputeBlocks = _minDisputeBlocks;
+  }
```
skeletor-spaceman, line 417: why is `minDisputeBlocks` needed? if both functions are `onlyOwner`

packages/deployments/contracts/contracts/messaging/connectors/SpokeConnector.sol
```diff
@@ -502,6 +729,7 @@ abstract contract SpokeConnector is Connector, ConnectorManager, WatcherClient,
    * @param _newRoot Received aggregate
    */
   function receiveAggregateRoot(bytes32 _newRoot) internal {
+    if (optimisticMode) revert SpokeConnector_receiveAggregateRoot__OptimisticModeOn();
```
skeletor-spaceman, line 732: why would we want to block incoming roots from the AMBs?

sanchaymittal (approved):

PR #4984, opened 2023-10-07
fix: tag images with the branch name

carlomazzaferro:
## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes

### DIFF
--- a/ops/modules/ecr/main.tf
+++ b/ops/modules/ecr/main.tf
@@ -12,12 +12,52 @@ resource "aws_ecr_lifecycle_policy" "remove_old_images" {
     "rules": [
         {
             "rulePriority": 1,
-            "description": "Expire images older than 20 days",
+            "description": "Expire main images that are not the last 50",
             "selection": {
-                "tagStatus": "any",
+                "tagStatus": "tagged",
+                "tagPrefixList": ["main-"],
+                "countType": "imageCountMoreThan",
+                "countNumber": 50
+            },
+            "action": {
+                "type": "expire"
+            }
+        },
+        {
+            "rulePriority": 2,
+            "description": "Expire staging images that are not the last 20",
+            "selection": {
+                "tagStatus": "tagged",
+                "tagPrefixList": ["staging-"],
+                "countType": "imageCountMoreThan",
+                "countNumber": 20
+            },
+            "action": {
+                "type": "expire"
+            }
+        },
+        {
+            "rulePriority": 3,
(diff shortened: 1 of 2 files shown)


preethamr (approved):

preethamr:
Fixes: https://github.com/connext/monorepo/issues/4983

carlomazzaferro:
closed in favour of https://github.com/connext/monorepo/pull/4984

ops/modules/ecr/main.tf
```diff
@@ -12,12 +12,52 @@ resource "aws_ecr_lifecycle_policy" "remove_old_images" {
...
         {
             "rulePriority": 1,
-            "description": "Expire images older than 20 days",
+            "description": "Expire main images that are not the last 50",
             "selection": {
-                "tagStatus": "any",
+                "tagStatus": "tagged",
+                "tagPrefixList": ["main-"],
+                "countType": "imageCountMoreThan",
+                "countNumber": 50
```
preethamr, line 20: What is the intuition behind having main > staging > prod counts?

ops/modules/ecr/main.tf
```diff
@@ -12,12 +12,52 @@ resource "aws_ecr_lifecycle_policy" "remove_old_images" {
...
+            "action": {
+                "type": "expire"
+            }
+        },
+        {
+            "rulePriority": 3,
+            "description": "Expire prod images that are not the last 5",
+            "selection": {
+                "tagStatus": "tagged",
+                "tagPrefixList": ["prod-"],
```
preethamr, line 44: Can we differentiate between `testnet-prod` and `prod` ?

preethamr (changes requested):

ops/modules/ecr/main.tf
carlomazzaferro: Mostly due to how often they get built. Merges to main happen almost daily, to prod anywhere from once every few days to once every 2 weeks. 

The idea is keeping a reasonable amount of images in case rollbacks are needed

ops/modules/ecr/main.tf
carlomazzaferro: That'd be together with `testnet-` - I don't think we need to differentiate them. Wdyt?

preethamr: ic, that works, and it is easy to change if we need to.

preethamr (approved):

rhlsthrm (approved):

PR #5068, opened 2023-10-27
feat: Linea mainnet deployment

sanchaymittal:
## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes


just-a-node (approved):

packages/deployments/subgraph/config/amarok-hub-v0.json
```diff
@@ -37,6 +37,11 @@
         "name": "BnbHubConnector",
         "address": "0x69009c6f567590d8B469dbF4C8808e8ee32b8a45",
         "startBlock": 17640454
+      },
+      {
+        "name": "LineaHubConnector",
+        "address": "0x076cD2B25cb1Ed7272d716bdeb4A8551CF606a3D",
+        "startBlock": 18468809
```
wanglonghong, line 44: I don't think we need `LineaHubConnector` address here if we don't have a network entity in subgraph.template.yaml. or maybe we are currently missing it in the template file???

wanglonghong (commented):
One thing more we have to take care of after the merge into prod is the gasEstimates values in chaindata repo. because it hasn't set up yet but also understandable right now. but we should definitely set the proper values as soon as we go live.

preethamr (dismissed):

ops/mainnet/prod/backend/config.tf
```diff
@@ -50,6 +50,9 @@ locals {
       "6778479" = {
         providers = ["https://rpc.ankr.com/gnosis"]
       }
+      "1818848877" = {
+        providers = ["https://rpc.linea.build"]
+      }
```
LayneHaber, line 55: wont we need a more stable rpc for the expected load?

LayneHaber: guess its more important for core

packages/deployments/contracts/deployConfig/shared.ts
```diff
@@ -476,6 +476,19 @@ export const MESSAGING_PROTOCOL_CONFIGS: Record<string, MessagingProtocolConfig>
...
+        prefix: "Linea",
+        ambs: {
+          // https://etherscan.io/address/0xd19d4B5d358258f05D7B411E21A1460D11B0876F
+          hub: "0xd19d4B5d358258f05D7B411E21A1460D11B0876F",
+          // https://lineascan.build/address/0x508Ca82Df566dCD1B0DE8296e70a96332cD644ec
+          spoke: "0x508Ca82Df566dCD1B0DE8296e70a96332cD644ec",
+        },
+        processGas: DEFAULT_PROCESS_GAS,
+        reserveGas: DEFAULT_RESERVE_GAS,
+        delayBlocks: DEFAULT_DELAY_BLOCKS,
```
LayneHaber, line 490: will these defaults be sufficient?

packages/deployments/contracts/src/cli/init/config/mainnet/production.ts
```diff
@@ -397,6 +414,7 @@ export const MAINNET_PRODUCTION_INIT_CONFIG: InitConfig = {
       "1634886255": "0x5C711DB90dEc0a5B81C626968DEa4187a7f9C1F2",
       "6450786": "0x9435Ba7C661a0Fd477deED640491de8c100325A7",
       "6778479": "0x7616Bc6d0dee5E250BA5b3dDa6cbbB71786FB638",
+      "1818848877": "0xade09131C6f43fe22C2CbABb759636C43cFc181e",
```
LayneHaber, line 417: this should be the protocol admin safe: `0x50d22a34183e7cD3164184011998a3D50303f75d `

packages/deployments/contracts/src/cli/init/config/mainnet/production.ts
```diff
@@ -397,6 +414,7 @@ export const MAINNET_PRODUCTION_INIT_CONFIG: InitConfig = {
...
     },
     watchers: {
       allowlist: ["0xade09131C6f43fe22C2CbABb759636C43cFc181e"],
```
LayneHaber, line 420: we are missing addresses of our mainnet deployed watchers here, need to add them

packages/deployments/contracts/deployConfig/shared.ts
LayneHaber: it will require owner to change, so we should make sure they are in advance

LayneHaber:
> One thing more we have to take care of after the merge into prod is the gasEstimates values in chaindata repo. because it hasn't set up yet but also understandable right now. but we should definitely set the proper values as soon as we go live.

we also have to add the assets properly to make sure they are included in the watcher v2 implementation

packages/deployments/contracts/deployConfig/shared.ts
LayneHaber: (if this is on the hub, the owner will be the SC once transferred)

packages/deployments/contracts/src/cli/init/config/mainnet/production.ts
preethamr: Updated

packages/deployments/contracts/src/cli/init/config/mainnet/production.ts
preethamr: Updated

packages/deployments/contracts/src/cli/init/config/mainnet/production.ts
preethamr: > we are missing addresses of our mainnet deployed watchers here, need to add them

What other new watchers do we have, the V2 watchers ? @LayneHaber

ops/mainnet/prod/backend/config.tf
preethamr: Yes, as SDK server is only using public RPCs.

packages/deployments/contracts/src/cli/init/config/mainnet/production.ts
LayneHaber: added ours, requested info from others. we can add with protocol admin, we should wait to whitelist until they are confirmed updated with the latest version

packages/deployments/contracts/deployConfig/shared.ts
LayneHaber: verified against other spoke connectors

packages/deployments/subgraph/config/amarok-hub-v0.json
LayneHaber: it is in the template file: https://github.com/connext/monorepo/blob/fad4783676bfa7124092b9d86dc691a89871234b/packages/deployments/subgraph/src/amarok-hub-v0/subgraph.template.yaml#L221-L247

LayneHaber:
chaindata on main already has the watcher flags

I ensured the multi minter branch used by the watcher is up to date as well

@wanglonghong are there other gas estimates that need to be added? I see one for `execute` but im not sure how accurate it is: https://github.com/connext/chaindata/blob/fc0f11f0e25ffd4db20ca9839ef563764f533a02/crossChain.json#L2555

LayneHaber:
> we also have to add the assets properly to make sure they are included in the watcher v2 implementation

done

LayneHaber (approved):

preethamr (approved):

wanglonghong:
> chaindata on main already has the watcher flags
> 
> I ensured the multi minter branch used by the watcher is up to date as well
> 
> @wanglonghong are there other gas estimates that need to be added? I see one for `execute` but im not sure how accurate it is: https://github.com/connext/chaindata/blob/fc0f11f0e25ffd4db20ca9839ef563764f533a02/crossChain.json#L2555

What we need is `execute` and `gasPriceFactor` but the default value (1e18) will be used if unset. so that's fine. 
In terms of accuracy, I think we can give it a try with 350000.

wanglonghong (approved):

liu-zhipeng (approved):

just-a-node (approved):

PR #5078, opened 2023-10-28
feat: add signature check to hub and spoke propose

preethamr:
## Description

Add signature check to hub and spoke propose

## Type of change

- [ ] Refactoring


packages/deployments/contracts/contracts/core/connext/helpers/RelayerProxy.sol
```diff
@@ -366,4 +442,22 @@ contract RelayerProxy is ProposedOwnable, ReentrancyGuard, GelatoRelayFeeCollect
...
+    bytes32 payload = keccak256(abi.encodePacked(_aggregateRoot, _rootTimestamp));
+    // Recover signer
+    address signer = payload.toEthSignedMessageHash().recover(_signature);
+    if (!spokeConnector.allowlistedProposers(signer)) {
+      revert RelayerProxy__validateProposeSignature_notProposer(signer);
+    }
+  }
+
+  function _proposeAggregateRootCooledDown() internal view returns (bool) {
+    return block.timestamp > (lastProposeAggregateRootAt + proposeAggregateRootCooldown);
```
gotzenx, line 461:
```suggestion
    return block.timestamp >= (lastProposeAggregateRootAt + proposeAggregateRootCooldown);
```
It should be greater or equal otherwise it would always be `proposeAggregateRootCooldown` + 1.

packages/deployments/contracts/contracts/core/connext/helpers/RelayerProxy.sol
```diff
@@ -300,6 +342,40 @@ contract RelayerProxy is ProposedOwnable, ReentrancyGuard, GelatoRelayFeeCollect
...
+
+    lastProposeAggregateRootAt = block.timestamp;
+
+    // Validate the signer
+    _validateProposeSignature(_aggregateRoot, _rootTimestamp, _signature);
+
+    spokeConnector.proposeAggregateRoot(_aggregateRoot, _rootTimestamp);
+
+    transferRelayerFee(_fee);
+    lastProposeAggregateRootAt = block.timestamp;
```
gotzenx, line 376: duplicated assignment here and in line 368

packages/deployments/contracts/contracts/core/connext/helpers/RelayerProxyHub.sol
```diff
@@ -491,6 +470,42 @@ contract RelayerProxyHub is RelayerProxy {
...
+
+    lastProposeAggregateRootAt = block.timestamp;
+
+    // Validate the signer
+    _validateProposeSignature(_snapshotId, _aggregateRoot, _signature);
+
+    // Propose the aggregate
+    rootManager.proposeAggregateRoot(_snapshotId, _aggregateRoot, _snapshotsRoots, _domains);
+
+    lastProposeAggregateRootAt = block.timestamp;
```
gotzenx, line 506: duplicated assignment here and in line 498

packages/deployments/contracts/contracts/core/connext/helpers/RelayerProxy.sol
```diff
@@ -366,4 +442,22 @@ contract RelayerProxy is ProposedOwnable, ReentrancyGuard, GelatoRelayFeeCollect
...
+
+  function _validateProposeSignature(
+    bytes32 _aggregateRoot,
+    uint256 _rootTimestamp,
+    bytes memory _signature
+  ) internal view {
+    // Get the payload
+    bytes32 payload = keccak256(abi.encodePacked(_aggregateRoot, _rootTimestamp));
+    // Recover signer
+    address signer = payload.toEthSignedMessageHash().recover(_signature);
```
0xng, line 454: If this is intended to be deployed in all domains, and some of the allowlisted relayer addresses are the same, then I believe this is subject to signature replay:
1- Call this `proposeAggregateRoot` on Chain A
2- Malicious actor sees the arguments the relayer used (aggregateRoot, rootTimestamp, provided signature)
3- Malicious actor calls `proposeAggregateRoot` on Chain B with the same parameters, which will recover to the same signer as on Chain A.

If I didn't miss anything, a potential fix would be to add "chainId" as a parameter to the function and check to ensure we are on the right chain. We may need history as well or some clever timestamp check to avoid someone submitting the same root, timestamp, and signature twice on the same chain. Adding history, or even adding a nonce would be expensive on mainnet and I feel there are cheaper solutions.

packages/deployments/contracts/contracts/core/connext/helpers/RelayerProxy.sol
preethamr: Done

packages/deployments/contracts/contracts/core/connext/helpers/RelayerProxy.sol
preethamr: Done

packages/deployments/contracts/contracts/core/connext/helpers/RelayerProxyHub.sol
preethamr: Done

packages/deployments/contracts/contracts/core/connext/helpers/RelayerProxy.sol
```diff
@@ -300,6 +354,36 @@ contract RelayerProxy is ProposedOwnable, ReentrancyGuard, GelatoRelayFeeCollect
...
+    bytes memory _signature,
+    uint256 _fee
+  ) external onlyRelayer nonReentrant {
+    if (!_proposeAggregateRootCooledDown()) {
+      revert RelayerProxy__proposeAggregateRootCooledDown_notCooledDown();
+    }
+
+    // Validate the signer
+    _validateProposeSignature(_aggregateRoot, _rootTimestamp, lastProposeAggregateRootAt, nonce++, _signature);
+
```
liu-zhipeng, line 379: Add chainId as well?

liu-zhipeng: Why don't we use `lastProposeAggregateRootAt` in  `_validateProposeSignature` function directly? I believe that will save some gas?

preethamr: @liu-zhipeng adding lastProposeAggregateRootAt, is sufficient right ? Don't need nonce as well.

liu-zhipeng: `lastProposeAggregateRootAt` might be insufficient. because malicious actor can propose with same signature again and again.

packages/deployments/contracts/contracts/core/connext/helpers/RelayerProxyHub.sol
```diff
@@ -585,7 +654,11 @@ contract RelayerProxyHub is RelayerProxy {
   }
 
   function _propagateCooledDown() internal view returns (bool) {
-    return block.timestamp > (lastPropagateAt + propagateCooldown);
+    return block.timestamp >= (lastPropagateAt + propagateCooldown);
```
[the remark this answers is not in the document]
liu-zhipeng: https://github.com/connext/monorepo/pull/5078/commits/5b7d185982e0f2b81c90a45ec0529a36a47ce410

packages/deployments/contracts/contracts/core/connext/helpers/RelayerProxy.sol
liu-zhipeng: @preethamr  you are correct, we increased `lastProposeAggregateRootAt ` after proposing. So we can use that as the nonce.

liu-zhipeng: Added chain id

packages/deployments/contracts/contracts/core/connext/helpers/RelayerProxy.sol
```diff
@@ -38,27 +43,39 @@ interface IKeep3rV2 {
...
+  uint256 public lastProposeAggregateRootAt;
+
   mapping(address => bool) public allowedRelayer;
 
   // ============ Modifier ============
 
   modifier onlyRelayer() {
     if (!allowedRelayer[msg.sender]) {
-      revert RelayerProxy__onlyRelayer_notRelayer(msg.sender);
+      revert RelayerProxy__onlyRelayer_notRelayer();
```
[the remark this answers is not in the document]
liu-zhipeng: We can debug without msg.sender

packages/deployments/contracts/contracts/core/connext/helpers/RelayerProxy.sol
```diff
@@ -300,6 +357,30 @@ contract RelayerProxy is ProposedOwnable, ReentrancyGuard, GelatoRelayFeeCollect
...
+   * @param _rootTimestamp Block.timestamp at which the root was finalized in the root manager contract.
+   * @param _signature Signature from the approved proposer.
+   * @param _fee - Fee to be paid to relayer.
+   */
+  function proposeAggregateRoot(
+    bytes32 _aggregateRoot,
+    uint256 _rootTimestamp,
+    bytes memory _signature,
+    uint256 _fee
+  ) external onlyRelayer onlyProposeCooledDown nonReentrant {
```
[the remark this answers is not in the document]
liu-zhipeng: > should we make this virtual so we can override with a revert on the `RelayerProxyHub`? see note from ng:
> 
> > Small note to remember that we MainnetSpokeConnector reverts automatically when someone attempts to propose something, therefore we don't need to propose on mainnet/hub. Instead, in mainnet we call sendRootToHubSpoke after finalizing on the RootManager

[https://github.com/connext/monorepo/blob/1419a6d0d9aa1e50047e6dc2814e5f704fd3e82b/packages/deployments/contracts/contracts/core/connext/helpers/RelayerProxy.sol#L370](url)

packages/deployments/contracts/contracts/core/connext/helpers/RelayerProxyHub.sol
```diff
@@ -574,25 +633,20 @@ contract RelayerProxyHub is RelayerProxy {
...
   function _validateProposeSignature(
     uint256 _snapshotId,
     bytes32 _aggregateRoot,
+    uint256 _lastProposeAggregateRootAt,
     bytes memory _signature
   ) internal view {
     // Get the payload
-    bytes32 payload = keccak256(abi.encodePacked(_snapshotId, _aggregateRoot));
+    bytes32 payload = keccak256(
+      abi.encodePacked(_snapshotId, _aggregateRoot, _lastProposeAggregateRootAt, block.chainid)
```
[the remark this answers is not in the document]
liu-zhipeng: > see above comment re: domain id
> 
> also, we can make this function protected so we don't have to define it in both `RelayerProxy` and `RelayerProxyHub`

actually `_validateProposeSignature` function of RelayerProxy and RelayerProxyHub is different function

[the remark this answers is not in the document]
LayneHaber: ah I see -- we should probably name them more distinctively

packages/deployments/contracts/contracts/core/connext/helpers/RelayerProxyHub.sol
```diff
@@ -505,12 +562,29 @@ contract RelayerProxyHub is RelayerProxy {
...
+   * @param _proposedAggregateRoot The aggregate root currently proposed
+   * @param _endOfDispute          The block in which the dispute period for proposed root finalizes
+   */
+  function finalize(bytes32 _proposedAggregateRoot, uint256 _endOfDispute) external onlyRelayer nonReentrant {
+    if (!_finalizeCooledDown()) {
+      revert RelayerProxyHub__finalizeCooledDown_notCooledDown(block.timestamp, lastFinalizeAt + finalizeCooldown);
+    }
+
+    // Finalized the proposed aggregate root
+    rootManager.finalize(_proposedAggregateRoot, _endOfDispute);
```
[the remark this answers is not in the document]
LayneHaber: deciding to do this offchain after @liu-zhipeng and @preethamr

LayneHaber (approved):

LayneHaber (dismissed):

liu-zhipeng (dismissed):

liu-zhipeng (approved):

PR #5381, opened 2023-12-19
fix: add safe guard to lh to not propose when subgraph stale

preethamr:
## Description

fix: add safe guard to lh to not propose when subgraph stale
## Type of change

- [] Configuration / tooling changes
- [ ] Refactoring

## Related Issue(s)

Fixes

### DIFF
--- a/packages/agents/lighthouse/src/tasks/propose/propose.ts
+++ b/packages/agents/lighthouse/src/tasks/propose/propose.ts
@@ -110,6 +110,9 @@ export const makePropose = async (config: NxtpLighthouseConfig, chainData: Map<s
     // Start the propose task.
     const rootManagerMode: RootManagerMode = await context.adapters.subgraph.getRootManagerMode(config.hubDomain);
     const domains: string[] = Object.keys(config.chains);
+    let proposeHubSuccess = true;
+    let proposeSpokeSuccess = true;
+
     for (const domain of domains) {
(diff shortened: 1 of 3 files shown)


LayneHaber (approved):

packages/agents/lighthouse/src/tasks/propose/propose.ts
```diff
@@ -140,14 +154,17 @@ export const makePropose = async (config: NxtpLighthouseConfig, chainData: Map<s
...
         }
       }
     } else if (rootManagerMode.mode === ModeType.SlowMode) {
       context.logger.info("In Slow Mode. No op.", requestContext, methodContext);
     } else {
       throw new Error(`Unknown mode detected: ${JSON.stringify(rootManagerMode)}`);
     }
-    if (context.config.healthUrls.propose) {
+
+    if (context.config.healthUrls.propose && proposeHubSuccess && proposeSpokeSuccess) {
```
wanglonghong, line 167: I am still wondering why it shouldn't send a heartbeat if proposing the snapshot on one of the spokes fails.

wanglonghong (approved):

packages/agents/lighthouse/src/tasks/propose/propose.ts
LayneHaber: so were alerted if messaging halts

preethamr: How will we know of that failure ?
We could increase the heartbeat threshold to allow for multiple rounds of proposal failures.

PR #5494, opened 2024-01-08
fix: use ethers instead of txservice on relayer

liu-zhipeng:
## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes

### DIFF
--- a/packages/agents/relayer/src/bindings/relays/index.ts
+++ b/packages/agents/relayer/src/bindings/relays/index.ts
@@ -28,7 +28,7 @@ export const bindRelays = async (_pollInterval?: number) => {
 
 export const pollCache = async () => {
   const {
-    adapters: { cache, wallet, txservice },
+    adapters: { cache, wallet },
     config,
     logger,
     chainToDomainMap,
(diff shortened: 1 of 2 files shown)


wanglonghong (approved):

packages/agents/relayer/src/bindings/relays/index.ts
```diff
@@ -113,32 +112,37 @@ export const pollCache = async () => {
           gasPrice: gasPrice.toString(),
         });
 
-        const gasLimit = await txservice.getGasEstimate(+domain, transaction);
+        //const gasLimit = await txservice.getGasEstimate(+domain, transaction);
+        const gasLimit = await rpcProvider.estimateGas(transaction);
```
LayneHaber, line 116: won't we run into more rpc errors without going through the txservice?

LayneHaber (approved):
if requesting the RPC provider is the only way to fix for x1 then we can merge, but its not a good practice to hit the providers directly

packages/agents/relayer/src/bindings/relays/index.ts
liu-zhipeng: I believe that the latest version of ethersjs manages most of them, but we need to check txservice deeply later. for now current changes will make x1-testnet work properly as I tested

liu-zhipeng: all good to merge? @LayneHaber

PR #5522, opened 2024-01-10
feat: Improve connext relayer

liu-zhipeng:
## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes


packages/utils/src/peripherals/gelato.ts
```diff
@@ -100,36 +99,8 @@ export const getConversionRate = async (_chainId: number, to?: string, logger?:
...
-            undefined,
-            undefined,
-            jsonifyError(error as Error),
-          );
-      }
-    }, retryInterval);
-  });
   try {
-    const res = await axiosGet(apiEndpoint);
+    const res = await axiosGet(apiEndpoint, undefined, 5, 2000);
```
preethamr, line 103: Neat !

ops/testnet/prod/core/main.tf
```diff
@@ -462,7 +462,36 @@ module "relayer" {
   ingress_ipv6_cdir_blocks = []
   service_security_groups  = flatten([module.network.allow_all_sg, module.network.ecs_task_sg])
   cert_arn                 = var.certificate_arn_testnet
-  container_env_vars       = local.relayer_env_vars
+  container_env_vars       = merge(local.relayer_env_vars, { RELAYER_SERVICE = "polling" })
```
preethamr, line 465: This module will not have http server and, will not be able to respond to ping.

preethamr (changes requested):
Not sure how this improves performance of the relayer.
Currently the relayer is not bottlenecked on http requests, it is instead on the rate at which it can place new transactions on chain.

Both https requests, and cache poll pending, happen in separate threads, already right ?

liu-zhipeng:
> Not sure how this improves performance of the relayer. Currently the relayer is not bottlenecked on http requests, it is instead on the rate at which it can place new transactions on chain.
> 
> Both https requests, and cache poll pending, happen in separate threads, already right ?

For now, that is not bottlenecked, but when a lot of transactions are pending, so HTTP server are processing them, polling didn't process them as I tested.

ops/testnet/prod/core/main.tf
liu-zhipeng: could you update that?

preethamr (approved):

wanglonghong (approved):

PR #5536, opened 2024-01-11
feat: op roots cost

wanglonghong:
## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes


ops/mainnet/prod/core/config.tf
```diff
@@ -366,6 +366,10 @@ locals {
       "6778479"    = 10,
       "1818848877" = 10
     }
+    proverBatchWaitTime = {
+      "6648936" = 14400,
```
preethamr, line 370: We should change this 12 hours for Mainnet, and communicate to mods that, in rare cases it might take 12 hours, for a slow path transfer to complete.

We should have 4 hours for all other chains, by default, since if the batch is fully by then, it will still submit before that time.

preethamr (changes requested):

packages/agents/lighthouse/src/tasks/prover/operations/publisher.ts
```diff
@@ -276,6 +275,7 @@ export const enqueue = async () => {
                   throw new NoMessageRootCount(originDomain, targetMessageRoot);
                 }
                 const batchSize = config.proverBatchSize[destinationDomain] ?? DEFAULT_PROVER_BATCH_SIZE;
```
preethamr, line 277: This is currently both, minimum and maximum for a batch size.

What are your thoughts on having a range ?
`batchSize` is min, and adding a new `maxBatchSize` as max,
This allows us to, push as many messages as we can, without much risk

If so, what can the max be ?

wanglonghong: I don't see the clear improvement than having only `batchSize` but it requires us to add more config.

ops/mainnet/prod/core/config.tf
wanglonghong: updated

preethamr (approved):

liu-zhipeng (approved):

PR #5675, opened 2024-02-08
feat: monitor router status

wanglonghong:
## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes

### DIFF
--- a/packages/agents/sequencer/src/bindings/server/index.ts
+++ b/packages/agents/sequencer/src/bindings/server/index.ts
@@ -23,7 +23,11 @@ import {
   ExecStatusRequest,
   ExecStatusResponse,
   ExecStatusResponseSchema,
+  RouterPingRequest,
+  RouterPingRequestSchema,
+  RouterPingMessage,
 } from "@connext/nxtp-utils";
+import { verifyMessage } from "ethers/lib/utils";
 
 import { getContext } from "../../sequencer";
 import { MessageType, HTTPMessage } from "../../lib/entities";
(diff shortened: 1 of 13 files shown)


packages/agents/sequencer/src/bindings/server/index.ts
```diff
@@ -188,6 +194,21 @@ export const bindServer = async (queueName: string, channel: Broker.Channel): Pr
...
 
+  server.post<{ Body: RouterPingRequest }>(
+    "/router-ping",
+    { schema: { body: RouterPingRequestSchema } },
+    async (req, res) => {
+      const { router, timestamp, signed } = req.body;
+      const signerAddress = verifyMessage(`${RouterPingMessage}-${timestamp}`, signed);
+      if (router.toLowerCase() == signerAddress.toLowerCase()) {
+        await cache.routers.setLastActive(router);
+        return res.status(200).send({ message: "OK" });
```
preethamr, line 205: Should we respond to the router with the current data in the cache ?

packages/agents/sequencer/src/lib/operations/execute/fastpath.ts
```diff
@@ -52,6 +52,13 @@ export const storeFastPathData = async (bid: Bid, _requestContext: RequestContex
     });
   }
 
+  // Update the last bid time for a given router.
+  await cache.routers.setLastBidTime(bid.router, {
+    originDomain: transfer.xparams.originDomain,
```
preethamr, line 57: Would also be useful for routers to know when the last successful/accepted bid was.
Wdyt ?
@wanglonghong

packages/agents/sequencer/src/bindings/server/index.ts
wanglonghong: I think yeah. It should be fine even if the cache cleared cause it's being updated in the next cycle.

packages/agents/sequencer/src/lib/operations/execute/fastpath.ts
wanglonghong: That's good suggestion.

preethamr (approved):

PR #5828, opened 2024-03-10
Xkeeper support

LayneHaber:
## Description

- Allows our relayer to operate as a keeper
- Assumes the relayer is using the Keep3r Relay to call the data

If this implementation is okay, then the call for prove and process should be updated to attempt keeper first.

Associated test vault can be found [here](https://xkeeper-3k7ldt839-defi-wonderland.vercel.app/vault/0x0ad045A64A4c90252C9Ac75e5a7e2F1c4e863917).

## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes


packages/agents/relayer/src/bindings/relays/index.ts
```diff
@@ -89,16 +89,41 @@ export const pollCache = async () => {
...
+          // lookup relay address (consistent across chains, minus zksync)
+          // lookup automation vault
+          const { automationVault, xKeeperRelayer } = config.chains[domain].deployments;
+          if (!automationVault || !xKeeperRelayer) {
+            throw new Error(`No automationVault or xkeeperRelayer for domain ${domain}`);
+          }
+
+          // call keeperRelay.exec with the transaction data
+          // TODO: batch exec calls from cache that are keeper flagged
+          // TODO: fallthrough if keeper sending fails. likely in cache
```
preethamr, line 117: We could handle fallback at the caller level `sendWithRelayerWithBackup` so that it tries in this order:
1 Connext Relayer using Keep3r network, setup Keep3r api endpoint/flag on relayer HTTP server
2 Gelato
3 Connext Relayer as hot wallet.

preethamr (commented):
This is neat.

I thought about this a bit, my main concern - In order to get significant savings this (keep3r) has to be the default. If we do that, then we are expecting our relayer infra to be able to handle production load on tasks/chains for which we use it.

We have seen our relayer do 2-3 requests per minute, beyond that we might see reliability issues, especially when there is a burst of requests from the users, or even when our messaging layer hits a 2 hour mark, and we fire off PnP for multiple chains at once.

Suggest, we can selectively roll this out on mainnet, and LH PnP, and then gradually figure out how well it scales.

packages/agents/relayer/src/bindings/relays/index.ts
LayneHaber: that's the idea

we can prioritize keeper for prove and process, which is where we will see the most gains

PR #5935, opened 2024-03-29
Whitelist with nonexisting canonical domain

prathmeshkhandelwal1:
## Type of change

- [ ] Refactoring

## Related Issue(s)

Fixes


packages/deployments/contracts/src/cli/init/config/mainnet/production.ts
```diff
@@ -345,28 +345,6 @@ export const MAINNET_PRODUCTION_INIT_CONFIG: InitConfig = {
         },
       },
     },
-    {
-      // TODO: This is the old config, to be removed in execution layer upgrade
```
just-a-node, line 349: Keep this for reference, just comment out.

packages/deployments/contracts/src/cli/init/helpers/assets.ts
```diff
@@ -29,15 +29,6 @@ export const setupAsset = async (args: {
     `\n\tVerifying asset setup for ${asset.name} (${asset.canonical.address}). Canonical ID: ${canonical.id}; Canonical Domain: ${canonical.domain}; Key: ${key}`,
   );
 
-  // Set up the canonical asset on the canonical domain.
```
just-a-node, line 32: This setup script needs to be able to register xERC20s using the pseudo-canonical domain as well as maintain the ability to register other assets that are homed on Eth. I don't think we can just remove the home checks, we should still have those guards. 

We need to separate logic based on whether `11111` is specifically set as the canonical domain for an asset. If not, then the original checks should still apply.

PR #6008, opened 2024-04-09
fix: added pseudo canonical domain

prathmeshkhandelwal1:
Fixed asset whitelisting script to include xERC20 whitelisting. I.e Including pseudo canonical domain.

There is another PR for similar fix which resolves it by removing home canonical domain logic 
https://github.com/connext/monorepo/pull/5935

The issue with above PR is it only relevant for xERC20s and bypass some important check for non-xERC20 asset.

Rather than removing canonical domain logic, This PR include creating a dummy network in init script for pseudo domains and bypass `updateIfNeeded` in case of domain: `11111`

### DIFF
--- a/packages/deployments/contracts/src/cli/init/helpers/assets.ts
+++ b/packages/deployments/contracts/src/cli/init/helpers/assets.ts
@@ -61,28 +61,31 @@ export const setupAsset = async (args: {
     throw new Error(`Unable to find canonical decimals in config for ${asset.name}`);
   }
 
-  await updateIfNeeded({
-    apply,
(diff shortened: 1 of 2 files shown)


just-a-node (approved):

just-a-node:
Assuming this has been tested on both psuedo-domain assets and regular assets? At least in dry run mode.

prathmeshkhandelwal1:
> Assuming this has been tested on both psuedo-domain assets and regular assets? At least in dry run mode.

Yes, Also tested for deploy new local assets in case of new registeration

